PR #13, opened 2021-08-02
add weak link to uio

stepansnigirev:
this change allows usage of `import io` instead of `import uio`


jreesun:
LGTM! Thanks, this is a much cleaner approach.

PR #36, opened 2021-10-21
Update de-DE.json

ghost:
I changed the wording and corrected the grammar. In my opinion, it is sometimes better to use the English Version of a word (e.g.: "Wallet" instead of "Brieftasche") for a better understanding. The syntax has to be checked in the lines 38 and 53, because I did bigger changes and am not sure if my used syntax is correct. 

Here a list of all edited lines:
Line 1, 2 , 4, 8-10, 11, 15-17, 22, 25, 27, 29, 31, 33, 38 (check syntax), 45-52, 53 (check syntax), 54-56

Question about line 48. What is meant with "Try more?"? What should be tried more? I am not sure about it, that is why my translation for line 48 might not be perfect.


jreesun:
This looks great! I'll pull it down and try it out.

> Question about line 48. What is meant with "Try more?"? What should be tried more? I am not sure about it, that is why my translation for line 48 might not be perfect.

In the Check Address section, you can scan the QR code of a receive address and Krux will search through valid receive addresses one at a time until it finds a match. To avoid searching endlessly, it checks 100 addresses at a time and then stops to prompt the user if they want to "Try more?" (addresses) if there isn't a match.

PR #53, opened 2021-11-04
spanish translation for Krux

bluejayws:


jreesun:
Looks good. I made some follow-up tweaks to get all the words to fit on the screen. 

Thanks again!

PR #92, opened 2022-03-01
Smart Typing for Seed

odudex:
There are a some "not recommended" commits, but isolated the "impossible letters" at the last one.
Please let me now if there is a better way of doing this.


jreesun:
You'll need to make a new branch with just that last commit and turn that into a PR. Try this:
```bash
# Fetch the latest code so you're branching from the latest version of `main`
git fetch --all
# Create a new branch called `smart-typing` branched from `upstream/main` and check it out
git checkout -b smart-typing upstream/main
# Use git cherry-pick to take the changes in your last commit and apply them to the new branch
git cherry-pick 528d1f7b0342b8d6d7cd59ac6247e76e6df1069d
# You'll probably run into a merge conflict at this point since we both made changes to that file and yours were based on an older version, so you'll need to resolve those manually. Once you do, run:
git cherry-pick --continue
# You should now have a branch with just that one commit. Then just push it up and make a PR:
git push -u origin my-single-change 
```

PR #93, opened 2022-03-02
Smart seed typing

odudex:
Makes seed typing faster by disabling letters which would not form words belonging to BIP39 dictionary


odudex:
There were some issues, I think it´s better for you to manually add the code. Next contributions I'll spend some time on improving workflow

PR #96, opened 2022-03-04
Add SYS_EXC_INFO for micropython to handle exceptions

odudex:
logging.py uses sys.exc_info() in exception handling, for that MICROPY_PY_SYS_EXC_INFO needs to be enabled on micropython port, otherwise system may crash.

### DIFF
--- a/firmware/MaixPy/components/micropython/port/include/mpconfigport.h
+++ b/firmware/MaixPy/components/micropython/port/include/mpconfigport.h
@@ -177,6 +177,7 @@ extern const struct _mp_print_t mp_debug_print;
 #define MICROPY_PY_SYS_EXIT                 (1)
 #define MICROPY_PY_SYS_STDFILES             (1)
 #define MICROPY_PY_SYS_STDIO_BUFFER         (1)
+#define MICROPY_PY_SYS_EXC_INFO             (1)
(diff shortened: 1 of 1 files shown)


jreesun:
Thanks! (By the way, the comment-coverage test failing was unrelated to your change)

odudex:
Your welcome! (One of the motivation of this tiny PR was check if it would succeed, as I'm trying to make cleaner contributions)

jreesun:
@odudex 
Update:
I'm currently working on breaking out my changes to MaixPy into their own repo and submodule (to resolve https://github.com/jreesun/krux/issues/78), and I just noticed that I already made this change within the micropython core config here:
https://github.com/jreesun/krux/blob/main/firmware/MaixPy/components/micropython/core/py/mpconfig.h#L1153

So, I think this change may be redundant. Did you run into an issue due to it not being in `mpconfigport.h`? If not, I may remove this change from my fork if it's unnecessary.

odudex:
I did run into issues without this.
It was during tests, other another error I did raised an exception. Without
this the system would crash, with this the exception was presented at
device screen.

Em seg, 7 de mar de 2022 01:59, Jeff ***@***.***> escreveu:

jreesun:
Okay, good to know. I'll leave it as-is then, thank you!

PR #140, opened 2022-07-17
WIP

jreesun:
@odudex I'm going to do my work off of your branch here rather than `develop` to minimize merge conflicts


jreesun:
@odudex Nevermind, I changed my mind; I'll wait for you to make a PR instead in case you're working in a different branch locally or something. 

Can you make a PR with what you have now so I can branch from it and begin applying my changes for the simulator?

odudex:
This was my last code, you could work from it. I think you should set the correct Maixypy source. I also updated other dependencies

jreesun:
> I changed a few things in camera:
> 
>     * Use camera color.
> 
>     * To adapt for Dock: Use ENTER key to leave QR capture.
> 
>     * Changed a little how light turns on, on hope to fix cases where it didn't work in Amigo.
> 
> 
> Now fixing the tests, I tested to remove all image pre-processing (histogram+binary conversion). It didn't affect the `find_qrcodes()` method, it is still super efficient, and I also got noticeable frame rate increase at the screen. Could I leave it like this?

No, please leave in the modifications to the camera I made with respect to grayscale and binary conversion; it was very necessary for the M5StickV and helped significantly in low-light and glare conditions. The camera preview is far less important than a fast, accurate scan.

odudex:
Ok, can I still preview it colored, as the code is on current PR?

jreesun:
I believe the image needs to be in grayscale in order for Otsu's thresholding to work (the method being used to binarize the pixels into white or black). Here's a write-up on the process, where they seem to first convert the image into grayscale before running the algorithm on it:
https://datacarpentry.org/image-processing/07-thresholding/#automatic-thresholding

If it's possible to capture the image in RGB, display it, then convert to grayscale before binarization without significantly impacting the memory usage, then I suppose that would be fine.

What is the reason for wanting a color preview?

odudex:
IMO, colored gives a better user experience, makes it easier to understand what is being pictured, recognize the surroundings. Present it in grayscale is already better than black & white (binary). BUT, as I said, it's just my opinion.
So, should I leave it B&W?

odudex:
Test fixed
- bugs catched by tests fixed
- camera preview shows in grayscale
- Coverage still needs to go from 91 to 95%

jreesun:
My argument for leaving it B&W is that you see what the `find_qrcodes` function sees, which should make it easier to know how to fix your camera (either the angle, distance, or focus). That said, I can see why a "natural" view from the camera would be more visually appealing and useful at solving those problems as well.

Here's a potential compromise: since all devices have 3 buttons now due to your PMU button work, that frees up a button we could use to toggle between grayscale and B&W. We could default to the more-natural grayscale view but allow the user to press `BUTTON_C` to display the post-processed B&W image. I personally find the B&W view easier to resolve my camera issues with, so I'm not willing to get rid of it entirely ;)

odudex:
Right, I also like the "debug" feature on B&W, to toggle the view with a button click is a great idea, I will implement that!

jreesun:
@odudex Could you make that a separate branch/PR after we get this one reviewed and merged? In the future it would be easier and quicker (for me, maybe not you 😅 ) if we could keep pull requests small and more focused around one feature at a time. I'd be able to review faster, give better feedback, and feel more confident merging something if it were based around a single piece (or logical chunk) of functionality.

For example, this branch adds passphrase support, refactors keypad input, adds rotary encoder support, changes UI around dice entropy input, etc., so it's a lot to grapple with in one go and makes it hard to be selective about which ones become part of a release or not.

Anyway, this one's fine as-is since I'm looking to get caught up on all the new stuff as quickly as possible, but I just wanted to make things smoother in the future and hopefully reduce my lag in reviewing and merging.

odudex:
Sure, I would make PR almost as small as the commits. But as I told you, I was waiting my first big PR(Amigo compatibility) get merged to main, meanwhile new things got piled up.
From now on I'll make smaller ones, including "toggle" câmera preview.

jreesun:
Thanks! Yeah, I was gone a lot longer than I expected to be and fell behind, sorry about that. Excited to finally get your work out there in a proper release :)

odudex:
Tests, coverage, translations, black, lint: ALL GREEN

PR #148, opened 2022-07-31
Release 22.08.0

jreesun:
In no particular order, this release:
- Adds support for the Maix Amigo, Maix Bit, and Maix Dock devices
  - This includes an all new UI when using a touchscreen device
- Updates the simulator to add support for Amigo and PC directly
- Tons of font work
  - Change font format from dense to custom sparse format, krux font format
  - Add `bdftohex.py` for converting from a source bdf font format to the new krux font format
  - Add `hextokff.py` script to convert only used unicode characters (derived from looking in `i18n/translations`) to new format
  - Add `bdftokff.sh` shell script to simplify font conversion
  - Update `ter*.bdf` fonts to include vietnamese and bitcoin glyphs in one source font
  - Fix vietnamese glyphs by accounting for bbox values in bdf when converting to hex
- Build translations into firmware with new `i18n.py bake` command and use crc32 hash as slug lookup
- Adds a new Brazilian Portuguese translation
- Adds a new i18n framework for adding translations of the website documentation
- Adds BIP39 passphrase support, asking a user if they want to  "add passphrase" after entering a mnemonic
- Mnemonic generation from dice rolls has been changed:
  - Rolls can be entered much faster and the last N rolls will always be displayed along the top in exactly the way they will ultimately be hashed
  - Now when rolling a D6,  _no_ separator is used between rolls prior to hashing to match ColdCard and SeedSigner results
    - We continue to use a `-` separator for D20 rolls to avoid accidentally reducing state space due to collisions (e.g., rolling 1-17 and 11-7 would result in the same `117` string without a separator
- Unlocks the ability for a M5StickV to navigate backwards in all input prompts by pressing the power button. Long-pressing the power button will still shutdown the device, but one-press resets have now been disabled.
- Adds support for rotary encoder input in lieu of prev/next buttons
- Adds a lot of UX improvements including the ability to escape out of mnemonic loading or creation
- Updates the docs site to be device agnostic and include screenshots in all supported languages
- Adds the ability to load or save a .psbt file to an SD card, which should hopefully alleviate the pain with using the M5StickV when you don't have a phone-quality webcam to be able to read its QR codes.


odudex:
Wow, this is getting amazing! :exploding_head:
Just FYI, I did a few more things.

[Flash script more flexible to find suitable port](https://github.com/selfcustody/krux/commit/132f3f8db8d54d5b16379535046c2831433a0298)
When there are two ports(Bit and Amigo) it will automatically try to flash the second port if the first fails, so it doesn't matter the order ports were listed by OS.

[M5stickV will never have inverted LCD](https://github.com/selfcustody/krux/commit/c6c9c8a8ddc9df37ef38b08b12bfc7264f2ca6b2)
Unreachable code

[Remove flickering on animated QRs](https://github.com/selfcustody/krux/commit/eb3053613038b8a9b43c734787c484c8bd4e9b1f)
[Before](https://photos.app.goo.gl/NTb91c8a9JEuwZhD6)  -> [After](https://photos.app.goo.gl/to1K63PwBkiWdV5A9)
This will make coordinator's camera job easier, as it will reduce oscillation on gain/exposure levels during scan.

[Compact SeedQR (Binary QR) support](https://github.com/selfcustody/krux/commit/f4e4b24850f7de7401209f2bf3e1b8cb6ccbef79)

Let me know(or pull) if you would like some of them to be boarded on this release:

jreesun:
1) I'll incorporate this one since it's simple and doesn't need test coverage. Good idea

2) This is already resolved

3) This one would be a worthwhile change to add to the release, but I tested it locally on the M5StickV and found a bug with it: if the QR codes being shown are different sizes (this happens when the last QR code in the sequence has remainder data that may be larger or smaller than the previous codes), then parts of a previous QR code will be left as artifacts that obscure the next code.

4) This one isn't as important and needs to go through a code review and have test coverage added. Quickly looking at it, a new QR format shouldn't be added since the data is likely being sent as-is (so, I would expect it to be classified as `FORMAT_NONE`). Formats describe how the data is encoded for _transmission_, not the underlying format of the data itself. Does that make sense?

jreesun:
@odudex I think this is ready for release. Any chance you could load it onto your devices (especially the Bit and Dock since I don't have them) to test before I merge?

odudex:
> @odudex I think this is ready for release. Any chance you could load it onto your devices (especially the Bit and Dock since I don't have them) to test before I merge?


I did some tests with your branch a few commits ago and it was all good. I can run more complete tests with recent updates on all devices only on Tuesday.
> 
> Also, with your permission, I'm going to update all license files and headers to be copyrighted under "Krux contributors" similar to how bitcoin says "Bitcoin Core developers."

Please, feel free to modify anything you wish to

jreesun:
Great, I'll wait until Tuesday and put out the release after you give the thumbs up. 

Thanks again for all your hard work here. I think people are going to love Krux on the Amigo!

jreesun:
@odudex I snuck in one last change: the ability to load or save a `.psbt` file to an SD card, which should hopefully alleviate the pain with using the M5StickV when you don't have a phone-quality webcam to be able to read its QR codes.

odudex:
> @odudex I snuck in one last change: the ability to load or save a `.psbt` file to an SD card, which should hopefully alleviate the pain with using the M5StickV when you don't have a phone-quality webcam to be able to read its QR codes.

@jreesun I managed to do some tests. SD signing worked fine! Some people demanded this!
I don't know if I missed something, but I noticed on Amigo the keypads, Yes/No and seed display are flipped, it seems `inverted_coordinates` should be used there too.

jreesun:
@odudex Which Amigo build did you use? I switched the "default" image to be the `amigio_tft` since it seemed like the majority of ours were TFT.

jreesun:
@odudex Closing this PR and opening a new one for the release. Trying to adhere better to [Git Flow](https://nvie.com/posts/a-successful-git-branching-model/)

odudex:
> @odudex Which Amigo build did you use? I switched the "default" image to be the `amigio_tft`

Sorry, I forgot you already changed default to tft. The tft version worked fine.
I also still don't know a solution to make all 3 versions work without too much hassle.
Some things you should be aware off:
When I flashed my "regular" Amigo with ips version, the colors didn't got inverted (I would expect it to get white/inverted).
The Amigo with inverted colors I have here (third type), in terms of x coordinates, behave like the regular, so with IPS version the colors got fixed, but x coordinates got flipped.
The Amigo with non-flipped x coordinates (second type) may work with ips version, once the invert init parameter seems to don't affect Amigos with regular colors.

odudex:
Just in case you want, binary QR compatibility in a version with decoding as FORMAT None:
2d73c656efdf8ac08da8e6ac09f008a8cec3dddc

jreesun:
@odudex Can you try this latest code on all devices? I think it may (hopefully) work on all of them. 🤞 Since the color inversion is "one-way," I changed the code to always do it. Therefore, the difference between the `tft` and `ips` builds are just the inverted coordinates now.

odudex:
> @odudex Can you try this latest code on all devices?

IPS is still with  `"inverted_coordinates": false,`
For "type 3" to work, it should be `"inverted_coordinates": false,`, as TFT. But this would break "type 2"

odudex:
Sorry, brain lapse here, flashed both with TFT version, all good!

odudex:
So, both regular and "type 3" will work with TFT version, while "type 2" will work with IPS version. Smart!
Also to notice, even colors are good in both, regardless of  `lcd.bgr_to_rgb(True)`
amigos

odudex:
**Correction:**
`lcd.bgr_to_rgb(True)` will affect and switch colors of other devices, so I would modify to:
```
lcd.bgr_to_rgb(invert)
```

odudex:
Some people using Ubuntu got in this issue:
af0cf50996fb8ac80bd8d7c0aab6768dac3e6a26

jreesun:
> **Correction:** `lcd.bgr_to_rgb(True)` will affect and switch colors of other devices, so I would modify to:
> 
> ```
> lcd.bgr_to_rgb(invert)
> ```

Which of the three amigos (lol) get the wrong colors?

jreesun:
> Some people using Ubuntu got in this issue: [af0cf50](https://github.com/selfcustody/krux/commit/af0cf50996fb8ac80bd8d7c0aab6768dac3e6a26)

Did this resolve their problem?

odudex:
> EDIT: Do you mean that the Amigos are fine but this change broke the Dock/Bit's colors?

Yes, exactly

odudex:
> Did this resolve their problem?

Yes, after a reboot (maybe text should specify that too)

odudex:
More about the brltty bug:
https://bugs.launchpad.net/ubuntu/+source/brltty/+bug/874181
There are several reports online

jreesun:
@odudex Okay, should be good to go. Can you give it a final test on all devices before I merge and release?

odudex:
@jreesun I think we're good to go! Did some basic tests on all devices, focused on things with recent changes and everything worked fine. I runs smoothly and fast, the most noticeable improvement I felt was when using translations.
Thank you for the mention on release notes. Being part of Krux have been a pleasure and a great learning experience for me. Thank you for creating this project and paving the road for people to join and grow with it. :rocket::rocket::rocket:

PR #162, opened 2022-08-11
Bug fix - Amigo wouldn't update with SDcard

odudex:
Buttons were disabled at boot as it starts in "touch mode" so `ENTER` press will return `None` at first time

### DIFF
--- a/src/krux/firmware.py
+++ b/src/krux/firmware.py
@@ -197,6 +197,7 @@ def upgrade():
         t("New firmware detected.\n\nSHA256:\n%s\n\n\n\nInstall?")
         % binascii.hexlify(firmware_hash).decode()
     )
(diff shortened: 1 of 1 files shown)


jreesun (approved):

jreesun:
@odudex Now that the new release is out, could you update any stuff you have pointing to the old release to point to `https://github.com/selfcustody/krux/releases/latest` instead? It'll always redirect to whatever the newest version is. I'd like to avoid people downloading the out of date version so it isn't a big issue come next release time

odudex:
> @odudex Now that the new release is out, could you update any stuff you have pointing to the old release to point to `https://github.com/selfcustody/krux/releases/latest` instead?

Sure, will also add this to Telegram group description

PR #164, opened 2022-08-12
Colocate settings with their modules and generate settings menu

### LINKED ISSUE #159: Refactor Settings page to be generated from settings object
Currently, the settings you can change under Settings are hard-coded. However, we should be able to generate the UI by looking at the settings object itself. This would clean up some code in `login.py` and make future settings instantly accessible within the UI.

### LINKED ISSUE #160: Colocate settings with the module that uses them
All settings are currently defined in the `settings.py` file, but a better pattern would be to have modules that need settings register their own namespace. For example, the `I18n` settings class here could easily be moved into `i18n.py`, and the `locales` value could be derived from the locales in `translation_table` instead: https://github.com/selfcustody/krux/blob/main/src/krux/settings.py#L94

To do this, a `register` classmethod on `Settings` could be added that each module would invoke to register its namespace, and then `Settings` could iterate over the namespaces to instantiate them in `__init__`.

jreesun:
Resolves #159 and resolves #160


odudex:
FYI, I'm working on:
99 rolls for 24 words from D6 2b14e6528d95b156027a36c394d7e951aab4f75f
Make cameras independent from boards (still need to fix tests for this) 63264ec9444331bcd918a74a1de4e558a9024edd

PR #169, opened 2022-08-17
Make cameras plug&play and independent from board.py

odudex:
Make cameras independent from board configurations, so hardware kits shipped with different camera models will work out of the box. "sensor" and also "orientation" fields on `board.py` could be eliminated from Maixpy projects.


jreesun:
@odudex Is this ready for review?

odudex:
> @odudex Is this ready for review?

Yes, already tested on all devices

jreesun (approved):

PR #173, opened 2022-08-20
Add mnemonic format setting with support for CompactSeedQR

### LINKED ISSUE #172: Add setting to change mnemonic encoding in QR code for display & print
This setting would default to the existing BIP-39 Plaintext encoding we currently use, but would allow the user to switch between  SeedSigner SeedQR, SeedSigner CompactSeedQR, and the crypto-bip39 UR type.

jreesun:
Resolves #172


jreesun:
Closing in favor of #180

PR #178, opened 2022-09-12
Fix zpub QR exports and single-part UR parsing

### LINKED ISSUE #177: "Cannot sign" PSBTs from Blue and Electrum
Warned by a Telegram user, I verified signing PSBTs transactions from wallets generated by zpubs, like in Blue and Electrum mobile, is not working.
This error is being raised:
```
sigs_added = self.psbt.sign_with(self.wallet.key.root)
        if sigs_added == 0:
            raise ValueError("cannot sign")
```
PSBTs from Sparrow, with wallets created with xpub containing fingerprint and derivation still works fine.

jreesun:
Fixes #177


jreesun:
@odudex Want to confirm that this fixes the issue for you too?

Issue #176, opened 2022-09-02
"Auto Go" + Erase words


odudex:
Also feedback from Telegram:
Keypad's "Go" could be automatically called after 4 numbers are typed when entering seed "as numbers".
To be able to correct mistakes, when text box is empty, the del "<" key could erase the last typed seed word (not only in numbers keypad).

odudex:
Implemented shutdown confirmation, auto-Go for "Via Text" and "Via Numbers" and word deletion for all types of inputs, that could save a lot of clicks/touches. [video](https://photos.app.goo.gl/Y5vQ5ptJEvYA3Z6c9)
But before if refactor tests and make a PR:
@jreesun What do you think about this?

jreesun:
I think the "auto go" part is good, and I like the confirmation when you choose to shutdown. But the `<` key being used to delete the previous word feels odd to me. Could we instead have the screen that shows the word you've selected (which currently requires a button press/tap regardless) have a Yes/No prompt before proceeding? If you select No, it drops you back at the input screen.

odudex:
Will do it the way you suggested. It is simpler and more intuitive.

Issue #175, opened 2022-09-02
Shutdown confirmation 


odudex:
Feedback from telegram group:
A shutdown confirmation would good to avoid accidental shutdown, specially after the seed was entered, sometimes after a long process like dice rolls.

PR #180, opened 2022-09-15
Compact SeedQR export

odudex:
This PR contains Compact SeedQR export, partially attending  #172, as well some UX suggestions like:
- SeedQR in "transcribe" mode, with grids as guides
- Auto-Go when entering seed as text and numbers, closing #176
- Erase words was regressed and replaced by a prompt on word confirmation
- Shutdown prompt, closing #175
- Print QR prompt as standard yes/no


jreesun:
> Newer version of pylint will acuse some circular imports on updated settings structure. I'm afraid to refactor that myself 😬

Yeah, no worries, I noticed that too. I'll tackle it in the next month or so

jreesun:
Is this ready for review?

odudex:
Yes, it can be reviewed.

jreesun:
Hoping to review this over the weekend

odudex:
About the "transcribe" feature which you'll see. The idea is to make it easier to copy the QR code without any equipment.
[This is one way](https://photos.app.goo.gl/haV6Fg8KgZtaDLPe6). We can also supply templates, [like these](https://github.com/odudex/krux_binaries/tree/main/templates)
There can be other ways.

odudex:
Jeff! How are you?
I removed cyclic imports by gathering config and `i18n.py` classes in another file. I know it looks less organized, but it was the only way I could do it, would you like to add this to current PR? You probably could do this in a more neat way.
I also have another proposals for future PRs. You can try them on my binaries, as:
- Battery indicator
- Sensor tweaks and "anti-glare" mode to better scan (specially screens with black backgrounds - dark mode) for  OV7740, other sensors don't need it.
- Some punched metal seed formats export (tiny seed, 1248)

jreesun:
> Jeff! How are you? 

Busy! 😓 Sorry for the delay, finally looking at this now.

src/krux/pages/home.py
```diff
@@ -54,11 +56,82 @@ def __init__(self, ctx):
 
     def mnemonic(self):
         """Handler for the 'mnemonic' menu item"""
+        submenu = Menu(
+            self.ctx,
+            [
+                (t("Words"), self.display_mnemonic_words),
```
jreesun, line 62: How about:
- Words
- Plaintext QR
- Compact SeedQR
- Back

The "Compact SeedQR" option could then just be the transcribe version by default (it's still scannable, right?), and at the end the print prompt could appear.

In other words:
1. remove `display_compact_qr`
2. add `self.print_qr_prompt(code, FORMAT_NONE)` to the end of `transcribe_compact_qr`
3. rename `transcribe_compact_qr` to `display_compact_qr`

src/krux/pages/home.py
```diff
@@ -54,11 +56,82 @@ def __init__(self, ctx):
...
+                    grif_offset,
+                    grid_size,
+                    qr_size * grid_pad + 1,
+                    lcd.RED,
+                )
+
+        code, qr_size = self._binary_seed_qr()
+        if self.ctx.input.touch is not None:
+            self.ctx.display.draw_hcentered_text(
+                "Touch disabled\nPress Enter twice to leave",
```
jreesun, line 118: Could you change this to:
"Touch disabled for transcription.\nPress enter twice to leave."

Also, needs to be translated

src/krux/pages/home.py
```diff
@@ -87,6 +160,11 @@ def wallet(self):
             self.print_qr_prompt(wallet_data, qr_format)
         return MENU_CONTINUE
 
+    def _binary_seed_qr(self):
+        binary_seed = bip39.mnemonic_to_bytes(self.ctx.wallet.key.mnemonic)
```
jreesun, line 164: This isn't the right way to generate a Compact SeedQR. They drop the checksum bits at the end. Check out my implementation here:
https://github.com/selfcustody/krux/blob/3e1a77a6440339c938fb13962c28eadae3df6c78/src/krux/pages/home.py#L62-L68

src/krux/pages/login.py
```diff
@@ -247,8 +247,9 @@ def load_key_from_qr_code(self):
 
             if not words:
                 try:
-                    data_bytes = data.encode() if isinstance(data, str) else data
-
+                    data_bytes = (
+                        data.encode("latin-1") if isinstance(data, str) else data
```
jreesun, line 251: Did you confirm that this (`data.encode("latin-1")`) works on the device? I noticed some differences between Micropython and Python 3 when it comes to encodings around strings and bytes... some things work in the simulator but not on the devices, and vice versa.

src/krux/printers/thermal.py
```diff
@@ -119,7 +119,7 @@ def __init__(self):
 
         self.setup()
 
-        if not self.has_paper():
+        if self.has_paper() == 0:
```
jreesun, line 122: `has_paper` should return True or False, not a number

src/krux/printers/thermal.py
```diff
@@ -201,7 +201,7 @@ def has_paper(self):
         # Bit 2 of response seems to be paper status
         res = self.uart_conn.read(1)
         if res is None:
-            return False
+            return None
```
jreesun, line 204: ?

tests/test_camera.py
```diff
@@ -12,6 +12,7 @@ def test_init(mocker, m5stickv):
     from krux.camera import Camera
 
     c = Camera()
+    c.initialize_sensor()
```
jreesun, line 15: Not necessary to add this call since it's covered by the `test_initialize_sensor` test. This test is just asserting that you can instantiate a Camera without an exception being raised

jreesun (changes requested):

jreesun:
> I removed cyclic imports by gathering config and i18n.py classes in another file. I know it looks less organized, but it was the only way I could do it, would you like to add this to current PR? You probably could do this in a more neat way.
That's okay, I'll tackle it after I merge this PR into dev

jreesun:
> I also have another proposals for future PRs. You can try them on my binaries, as:
>
>    Battery indicator

Awesome!

>    Sensor tweaks and "anti-glare" mode to better scan (specially screens with black backgrounds - dark mode) for OV7740, other sensors don't need it.

👍 Could you also add the black-and-white camera display mode we previously discussed here? https://github.com/selfcustody/krux/pull/140#issuecomment-1187905279

>    Some punched metal seed formats export (tiny seed, 1248)

Cool!

simulator/kruxsim/mocks/lcd.py
```diff
@@ -131,7 +129,7 @@ def run():
                         x = og_x * scale + j
                         og_yx_index = og_y * (starting_size + 1) + og_x
                         screen.set_at(
-                            (offset_x + x, offset_y + y),
```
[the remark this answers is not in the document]
odudex: Yes, it worked!

src/krux/pages/login.py
odudex: yes, I came to this after looking for something that would work in both

src/krux/printers/thermal.py
odudex: I was getting value error raise when printer was not connected. I will do a better fix for this.
By the way, I printed with Amigo, made some small changes later on this topic too.
[printing](https://photos.app.goo.gl/7SyEPrKSU4z929Lw9)
[scanning](https://photos.app.goo.gl/SSzkPEJR2JwpZfbi9)

tests/test_camera.py
odudex: I've changed camera's code later so Camera is not initialized at boot, because there's no need to spend time and power on this at boot since it will be re-initialized later. But that can be reversed too.

jreesun:
Nice work! Sorry for making you wait so long. Hoping to come back to work on this project in a larger capacity again in the coming months :)

Also, I saw your Tiny Seed work. Very cool!

odudex:
Nice to hear from you! Don't worry!
Yes, I'm working on Tiny Seed scan now, there's still a lot to improve.
Let us know when you are back!

PR #183, opened 2023-01-31
Update installing.en.md

tadeubas:
Fix ktool install instructions for maixpy_dock (`-B dan` parameter)


jreesun:
Nice catch, thank you!

PR #185, opened 2023-02-07
Multiple Features

odudex:
This version is stable with all features it has. Commits are self explanatory, has updated translations, pylint, black. But tests are outdated.
@jreesun , do you think you can cherry pick the commits according to your priorities?


jreesun:
Great! I'll try to review it this weekend and cherrypick from it

odudex:
Hi @jreesun !
As we have a new contributor, @tadeubas, who is assimilating the krux project really fast and is excited to contribute, it would interesting if we could all share some kind of planning, and of course, I think you should lead this.
There are a lot of ideas to discuss, commits to review and tests to fix. I think the best way for us to work as a team is to set some tasks/milestones for the next release. And you are the one to set them.
I really would like to know your priorities and thoughts on this.

jreesun:
> As we have a new contributor, @tadeubas, who is assimilating the krux project really fast and is excited to contribute, it would interesting if we could all share some kind of planning, and of course, I think you should lead this.
There are a lot of ideas to discuss, commits to review and tests to fix. I think the best way for us to work as a team is to set some tasks/milestones for the next release. And you are the one to set them.
I really would like to know your priorities and thoughts on this.

Yes, I agree. Open to suggestions you or @tadeubas might have on what you'd like to see or find useful, and if this should all live on github or not? Currently, I have a milestone representing work I wanted to see completed before cutting the next major release, here: https://github.com/selfcustody/krux/milestone/3 But I'm not sure if that's the best way to go about it. I'll  look into what calver projects like Ubuntu do to organize their release schedule and decide which features go where.

PR #196, opened 2023-02-23
Thermal print scale and speed settings

odudex:
I've made some changes to improve thermal printing speed and reliability.
Most of Krux users from Telegram group who have thermal printers and I have gojprt qr203 model. The code works fine with it. One has gojprt qr204 model, which seems to have an issue with buffer (lost bytes), so I added the `scale` and `line delay`(speed) that help solve the issue. No one there has the Adafruit printer, so @jreesun could you test the thermal prints on it?
Other finding: Reducing scale can make blacks darker, I believe some printers/power supplies may have power limitations.
As all Krux users printers use 11 `heat dots`, I removed this config so all others fit in one screen in Amigo.

### DIFF
--- a/src/krux/display.py
+++ b/src/krux/display.py
@@ -308,23 +308,24 @@ class Display:
         time.sleep_ms(duration)
         self.clear()
 
-    def draw_qr_code(self, offset_y, qr_code, bright=False):
-        """Draws a QR code on the screen"""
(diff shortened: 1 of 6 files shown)


odudex:
Sorry to add features on a cleanup branch. But I thought regions mode was node ready unless it also had a zoomed mode, specially for M5stickV

Issue #38, opened 2021-10-27
Add option to generate mnemonic from camera feed


ghost:
It looks like SeedSigner does this, and it seems like a promising way to generate a mnemonic that is random if done right.

Will need to brush up on my stats knowledge, but here's my initial rough plan:

I'm thinking you would calculate entropy from the images using Shannon's formula and display to the user a progress bar / percent until enough entropy is collected. Once you have good/enough entropy, the words (numbers, really) are generated and then a Chi-squared test is performed on the generated words as a sanity check.

May want to also use a similarity function to compare the images and reject "duplicates" so that the user is forced to move the camera around?

tadeubas:
This issue could be closed, as it already was implemented by @odudex on the branch __new_features__
https://github.com/odudex/krux/

jreesun:
I asked @odudex to mark the feature experimental until we can add in some code to better gauge the entropy of the captured image. One image probably has more entropy than I was originally thinking, but I know it'll be less than the amount of bits in the image. I'd rather we try to estimate the entropy and display it to the user somehow, hence why I wrote this:
> I'm thinking you would calculate entropy from the images using Shannon's formula and display to the user a progress bar / percent until enough entropy is collected. Once you have good/enough entropy, the words (numbers, really) are generated and then a Chi-squared test is performed on the generated words as a sanity check.

This and https://github.com/selfcustody/krux/issues/2 are on my todo list for the next release

tadeubas:
Closing this issue as the featuare is already on the latest release and marked as experimental - https://selfcustody.github.io/krux/getting-started/usage/generating-a-mnemonic/#camera

tadeubas:
Can we consider this done in [release 24.03.0](https://github.com/selfcustody/krux/releases/tag/v24.03.0) ?

odudex:
It's still tagged as experimental. Let's wait for more feedback, study the possibility, or necessity of adding up different entropy sources, as other signers do, and then when we remove the experimental tag and close this issue

PR #197, opened 2023-02-23
Tests fixes + load_key_from_digits now makes usage of possible_letters

tadeubas:
load_key_from_digits now only allow valid numbers (1-2048), avoiding 0 and 2049


jreesun:
This new feature requires more discussion and should have an issue made for it.

Can you put only general cleanup / test fixing changes in these PRs into `dev-cleanup`? My focus with this branch is to essentially make changes I would have asked for during code review, but since you and odudex were building off the same branch I decided it'd be easier to just merge and follow it up with my changes over the next few weeks when I get time.

As for new features, for the 23.04.0 release milestone I'm focused only on refactoring and polishing what's here already. There is an outstanding feature I'd still like to see added, though, if you are up for trying: https://github.com/selfcustody/krux/issues/38

PR #198, opened 2023-02-23
Just translations, cleanup, refactor and tests fixes

tadeubas:
3 tests commented with a `TODO:`, 2 on `tests\test_input.py` (because of deadlock), and one on `tests\printers\test_thermal.py` because the error is so big that was causing the loss of the terminal previous logs on Windows


odudex:
I've testing recent changes and I would modify a few details on SD card reading and signing.
1 - **Present a message while trying to read the SD**: Even thought it's fast now, when there's no card it gives the impression Krux froze or crashed. So it could display something like "Checking for SD card" the same way it is done when loading the camera.
2 - I would **increase the time the current messages about saving or not configs on SD are displayed**. There's not enough time to read the hole message. It could be a `wait_for_button` too.
3 - I would **add a 4th item on `Sign` submenu:** `Sign SD PSBT` or `Sign PSBT from SD card`, so user choose what he wants right away and don't have to go through a sequence.
@jreesun what do you think?

jreesun:
> I've testing recent changes and I would modify a few details on SD card reading and signing. 1 - **Present a message while trying to read the SD**: Even thought it's fast now, when there's no card it gives the impression Krux froze or crashed. So it could display something like "Checking for SD card" the same way it is done when loading the camera. 2 - I would **increase the time the current messages about saving or not configs on SD are displayed**. There's not enough time to read the hole message. It could be a `wait_for_button` too. 3 - I would **add a 4th item on `Sign` submenu:** `Sign SD PSBT` or `Sign PSBT from SD card`, so user choose what he wants right away and don't have to go through a sequence. @jreesun what do you think?

I agree with 1 and 2. Haven't done a ton of SD card testing yet since the hot-plugging changes were added, so I wasn't aware of a slowdown. I'll check it out and see if it can be sped up.

As for 3, maybe we could do it as a submenu after "Sign PSBT" is selected, similar to how we do it elsewhere in the login page. The next screen would have "Via Camera" or "Via microSD" as options. However, I do think we should continue to allow either of those methods to output QRs _and_ optionally write the signed PSBT to microSD.

tadeubas:
> ...so I wasn't aware of a slowdown. I'll check it out and see if it can be sped up
The slowdown (~1s) only happens when you are using Krux without an SD Card or with a incompatible one. The code to adjust the timeout is [here](https://github.com/selfcustody/MaixPy/commit/9f0391a3605c6d7ca932b568a885751588870fbb#diff-45a821a6534e538b22f113f783c0138efac80d97a9781de1cfb0e1c282c40dddR170)

@odudex are you sure that you are using this recent modified version of MaixPy? I ask because we tend to just update Krux and forget the MaixPy external repo dependency.

I agree with **1** and **2**. For the **3**, I agree with @jreesun

odudex:
> @odudex are you sure that you are using this recent modified version of MaixPy?

I was not, sorry! With updated timeout I think it's no necessary

tadeubas:
> > @odudex are you sure that you are using this recent modified version of MaixPy?
> 
> I was not, sorry! With updated timeout I think the **1** and even **3** are not necessary

😄 It happens! When you said you got the impression that Krux froze or crashed I knew you probably hadn't updated because with the update of MaixPy it takes only around 500ms

PR #204, opened 2023-05-21
Create nl-NL translation (Dutch)

PMK:


tadeubas:
Hey @PMK thx for all the hard work!! We are enhancing Krux with lots of new features, but it is currently in Beta... It would be nice if you could test it and create the translations for all other phrases that now exists! Here you can see the code being develop by me and @odudex https://github.com/odudex/krux/ (branch new_features), and here you can access the compiled binaries: https://github.com/odudex/krux_binaries/ - It is important to emphasize that this version is in BETA and should not be used with large amounts, just for testing purposes.

A list of the new features are shown here (it is already outdated 😄 but it is better than nothing) https://github.com/selfcustody/krux/discussions/202

PR #206, opened 2023-05-29
Added translation to Polish

streamofstars:


odudex:
Nice! If you make the PR on [here](https://github.com/odudex/krux_binaries) I can quickly add it, but there's a LOT of new features to translate there too. Please check them out.

streamofstars:
You mean rather [here](https://github.com/odudex/krux/tree/odudex_develop)?

odudex:
> You mean rather [here](https://github.com/odudex/krux/tree/odudex_develop)?

Yes!

tadeubas:
No! This is already outdated... the correct branch on odudex repository is the new_features

odudex:
Sorry again, didn't notice the link was to "develop." Yes, "new_features", which is the default branch now, is also the one that contains latest code

streamofstars:
Can someone who knows the menu structure of the "new_features" branch inside-out enlighten me, are there any labels that have specific and strict length constrains? For example, I believe that - and I only realized that now, that "Esc" cannot be longer than 3 characters as it is placed on a on-screen keyboard button (or is it also used somewhere else?). Are there any other labels like that? Some labels in some languages can be considerably longer. For example "Go" in Polish will never be that short, depending on the context it can be "Dalej", "Następny", "Przejdź", "Uruchom", "Rozpocznij"...

Also, are there any labels that are listed just once in the json file but are used multiple times in different context? I don't see any like that on my own but if there are some like that, they need to be listed separately because in English it might be exactly the same word whereas in some other languages context and grammar rules may change the word. This is not the case here but imagine there is a label "Done" which is used for both: announcing some process has finished on the software side and as a button label for a user to click to confirm they finished doing something. In Polish the first case will be e.g. "Koniec" but the second use case will be "Zakończ". Of course, some Poles can argue with me that "Koniec" can be used in both cases and perhaps it can, but this specific example is just to show to potential problem.

odudex:
Hi, I would merge the PR and help complete the translations, but I noticed Polish uses special characters(like ęśćł) that are not supported by Krux firmware at the moment. We will have to work on the firmware to add Polish glyphs.

jdlcdl:
Darn, I hate to see a contributor walk-away, especially when they could be a potential contact/organizer for a community that might otherwise be out-of-reach.

I remember seeing this months ago, thinking "How cool is that?  A careful 'unknown' puts their head down, focuses, and then a whole new market becomes available, just like that."

odudex:
I do all I can to welcome and help and even motivate newcomers. I used to focus only what was fun for me, that is creating new features, but last months I realized the importance of keeping the project better organized. Krux users base is growing, I promise I'll dedicate more time on helping contributors onboarding. There are a similar case for Korean translation, a small community, excited to join the project waits for Korean glyphs.

jdlcdl:
I think I now understand the process of how adding a new translation works in krux, and how to update the firmware so that it has fonts that it will need based on existing translations files.

The next time I see a similar effort as streamofstars made here (I had seen this originally, but I had no clue what to do), I'll help to make sure that the translator has an updated translations file (instead of one based on last year's repo) and will be able to help their effort through the rest of the process so that it's ready for others to review/merge.

odudex:
Agreed, having a new release and an updated repository is n1 priority to avoid frustrated contributions attempts. Krux already has the tools to merge fonts and glyphs as it is done with Vietnamese, so it's no complicated. We have to measure the impact of adding many glyphs on firmware size, but I think there's room for more!

PR #242, opened 2023-09-15
Documentation for Settings: rotary encoder

tadeubas:
As @odudex pointed out here: https://github.com/selfcustody/krux/discussions/240#discussioncomment-7017173


odudex:
Really cool workflow automation! But there are a few issues to fiz in images. I noticed some brightness differences on fonts, not big deal, but there are some missing glyphs too.
image

tadeubas:
I don't know why the simulator on Windows behave that way with the Krux logo... but I've returned the correct ones. At least now it is very easy to generate all images and replace them in docs already scaled and with transparent background!

odudex (approved):
Looks good! Thank you!

PR #249, opened 2023-09-18
Other docs small fixes

tadeubas:
Fixed PR with only doc changes now


odudex:
Commit merged together with #250

PR #255, opened 2023-09-29
Screensaver

tadeubas:
Added screensaver on menu screens and other small changes!


jreesun:
As Krux matures, we should really try to avoid doing "drive-by" feature additions like this without either filing an issue or opening a discussion first to first see if we want to add it, and second hash out the details of how it should look and work.

tadeubas:
This is just a already working PR, we can discuss it more here, create an issue or open a discussion pointing to this PR... and if we find this irrelevant, just close the PR pointing the arguments / issue / discussion

jdlcdl:
This branch worked for me, as far as I can tell.

I used ktool's "-s" argument to boot the firmware without touching SPI Flash.  In settings, screensaver was set to 5m so I switched it to 1.
When the screensaver came on, it did what I expected, and I left it running wondering if the board would heat up noticeably to the touch, but it didn't (this is a good thing).

However, after a few rounds of the screensaver starting and being woken up by touch, I noticed that my touches were not registering.  When I went to settings/touchscreen, I was on 22 as expected for my amigo.   I reset the board and asked ktool to boot the firmware again with -s option, and I have not yet been able to reproduce the problem where waking from screensaver effectively "bugs" my touchscreen.  I don't know what happened there.

One error that I did catch, never seen before, is entering touchscreen and then hitting "Esc" without changes shows me
```
Error: AttributeError("'Touch' object has no attribute 'touch_driver'",)
``` in red.   But I notice no other problems after touching past this error screen.

odudex:
> ```
> Error: AttributeError("'Touch' object has no attribute 'touch_driver'",)
> ``` in red.   But I notice no other problems after touching past this error screen.
> ```

I did some recent adjustments on touch, some bugs may be expected, I think this branch don't have the latest

jreesun:
> This is just a already working PR, we can discuss it more here, create an issue or open a discussion pointing to this PR... and if we find this irrelevant, just close the PR pointing the arguments / issue / discussion

@tadeubas Can you update your PR to explain why Krux needs a screensaver with thoughts on how it should function? Is this privacy/security related or aesthetics? I'd like to know the motivation for it. Does the battery life suffer because of this?

For future reference, this is how I'd like to see development progress: https://github.com/selfcustody/krux#contributing

Without a paper trail or social protocol to decisions like this, the project is in chaos. That's fine when you're a solo developer, but not when you're working as a group.

tadeubas:
The project is not in chaos, I am always verbalizing myself on the Telegram group and here on Github, also I always check things with odudex personally, I just don't want bureaucracy and other things that makes me less likely to contribute, because I really like the project! Anyway, I think this is a "develop" branch and nothing added here NEEDS to be in the release, so I don't think this screensaver is of a big issue to worth discussing, it is just a screensaver... do you use Krux? I have a dock device, sometimes I left it on testing and working with things, the screen is very bright and I wonder if pixels could fade with time, also I live in Brazil and I don't have too much money to be spending on new devices.

PR #272, opened 2023-10-19
erase_spiffs() to do so in 4k chunks

jdlcdl:
As discussed [here](https://github.com/selfcustody/krux/commit/3d6a3bb26bbe0d71277f49973f1ccf705a3f0a71#r130414686)


odudex:
I reproduced your tests and results. Thank you!

PR #275, opened 2023-10-24
Wallet output descriptor load from and save to SD card

tadeubas:
- Wallet output descriptor load from and save to SD card
- Fix screensaver button


odudex:
Created a discussion about UI enhancements https://github.com/selfcustody/krux/discussions/280

PR #279, opened 2023-10-26
Changes atop commit 0752bcf intended so that tests pass

jdlcdl:
I'm just taking baby-steps with some 'harmless' (I hope) tests changes, but mostly so that I can have feedback where I've missed the more correct solution.

* 1st commit feels safe, print_standard_qr() is now in utils.Utils.
* 2nd commit, not sure if it's a solution, or if it's disguising an error, because I'm not able to locate which "wait_for_something" prompt was added that makes an extra `<enter>` necessary.
* 3rd commit feels safe, screensaver waits for button instead of press, however I also altered the setting of "logo" variable because I assume that the [1:-1] slice notation was meant to strip the '\n' chars that are first and last, instead of only stripping one last space among others that would not be stripped.  I also considered expressing it like "logo string".split('\n')[1:-1] so that it's expressively stripping first and last blank lines.

### DIFF
--- a/tests/pages/test_home.py
+++ b/tests/pages/test_home.py
@@ -256,7 +256,7 @@ def test_mnemonic_standard_qr(mocker, m5stickv, tdata):
         home = Home(ctx)
 
         mocker.spy(home, "display_qr_codes")
-        mocker.spy(home, "print_standard_qr")
+        mocker.spy(home.utils, "print_standard_qr")
         home.mnemonic()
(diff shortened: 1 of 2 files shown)


odudex:
Thank you Jean!
About 2nd commit, don't worry about it. I have plans to replace those sequence of prompts by a menu, so tests will have to be adjusted again for this one.

PR #288, opened 2023-11-18
Bugfixes

odudex:
Bugfixes
Encrypted Mnemonic QR codes would fail to decrypt if PBKDF2 iterations settings was changed to non multiple of 10,000.
QR code transcription helpers that highlight regions could crash on edges of some QR code sizes.
Address navigation "previous" menu option wouldn't show correct number.


i18n/translations/fr-FR.json
```diff
@@ -225,6 +225,7 @@
     "Use camera's entropy to create a new mnemonic": "Utilisez l'entropie de la caméra pour créer un nouveau mnémonique",
     "Used: ": "Utilisé: ",
     "Value %s out of range: [%s, %s]": "Valeur% s hors de portée: [% s,% s]",
+    "Value must be multiple of %s": "La valeur doit être multiple de% s",
```
jdlcdl, line 228: "La valeur doit être un multiple de %s"

No big deal on the missing "un", but the space is on the wrong side of the % above and won't replace, i think.

odudex: oh I missed the "un". If I get another erro I'll put this together, otherwise let's leave it for next release

tests/pages/test_login.py
```diff
@@ -1,5 +1,13 @@
+import sys
 import pytest
+from Crypto.Cipher import AES
```
tadeubas, line 3: plz remove this unused libs

tests/test_settings.py
```diff
@@ -90,3 +99,105 @@ class TestClass:
...
+                BUTTON_PAGE_PREV,
+                BUTTON_ENTER,
+            ),
+            # Assert iterations changed to a multiple of 10,000
+            lambda: Settings().encryption.pbkdf2_iterations == 110000,
+            NumberSetting,
+        ),
+    ]
+    for case in cases:
+        ctx = create_ctx(mocker, case[0])
```
tadeubas, line 196: Context and Login objects are not necessary, you could call the Settings functions as other examples on this file. So all the navigation procedures would also be removed from the test

i18n/translations/fr-FR.json
odudex: Added this fix to integrated_changes after I merged changes from main

PR #290, opened 2023-11-21
Encryption load error now shows only one error message

tadeubas:
* Encryption load error now shows only one error message
* Test case for encryption + decription in sequence


odudex:
Great!

PR #292, opened 2023-11-22
Refactoring krux script

tadeubas:
* Adding checks for params and error messages on krux script


tadeubas:
@odudex I've added a commit to fix simulator with missing mocks and a small fix to the first words suggested to import mnemonic from words (none word start with X, so it is disabled for the first char of the words). Can you plz accept this PR?

odudex:
Thank you!

PR #294, opened 2023-11-25
qr_code_frame_calc branch: code review + tests

tadeubas:
Doing code review and tests on the new branch `qr_code_frame_calc`

### DIFF
--- a/src/krux/pages/addresses.py
+++ b/src/krux/pages/addresses.py
@@ -91,7 +91,7 @@ class Addresses(Page):
                     self.ctx.display.clear()
                     self.ctx.display.draw_centered_text(loading_txt % (num_checked + 1))
 
-                    pos_str = str(num_checked + 1) + ". "
+                    pos_str = str(num_checked + 1) + "." + " "  # thin space
                     qr_title = pos_str + addr
                     items.append(
                         (
--- a/src/krux/qr.py
+++ b/src/krux/qr.py
@@ -185,10 +185,9 @@ def max_qr_bytes(max_width):
     max_width -= 2  # Subtract frame width
     qr_version = (max_width - 17) // 4
     try:
-        capacity = QR_CAPACITY[qr_version - 1]
+        return QR_CAPACITY[qr_version - 1]
     except:
-        capacity = QR_CAPACITY[-1]
-    return capacity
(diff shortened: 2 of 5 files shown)


src/krux/pages/addresses.py
```diff
@@ -91,7 +91,7 @@ def list_address_type(self, addr_type=0):
                     self.ctx.display.clear()
                     self.ctx.display.draw_centered_text(loading_txt % (num_checked + 1))
 
-                    pos_str = str(num_checked + 1) + ". "
+                    pos_str = str(num_checked + 1) + "." + " "  # thin space
```
jdlcdl, line 94: I'm not seeing the difference in these two values.  Is it just to be more explicit that the space is needed, so it is not carelessly removed in the future?

odudex: Thin space is used as a trick. Krux line breaker will split words detecting spaces, but "thin spaces" won't be detected. On this case Tadeu found out multisig address wouldn't fit on Dock's screen, thin space will ensure the address number prefix will be in the same line as the address start (not in a exclusive line) so the screen space will be better used and address will fit.

tadeubas: @jdlcdl here you can see the unicode character: https://www.compart.com/en/unicode/U+2009

jdlcdl: Nice! I didn't even realize it was not a normal space character.  Thank you for the explanation.

src/krux/qr.py
```diff
@@ -88,7 +95,7 @@ def total_count(self):
             # Single-part URs have no expected part indexes
             if self.decoder.fountain_decoder.expected_part_indexes is None:
                 return 1
-            return self.decoder.expected_part_count()
+            return self.decoder.expected_part_count() * 2
```
odudex, line 98: Where `*2` comes from? 
`fountain_decoder.py` has an `estimated_percent_complete()` which uses:
`estimated_input_parts = self.expected_part_count() * 1.75`

tadeubas: The 1.75 is an estimative they use, but it is wrong! In the end the scan will only finish after reading X estimated parts + X index parts... so it is 2 times the estimated parts

PR #296, opened 2023-11-26
Reduced settings writes without defaults

jdlcdl:
Building on top of #295, this pr (arguably more dangerous) alters the settings.Store structure so that default values and their empty namespaces are never stored in settings.json.

* krux.settings.Setting.set still calls store.set() but only for custom settings, it will call store.delete() for defaults -- to clean-up the store
* krux.settings.Store.get no longer builds-out the settings dictionary, works with a deepcopy instead
* krux.settings.Store.delete is used to remove default values and empty parent namespaces
* an all-defaults settings.json file will contain an empty dict "{}"
* adds new tests and adjusts existing tests.

### DIFF
--- a/src/krux/pages/settings_page.py
+++ b/src/krux/pages/settings_page.py
@@ -25,7 +25,7 @@ from ..themes import theme, RED, GREEN, ORANGE, MAGENTA
 from ..settings import (
     CategorySetting,
     NumberSetting,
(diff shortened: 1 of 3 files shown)


odudex:
Thank you @jdlcdl! This is NEAT! Let's go with this one, less is more!

Issue #299, opened 2023-12-04
Better inform the user the risks of using encrypted mnemonic stored on flash or SD card


tadeubas:
As of `23.09.0` Krux allows users to store an encrypted mnemonic on flash or SD card. It is also possible to delete the stored mnemonic. But as @jdlcdl investigations pointed out https://github.com/selfcustody/krux/discussions/269 the delete word may pass the wrong idea to the user. Users need to be aware that deleting a mnemonic stored on flash will not remove completely the mnemonic from flash, but will only mark the content as deleted and an attacker can obtain that mnemonic later with physical access to the device. It is important to notice that this issue may affect the storage on SD card too.

I know we can add more info on the [docs ](https://selfcustody.github.io/krux/getting-started/usage/navigating-the-main-menu/#encrypt-mnemonic) but my point is that we can't rely only on the docs to inform the user, we need to add a warn message every time the user will store or delete a mnemonic from flash or SD storage. This is specially important because the user may test this functionality with their real mnemonic and a poor key (just to see how encrypted mnemonic works), then later delete the mnemonic hoping everything is ok and the device or the SD card are clean. As @jdlcdl pointed to me "delete really isn't delete" and we need to make that clear to the user, tell them what Krux does (and maybe what they should do).

odudex:
As Krux features grow it will get harder to forecast every possibility of usage and how users understand the inner workings and the risks of the tools they use. We must aim for a balanced set of instructions and restrictions. Too much and them and the UX will be compromised, to little and we may have risks and bad usage materializing. We will not be able to teach about filesystems through M5stickV screen, so it is important to be objective and minimalist at the device's UI, while keeping details in documentation. I believe the "delete" word is still appropriate, as it does the same thing as it does in any OS, where there's no warning someone could retrieve that information using memory sweeping tools.
Having that said, i think we could:
- Add a message when deleting from SD, ex:
  "To ensure data is not recoverable, full format your SD card using a computer."
  On documentation we better describe what a full format is and when it is appropriated to do it.
 - Add a message when deleting from flash, ex:
  "To ensure data is not recoverable, use Wipe Device feature from Settings"
  On documentation we better describe wipe feature.
- Move `Wipe Device` feature from `Settings` to `Tools`, to be together with `Delete Mnemonic`

tadeubas:
I would suggest some additions / changes:
* On the screen `Home > Encrypt Mnemonic`:
  * Store on SD:
[...]

PR #300, opened 2023-12-07
seeds.json on sdcard padded to avoid abandoned bytes

jdlcdl:
Related to issue #299

### Description
On an sdcard's FAT filesystem, when a file grows into additional sectors and is later trimmed, bytes will remain abandoned in previously used sectors.  For `seeds.json`, which holds encrypted mnemonics, this is less than ideal because it could result in ciphertext with weak encryption keys remaining on the sdcard media yet hidden from plain sight for the user.

This pr alters krux.encryption.MnemonicStorage's `.store_encrypted()` and `.del_mnemonic` methods so that `seeds.json` gets padded with spaces -- to its original size.  That is, it will never be trimmed.  Care elsewhere should be taken so that `os.remove()` is not used for `seeds.json`, as it abandons all bytes.

CAUTION: This pr may create a false-sense-of-security in the event that it fails to work, and it certainly can do NOTHING to save the user from themself if they choose to edit/remove seeds.json from the microsd outside of krux.

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/src/krux/encryption.py
+++ b/src/krux/encryption.py
@@ -150,8 +150,11 @@ class MnemonicStorage:
             # load current MNEMONICS_FILE
             try:
                 with SDHandler() as sd:
-                    mnemonics = json.loads(sd.read(MNEMONICS_FILE))
+                    contents = sd.read(MNEMONICS_FILE)
+                    orig_len = len(contents)
+                    mnemonics = json.loads(contents)
             except:
+                orig_len = 0
                 pass
 
             # save the new MNEMONICS_FILE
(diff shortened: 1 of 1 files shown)


odudex:
Another neat improvement! Of course documentation should contain detailed instructions and the warnings you raised, but with this commit do you think it is still necessary to add a message when removing encrypted mnemonics from SD card?

jdlcdl:
I still like the idea of a message.  The user should not be misled into believing that we've mastered the complex layer that is the filesystem, and that it is no longer capable of hiding data storage (which is it's primary job) from the level that krux works with file storage.

A careful user, hopefully via the docs, should know the difference between a verifiably scrubbed sdcard and our "best efforts to trick the filesystem into destructive management.".  I almost think that this pr is "treachery"... however it seems to be working so far (when files shrink, as well as when they get fragmented into non-contiguous sectors).

jdlcdl:
As devils advocate (far from evil... leaning in the opposite direction from this pr, towards "ultra conservative"),

I think a very strong argument can be made for:

We should not imply any functionality of "deleting" a stored encrypted mnemonic.  Not on sd, and not on flash.  That way, the user has no expectation that it can be removed securely via simple management, and they'll be motivated to find 'wipe spiffs' and/or a verifiably destructive sdcard format tool.

jdlcdl:
from draft to ready... but before the next release...

**WHICH FILES MUST WE BE CAREFUL WITH ???**... is it just seeds.json, or is it encrypted seedqr's saved to sdcard too? and what else?

PR #302, opened 2023-12-25
an alternative to krux_colors.py

jdlcdl:
### Description

An alternative utility script to krux_colors.py, for review by others...
potentially to replace krux_colors.py because it converts both ways, and because I believe there is a green error in krux_colors.py (however if I'm correct and krux_colors.py is adjusted to solve green, rgbconv.py would still require an option to swap byte-order).

I have a gist with further thoughts: [here](https://gist.github.com/jdlcdl/b9f476be4f5e1540ae403eb7e8310b1e)

- [* ] Bug fix
- [* ] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/firmware/scripts/rgbconv.py
+++ b/firmware/scripts/rgbconv.py
@@ -4,9 +4,10 @@
 
 from binascii import hexlify, unhexlify
 
-def rgb24_to_rgb16(rgb):
+def rgb24_to_rgb16(rgb, big=False):
     """
         convert 3 bytes of rgb888 into 2 bytes of rgb565
+        default to little-endian, so rgb565 becomes gbrg3553
     """
     assert type(rgb) == bytes and len(rgb) == 3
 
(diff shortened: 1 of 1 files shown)


odudex:
Good catch! I wonder your previous work reviewing Shannon's entropy, sweeping color spectrum in search of higher entropy lead you to this finding.
I believe we will have to review theme colors. Probably some funny things I noticed on colors aren't caused by display quality as I previously though, but this green bits order issue.
Your neat quality enhancing contributions are very inspiring. Thanks!

jdlcdl:
>I wonder your previous work reviewing Shannon's entropy, sweeping color spectrum

Yes, that's exactly what caught me onto this.   For days, it's been bugging me, making me feel guilty that I never understood what was happening... and that I had bailed before figuring it out.

odudex:
Great PR, now it's time to review theme colors, especially grays. Will also change simulator and Android code using `rgb16_to_rgb24` as reference. Thanks!

PR #327, opened 2024-02-27
Added  wdt.feed() on tools > wipe divice

tadeubas:
* added wdt.feed() on `tools > wipe divice`
* `tools > wipe divice` extra msg: "Do not power off, it may take a while to complete."
* Translations

Docs:
* Removing ex: as this is best to avoid in english
* docs more info about printers

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


jdlcdl:
Yikes, this never even crossed my mind, but it should have.  I suspect it was rebooting leaving user with false sense that it worked.

tadeubas:
Hello @jdlcdl! Amazing job you and @odudex did on the wipe device feature! On my dock the internal flash memory is about 2.7 MB and erasing a full memory would take about 38s (currently wdt timeout is 30s), so yes, it could be the case (in rare circumstances) that it was rebooting before finishing the work!

PR #334, opened 2024-03-13
Code review, small changes

tadeubas:
Found this small changes while doing code review:

* Coverage: returning target to 100% (as Jeff wanted, our goal is 100%) - set threshold to 10% (we currently have 93% code coverage, we could lower the threshold to 8% if wanted)
* Updated versions of pytest and pytest-cov to avoid warnings
* Disabled pylint W0212 in the whole `camera.py` file instead of per line. If wanted, it can be disabled by function too

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other


odudex:
Target 100% with 10% threshold won't change current approval behavior(95-5 == 100-10). I made this change, together with new patch requirements, to reflect a more realistic target. 100% is not only impractical, but, IMO, can also be prejudicial (as any utopic goal). Jeff said he would like 95% coverage, minimum 90%, these requirements are and will be kept.
Thanks for the review! We are about to release now, please do a PR with package updates on next development branch.

PR #335, opened 2024-03-13
Code review, refactor and small changes

tadeubas:
Found this small changes while doing code review:

* Updated versions of pytest and pytest-cov to avoid warnings
* Disabled pylint W0212 in the whole `camera.py` file instead of per line. If wanted, it can be disabled by function too
* Removed `flash_text()` from `firmware` and `pages/__init__`. Added to `Display`
* Created new helper function `flash_error` in `Display`
* Optimized `Dockerfile` for better cache usability (tested the build process and it yields the same sha256sum as before)
* Simplifying translations by removing format characters like `\n`, `%s` and `%d`
* Refactored `i18n.py` for better usability
* Created new poe tasks for `i18n`(translations) and `simulator`
* black + lint

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other


i18n/i18n.py
```diff
@@ -86,8 +86,8 @@ def validate_translation_files():
         sys.exit(1)
 
 
-def fill_missing():
```
jdlcdl, line 89: ok, got it.  the command is still "fill" but this script doesn't fill/alter translation files, rather it prints what is missing.

jdlcdl:
as of release 24.03.0, I'm doing my best to follow each pr (as long as I get a chance before it gets merged).  I used to react with "eyes" emoji on the convo to indicate that I've read the code changes, but it's not so useful when subsequent commits are added.  I'll be using a different notation for the future.  Please interpret the following as "jdlcdl has reviewed code changes as of commit x".  I'll leave more indications than below if I see an obvious problem.

a/o c88dc55: seen/unit-tested

odudex:
Some conflicts were created after I merged your `main` documentation PR, can you solve them?

PR #345, opened 2024-03-28
Dockerfile better cacheability, translations and refactors

tadeubas:
* Dockerfile better cacheability
* Poetry poe new task git-update 
* Refactor translations for simplicity, removing text formatting
* i18n `translations.py` outputs files ordered
* Refactor encryption_ui.py for code reuse
* Merged with feature "automatic shutdown"
* Fixed failing tests from feature "automatic shutdown"
* Added translations from feature "automatic shutdown"

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other


src/krux/pages/login.py
odudex, line 1: Won't we have issues separating ":", or other text punctuation?

jdlcdl:
re: translations, I like the recent pattern of removing and adjacent-white-spacing and punctuation from translation strings, but I don't know if latin punctuation is understood for all locales, perhaps they have their own?

regarding no-longer translating some Proper Names, I'm assuming that this is because these names are NOT expected to have translations? That viet-namese speakers will know "BIP39", "TinySeed", "Adafruit", "SeedQR" etc -- instead of localized words that might be better?  If so, "Stackbit-1248" and "Compact SeedQR" might belong non-translated too?

I think I like "ABC" untranslated for as long as multiple keyboards are not offered.

src/krux/pages/qr_view.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, ctx, binary=False, data=None, title=None):
                 self.title = t("Compact SeedQR")
                 self.code = self._binary_seed_qr()
             else:
-                self.title = t("SeedQR")
+                self.title = "SeedQR"
```
jdlcdl, line 64: no translations for "SeedQR", but translations for "Compact SeedQR"?

src/krux/pages/login.py
```diff
@@ -111,7 +111,7 @@ def load_key_from_manual_input(self):
             [
                 (t("Words"), self.load_key_from_text),
                 (t("Word Numbers"), self.pre_load_key_from_digits),
-                (t("Tiny Seed (Bits)"), self.load_key_from_tiny_seed),
+                ("Tiny Seed (Bits)", self.load_key_from_tiny_seed),
```
jdlcdl, line 114: no translations for "Tiny Seed (Bits)" but translations for "Stackbit 1248"?

odudex:
My take on translations is: When in doubt, not use them. Ex: Not use them for names, BIPs. I consider Bitcoin, Tiny Seed, SeedQR etc. to be names. At least in Brazil we won't translate those. The "Compact" Compact SeedQR, is questionable, but my vote is for to be considered a name too, so not be translated. Less translations require less maintenance e less room in flash. But off course, I'm totally open for contrary arguments!

src/krux/pages/qr_view.py
tadeubas: Good point! I didn't changed because a lot of languages translated this In portuguese "compact" means "compacto", and it makes sense to me. What do you think about @odudex ?

src/krux/pages/login.py
tadeubas: Wow, nice catch. I missed it! Will commit 😉

PR #353, opened 2024-04-06
Wallet customization: Account "Number" becomes "Index"

jdlcdl:
### Description

Only one ux change when customizing wallet:
* "Account Number" becomes "Account Index" (since the count starts at 0)

Many lines of code changes (no functional change) to clarify and/or simplify:
* `account_number` names simplified to `account` mostly for local variable names when not confusing.
* `account_number` changed to `account_index` where it needed to be explicit, because `account` is already an embit key.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other
Just a picky edit towards naming conventions: so that 
* "number" specifies counting from 1 while a behind-the-scenes "number-1" is likely taking place, and 
* "index" specifies counting from 0, just like at the end of the derivation path.

Note: existing naming conventions already use "_number" for ints that start at 1 and "_index" for those that start at 0 (except for cnc code which uses "_idx").

### DIFF
--- a/src/krux/key.py
+++ b/src/krux/key.py
@@ -45,18 +45,18 @@ class Key:
         multisig,
         network=NETWORKS["test"],
         passphrase="",
-        account_number=0,
+        account=0,
     ):
         self.mnemonic = mnemonic
         self.multisig = multisig
         self.network = network
-        self.account_number = account_number
+        self.account_index = account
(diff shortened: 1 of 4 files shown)


odudex:
Thank you Jean!

PR #357, opened 2024-04-08
Multisig "Account"s for wallet_customization

jdlcdl:
### Description
Adds multisig "account" support to wallet_customization branch.

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/src/krux/key.py
+++ b/src/krux/key.py
@@ -32,7 +32,7 @@ from embit.wordlists.bip39 import WORDLIST
 from embit.networks import NETWORKS
 
 DER_SINGLE = "m/84h/%dh/%dh"
-DER_MULTI = "m/48h/%dh/0h/2h"
+DER_MULTI = "m/48h/%dh/%dh/2h"
 HARDENED_STR_REPLACE = "'"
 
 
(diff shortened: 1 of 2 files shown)


jdlcdl:
> What you mean with "until PSBT validation improvements"?

During testing, I realized that krux can already sign multisig wallets from custom derivations, just so long as it gets far enough to call PSBTParser.sign() method.  This is because the sign operation is done with the master root xprv, rather than being signed by the exact wallet that is loaded in krux and was loaded by the wallet_descriptor.  This is convenient for signing because "if it can sign, then krux should have a way to sign it." for script types that perhaps krux doesnt understand yet (Im thinking miniscript).  However, now that multisig accounts will be supported, I believe that if a psbt is loaded for a wallet 1) other than exactly the same as the wallet-descriptor or 2) where one of the signers is exactly the same as the current wallet loaded in krux... that validation should raise an error.   I'm still looking into, and trying to understand the PSBTParser class.

As well, I'll take a look at validation when an xpub or wallet descriptor is loaded, and make sure that everything that can be checked is being checked.  Currently that's the xpub only for multisig and I don't know about single-sig.  Maybe krux can do more validation, checking xpub, derivation path, and root fingerprint (though if the root fingerprint doesnt match, signing will be impossible).

PR #360, opened 2024-04-10
Turkish Translation Improvement

plebbitor:
Some words (8) that did not sound natural were changed

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ X] Other


odudex:
How about this:
`"Extended Public Key": "Genişletilmiş Public Key",`
I found a little weird that it is a single term, but it is partially translated. Is this how you usually call it?

PR #361, opened 2024-04-11
path_mismatch refactor for every input, every pubkey, including multisig

jdlcdl:
### Description
If it's good for single-sig, why not for multisig too?
* checks all inputs (instead of first only)
* checks for all xpubs (which seems fine for sparrow but must test other coordinators),
* might return more than a single path, comma separated
* tested for up to 21 inputs (but many inputs and multisigs of high "n" might get out of hand)

As of Friday April 12th, testing has succeeded for up to 400 path-mismatch checks on sparrow, specter-desktop and nunchuk.  (via a 6input to 1output 9of16 multisig psbt of size 17724 base64 chars).

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other

### DIFF
--- a/src/krux/psbt.py
+++ b/src/krux/psbt.py
@@ -102,22 +102,22 @@ class PSBTSigner:
 
     def path_mismatch(self):
         """Verifies if the PSBT path matches wallet's derivation path"""
-        if not is_multisig(self.policy):
-            # Single-sig check
-            first_input = self.psbt.inputs[0]
-            first_pubkey = next(iter(first_input.bip32_derivations))
-            # Get the derivation path for the first public key
-            derivation_path = first_input.bip32_derivations[first_pubkey].derivation
-            textual_path = "m"
-            for index in derivation_path[:3]:
-                if index >= 2**31:
-                    textual_path += "/{}h".format(index - 2**31)
-                else:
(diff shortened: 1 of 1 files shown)


odudex:
My only doubt is if it's necessary to loop through all inputs. I'm doing the same to pick the fingerprints, probably what's valid for one is valid for the other too.

jdlcdl:
I suspect this branch is doing more looping and more checking than is normally necessary (from a single wallet that prepares input-spends, they'll all have the same derivation path as the first).

That these extra loops remain:
* can only be useful if the extra checks save us someday when they're actually helpful
* doesn't seem to be causing a performance bottleneck (at least one that matters much).  The bottleneck seems to be how large the psbt is.  When a psbt is so large that it cannot finish scanning, it fails.  When it can be scanned, the check happens w/o failure from what I can tell, and the psbt can be reviewed and signed.

If the extra looping does not remain, and is instead only the first input or short-circuits on first mismatch, I do like the fact that this branch is checking both single-sig and multisig.

odudex:
From my early tests, I can also say this branch works with Taproot PSBTs, while original doesn't. I'll let you choose if we loop through all inputs or not. Both work, and we can revisit this in case any inconvenient shows up.

PR #371, opened 2024-04-25
"New Mnemonic" tweaks atop branch: wallet_customization

### LINKED ISSUE #370: [Enhancement] new mnemonic via "words", and help w/ the last word
#### new mnemonic manual "words" input + help w/ last word

Currently, there is no option to create a new mnemonic by inputting words manually,
However, user can recover an existing mnemonic this way, and if they don't select the last word, krux will randomly choose a valid last word.

This was discussed this morning in the main telegram group.

It may be confusing for users to create a new mnemonic (where they need krux to calculate the checksum and propose one) via the "load mnemonic menu".  Further, it is less than ideal to "lead" a user to a particular last word during recovery; it's better to let them enter it and learn that there is a checksum error (so that they verify all the other words too).

It may be better to have two modes... one much like the existing one in "Load Mnemonic" except it arguably might not offer any choice to leave it blank.

...and to have another mode in "New Mnemonic" where the user first selects for a 12-or-24-word mnemonic, and then as the 12th or 24th word is manually input... the list of available bip39 words is reduced to only those that could result in valid mnemonics.  This way, we'd be helping users to choose a last word ONLY during new mnemonic, and wouldn't be leading them towards one during recovery.  Further, they'll be in control of the choice for the remaining bits of entropy (selecting a word) rather than using krux's random choice.

---

Odudex referred to this code for what must be done to create the list of valid last-word candidates.
https://github.com/selfcustody/krux/blob/58d6e7332cc4e8073bbe5b66430bef401a20d436/src/krux/pages/tiny_seed.py#L396

---

I wonder if similar functionality isn't best placed within a method of embit.bip39.  I've continued to prototype what I shared previously (and since removed).  The original (expanded bits) and an improved version are now 2 comments below this post.

jdlcdl:
### Description

resolves #370

This pr intends to differentiate keyboard-entry of mnemonic words by purpose:
* "New mnemonic": where user is led towards a valid last word,
* "Load mnemonic": where user is NOT "led" towards a valid last word, because they should know it.
...currently these separate actions are combined within the "Load mnemonic" menu.

Functionally:
* There is no longer the ability to create a new mnemonic via Numbers: Decimal, Hex or Octal where the last word is picked by krux (or guided via reduced keyboard).  Can still load via these methods, must know the exact mnemonic.
* New method `Key.get_final_word_candidates()` returns a list of valid final words.
* Existing method `Key.pick_final_word()` uses above instead of randomly picking and checking for validity.
* New reusable func in `__init__.choose_len_mnemonic()` returns user choice of 12, 24, or None if "Back".
* Existing `.load_key*()` methods have an optional `new` param (defaulted False) for reuse in "New Mnemonic" mode.
* For New Mnemonic, users can select "Via Words", then "12 or 24 words".  They will be warned before the last word that they can leave it blank... else their keyboard will be limited to lead them towards a valid mnemonic.
* For Load Mnemonic, users won't have any warning before 12th or 24th word.  They cannot leave it blank and will not be led via a reduced keyboard to a valid final word; they must know their exact mnemonic and learn later if it is valid or not.  They will be prompted with "Done?" after the 12th word.

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Other


jdlcdl:
While adjusting tests, I realized that loading keys by number: decimal, hex, and octal -- all allowed for deriving the final word.  That is, because the tests expressly tested this (via test_possible_letters_from* tests), it seems that it was intentional that this functionality exists.  I even began to implement similar functionality as the New Mnemonic via Words for these methods, but then asked myself "Why would anyone create a new mnemonic via these methods?" and settled on "They wouldn't, they may have recorded their complete mnemonic in this manner, but they would have known the full values which covered checksum bits... therefore these methods are only available for Load, and not for New.  Please correct me if I'm missing something, because this functionality has been removed for now.

jdlcdl:
Baring requests for changes to this pr, I am done with code commits.  Exceptions are:
* more tests may be added (but I also have some thoughts to share on the value of end-to-end flow tests while ui improvements remain in the near future.
* I do intend to review documentation and add docs commits for changes related to this.

src/krux/pages/home_pages/bip85.py
odudex, line 1: Is it necessary to import  the `choose_len_mnemonic` function on pages which are child from `Page` class?

src/krux/pages/__init__.py
odudex, line 1: This is great!

src/krux/pages/home_pages/bip85.py
jdlcdl: I believe it's necessary, but only because it's a stand-alone function and not a method of Page.  If it were a method of page, wouldn't need to pass ctx too.

I wasn't sure how this reusable function would be received by others, but I added it to keep my own coding errors minimized since I thought I might be adding it in a few different places (especially since I thought I'd be implementing new-mnemonic for numbers in dec/hex/oct... and before I opined that this wouldn't make as much sense as just letting user create new via words).

odudex: The function is and its reuse great! I believe it could be part of `Page`

jdlcdl: Can I humbly beg NOT to add it as a method to Page just yet.

It's a new pattern to have a one-liner common menu choice, and it might very well fit into Page eventually, but as I look at that class, it's currently so clean and abstract that it feels premature and dirty to add this very particular menu to it.  Leaving it as is, for now, has the benefit/cost of an extra import per module that uses it, so it will be easy to see which modules truly need it (certainly not all of them do).  The extra cost of having to pass ctx into the func is not too high, and if the time comes that we find the pattern useful (to easily include common menus that always look/act exactly the same), then we'll learn over time which code pattern seems best.

odudex: All right, I'm convinced :)

odudex:
Thank you!

PR #374, opened 2024-04-28
Psbt Warning Path mismatch txt change

tadeubas:
- Merged main branch
- Warning: Sign PSBT Path mismatch now shows wallet path for comparison
- ~~Check whether the current key can sign the PSBT in advance to avoid late errors~~

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other


odudex:
About "Check if can sign PSBT in advance to avoid late error"
I believe we shouldn't sign before user agrees to do so, independently if the signature will be exported or not.
As we talked before, I'm in favor of give a little bit more details on why some PSBT eventually can't be signed.
What we could do in advance is to run a specific test, like detect xpub or fingerprint mismatch. But it would still be redundant, and as this errors usually represent a coordinator setup problem, it won't be a recurring waste of time for users, once they fix their coordinator they won't have the issue again.

jdlcdl:
I agree with Odudex above on not signing the psbt (even if krux is careful not to export it) until after the user has confirmed.  I'm imagining 2 scenarios, one unlikely and the other very likely:
* a future attack where electro-magnetic waves of signature are picked up by hacked coordinator who altered the psbt and will publish the tx whether or not the user agrees.
* any user who can read the code and raises concern for something like above or just based on principal alone: "Security over ease-of-use... Why is krux signing anything without my explicit approval?".

If user has loaded a wallet descriptor, then one of the initial checks are that the wallet descriptor matches the current wallet in krux, and also that the exact xpub can be derived via krux+wallet which is a guarantee that private keys can be derived in order to sign (so besides strange signing incompatibility or unsupported schemes, all psbts could be signed).

When user doesn't load a wallet descriptor, the derivation-path warning is helpful, but not full proof for the purpose of knowing that the psbt can be signed.

An alternate check might be to do as happens with a wallet-descriptor...  can krux derive the exact xpubs needed to sign this psbt?... and fail early  if the answer is no.  I like tadeu's thought of counting how many sigs can be added.  Don't need to continue the process if krux cannot add additional sigs... unless there is a use case that maybe someone wants to verify that krux will create exactly the same sigs??? (might that be an anti-exfil mitigation usecase?)

tadeubas:
Thanks for the comments guys! I was just thinking from the user's perspective, know upfront that this psbt can be signed rather than running the user through the entire process... if there was a way to do this without signing I would choose that option, but I don't believe there is...

I will remove this line and update the PR

jdlcdl:
Now Im rethinking everything.

To have derived the master xprv in krux... and to somehow leak that (via imaginary or not risk of waves), would be every bit as bad or worse than leaking a particular signature.

As long as the sig-op can be done w/o affecting the psbt structure, and quickly forgotten before the function returns, Im not sure there is any extra harm that is not also possible with a creative imagination knowing that the master secret is already known to krux device.

odudex:
I believe your second thought, about signing procedure and code, is relevant enough.

PR #378, opened 2024-05-01
proposal for pip install poetry

j-moreno-c-r:
the idea is create a python venv and inside import the poetry, in this way we avoid conflict problems enter the local pip and the poetry and the pythons...


README.md
```diff
@@ -69,7 +69,21 @@ The Krux code is a Python package that should be installed with [Poetry](https:/
 pip install poetry
 poetry install
 ```
-
+There may be a problem at this stage when trying to directly install Poetry, we have a solution proposal(this comand is for unix system mac or linux):
```
tadeubas, line 72: I would only simplify this line to:
If you have a problem installing Poetry on Linux OS try:

j-moreno-c-r: Ok good idea

j-moreno-c-r: better?

tadeubas: awesome! Now just wait for @odudex to merge the PR 😉

tadeubas:
@j-moreno-c-r "This branch has conflicts that must be resolved", you need to update your branch, I the README a short time ago and as you've changed it too it is caused a conflict to git. You'll have to resolve the conflict locally and then update your branch, then the PR here will be automatically updated

PR #379, opened 2024-05-03
Small changes to README, Docs, poe flash and krux script

tadeubas:
- krux script more checks on build > device name
- poetry fix command flash to accept -p as argument for yahboom
- poetry poe new task `git-pull-https`
- README info about live Python REPL in device's terminal
- Docs adding more info about yahboom flash process

- [ ] Bug fix
- [ ] New Feature
- [X] Documentation update
- [X] Other


krux
odudex, line 1: What does this change do? How that ```poetry run``` does in that ```if``` in line 59?

tadeubas: The change to poetry is in the [pyproject.toml](https://github.com/selfcustody/krux/pull/379/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711)

The if in Krux script is just another check of the device name, to see if it is correct. This avoids running docker with a wrong name that will fail later. Example: If you execute`./krux build maixpy-yahboom` it would fail after some time because the correct name uses underline.

odudex: Screenshot_20240505_081914_Chrome
I still didn't get this on krux script

tadeubas: wow, a typo!

odudex:
Thank you!

PR #381, opened 2024-05-04
Battery display as percentage

tadeubas:
- Battery display as percentage or as an icon
- Other small comments and code refactor

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


odudex:
Currently battery will be limited to 90% unless charge is completed, then it will be 100%. This is done for the user to be able to know when charge is completed.
This is OK with a "progress bar" of a few pixels, but with numbers it will be weird the see the percentage rising with 1% precision, then jump from 90 to 99%.
Why 99%? Will users assume charge is completed with 99%?
How will this behave the moment power is connected? The voltage immediately rises (with an equivalent of more than 30% charge difference). How will this compensation work when displaying charge with 1% precision? How will this behave in different devices, with different battery ages? Tests should be done.

The method we use is not precise, it's no neat to show more precision that we can offer. I would suggest changing from this:
```charge = min(int(charge * 100), 99)```
to something like
```charge = int(charge * 10) * 10```
To show 0%, 10%, 20% ...100%

Finally, is there room for "100%"?

tadeubas:
Yes, changes can be made, there is no problem with that, I just want users to be able to see the battery as %, I know it won't be perfect but the icon isn't perfect either. I don't think we have room for 100% in m5stickv, so I think it would be strange to allow Amigo to display 100% but in m5stickv only 99%, but that can be changed too!

PR #382, opened 2024-05-04
proposal of solution for poetry installation

j-moreno-c-r:

- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other


odudex:
We typically avoid delving into details about installing dependencies since they frequently change. However, because issues with installing Poetry are recurring, addressing this can be very beneficial. Thank you, João!

PR #383, opened 2024-05-06
Sign PSBT without fingerprint if wallet can sign

tadeubas:
Sign PSBT without fingerprint if wallet can sign. This is the correct way to fix this [issue from Feb 2023](https://github.com/diybitcoinhardware/embit/issues/41) that still affects SeedSigner as seen [here](https://github.com/SeedSigner/seedsigner/issues/359)

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


odudex:
I do not consider the option to require a matching key fingerprint in PSBTs to be a bug, but I am open to discussion.
Please elaborate on why you consider this to be a bug.

In my opinion, redundancies like fingerprints, checksums, etc., are an important part of protocols and help to avoid problems and mistakes.
I am in favor of helping users diagnose issues and properly set up their coordinators, such as having a more detailed error message in case the PSBT fingerprint does not match the signer fingerprint. But to ignore mismatching fingerprints, I would like to hear more arguments from more people.

jdlcdl:
off the top of my head thoughts:

A missing fingerprint is different than a fingerprint mismatch.  If fingerprint is 0x00 * 4, then it wouldn't be wrong to consider it missing/unset.  I like that the first commit here is checking for that and not filling fingerprints that are otherwise 'set'.  I think it would be nice to inform the user that this has happened: "Missing fingerprint in PSBT ignored." 

Im not convinced that this pr, as it is currently, will work for similar situations with taproot (because I think there are other psbt fields to look for) but Im not sure it won't work either.  TODO: test similar case for taproot wallet and psbt.

PR #388, opened 2024-05-12
Wallet sans key

jdlcdl:
### Description
A Wallet Descriptor can be loaded without the user loading private key material... to verify/view addresses.

* A wallet can be loaded with key=None (ie: krux is in watch-only mode... still in login())
* wallet has a notion of "persisted" so that redundant offers of "save to sdcard" can be reduced.
* pages that, currently -- arguably, should ONLY be available once user has logged in, are being re-used prior to login.
* prevailing network will be the one defined by the user with a loaded mnemonic and wallet... else will be defined by the wallet descriptor which is loaded sans key.
* Using tools/descriptor-addresses to load a wallet descriptor before logging in does NOT interfere with loading a mnemonic.  A new wallet will be initialized w/o the previously loaded descriptor.
* AssumptionWarning (a new exception) will be raised if assumptions must be made building a descriptor (the case with BlueWallet's vague legacy p2pkh, an xpub without key-origin; if raised, user should decline to accept the assumption of bip84 native segwit; they can instead use an xpub or descriptor directly from krux for checking legacy addresses.  For bluewallet ypub/zpub, this is not the case because of their specific "versions".

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


odudex:
This looks great!
About the network, as any other path attribute, IMO it should use descriptor's.
Maybe descriptor path attributes could be displayed in an "infobox" like it's done with wallets before user enters addresses menu.

tadeubas:
I was testing the feature and I think the infobox could appear on the address menu, no need for a interaction or a new screen "before", also if the user is using multiple devices it can forget what is loaded on what and the infobox appearing on the address menu would solve that issue

src/krux/pages/home_pages/wallet_descriptor.py
```diff
@@ -39,41 +39,41 @@ def __init__(self, ctx):
...
+        title = t("Wallet output descriptor")
         self.ctx.display.clear()
         if not self.ctx.wallet.is_loaded():
             text = t("Wallet output descriptor not found.")
-            if not self.ctx.wallet.is_multisig():
-                text += " " + t("It is optional for single-sig.")
             self.ctx.display.draw_centered_text(text)
             if self.prompt(t("Load one?"), BOTTOM_PROMPT_LINE):
-                return self._load_wallet()
+                self._load_wallet()
```
tadeubas, line 49: removing the return here introduced a bug alongside with the other change on the line with the comment `# Try to save the Wallet output descriptor on the SD card`

src/krux/pages/home_pages/wallet_descriptor.py
```diff
@@ -39,41 +39,41 @@ def __init__(self, ctx):
...
-                save_page = SaveFile(self.ctx)
-                save_page.save_file(
-                    self.ctx.wallet.descriptor.to_string(),
-                    self.ctx.wallet.label,
-                    self.ctx.wallet.label,
-                    title + ":",
-                    DESCRIPTOR_FILE_EXTENSION,
-                    save_as_binary=False,
-                )
+        # Try to save the Wallet output descriptor on the SD card
```
tadeubas, line 58: removing this from the else clause introduced a bug alongside with the other change on the line 49 ` self._load_wallet()`

src/boot.py
```diff
@@ -82,7 +82,7 @@ def home(ctx_home):
     """Loads and run the Login page"""
     from krux.pages.home_pages.home import Home
 
-    if ctx_home.wallet is not None:
+    if ctx_home.wallet is not None and ctx_home.wallet.key:
```
tadeubas, line 85: This new form to check if a wallet is loaded would be better as a helper function in the context class

src/krux/pages/__init__.py
```diff
@@ -726,7 +726,7 @@ def draw_battery_indicator(self):
 
     def draw_wallet_indicator(self):
         """Draws wallet fingerprint or BIP85 child at top if wallet is loaded"""
-        if self.ctx.wallet is not None:
+        if self.ctx.wallet is not None and self.ctx.wallet.key:
```
tadeubas, line 729: this could use the helper function in context class

src/krux/pages/__init__.py
```diff
@@ -747,6 +747,7 @@ def draw_network_indicator(self):
         """Draws test at top if testnet is enabled"""
         if (
             self.ctx.wallet is not None
+            and self.ctx.wallet.key
```
tadeubas, line 750: this could use the helper function in context class

src/krux/pages/home_pages/wallet_descriptor.py
tadeubas: The bug is, if you don't load the wallet output descriptor or load an with the error `Invalid wallet:`, it will still asks the user to save the file on the SD card

src/krux/wallet.py
```diff
@@ -58,27 +60,38 @@ def __init__(self, key):
...
-                raise ValueError("xpub not a cosigner")
+        if self.key:
+            network = self.key.network
         else:
-            if not descriptor.key:
-                if len(descriptor.keys) > 1:
-                    raise ValueError("not single-sig")
-            if self.key.xpub() != descriptor.key.key.to_base58():
-                raise ValueError("xpub does not match")
+            network = NETWORKS[Settings().wallet.network]
```
tadeubas, line 78: What happens if the user customize the wallet to main/test with this line? The Settings only serves for default now, not for the current loaded wallet.

tadeubas: I see, this will only applies for when loading only the descriptor right? So why not use the descriptor derivation path info?

src/krux/wallet.py
```diff
@@ -112,9 +125,13 @@ def obtain_addresses(self, i=0, limit=None, branch_index=0):
         """Returns an iterator deriving addresses (default branch_index is receive)
         for the wallet up to the provided limit"""
         starting_index = i
+        if self.key:
+            network = self.key.network
+        else:
+            network = NETWORKS[Settings().wallet.network]
```
tadeubas, line 131: Here again, why not use the descriptor derivation path info? It would serve for the two cases, when loading the wallet and when loading only the descriptor

src/krux/pages/tools.py
```diff
@@ -173,3 +174,16 @@ def create_qr(self):
...
             return seed_qr_view.display_qr(allow_export=True)
         return MENU_CONTINUE
+
+    def wallet_addresses(self):
+        """Handler for the 'Wallet Addresses' menu item"""
+        from .home_pages.wallet_descriptor import WalletDescriptor
+        from .home_pages.addresses import Addresses
+        from krux.wallet import Wallet
+
+        if not self.ctx.wallet:
```
jdlcdl, line 184: note: Tadeu and I discussed that preserving the wallet while not yet logged in (in tools) doesn't feel right, this line to be removed so that reviewing full descriptor and qrcode is not possible until actually logged in with a real HDKey.

src/boot.py
jdlcdl: Context method ".is_logged_in()" is the name I have in mind for this.

src/krux/pages/home_pages/wallet_descriptor.py
jdlcdl: Thank you!

src/krux/pages/tools.py
```diff
@@ -51,6 +51,7 @@ def __init__(self, ctx):
                     (t("Check SD Card"), self.sd_check),
                     (t("Print Test QR"), self.print_test),
                     (t("Create QR Code"), self.create_qr),
+                    (t("Wallet Addresses"), self.wallet_addresses),
```
jdlcdl, line 54: note: Tadeu and I discussed a renaming of this to  "Watch Only"... which is the mode a user will have in mind when using tools to review addresses (even if krux cannot watch anything, it seems as good a name as any other to me too).

src/krux/wallet.py
jdlcdl: While there are preferred derivation path standards which differ based on the network, I feel the path is a much softer hint of network compared to the version bytes on the front of the xpub.  It really feels like embit.descriptor deserves to have a "network" attribute, else possibly krux.wallet needs one... but for now I just want to get it working for single-sig and multisig address listings.

As a trial, I'll use the first network I find in the order of:
1) whatever the krux user loaded their mnemonic with (which comes from settings but they may have changed it),
2) the first xpub in the descriptor, via the version bytes and a value-to-key lookup of embit.networks.NETWORKS
3) if for some reason it's not found, but I cannot imagine why, the network from settings.

...but if krux.wallet had a network attribute that got filled in by parsing the descriptor or by the user when they load the mnemonic, it would be much cleaner and faster for scanning many pages of addresses based on descriptor alone... and could further serve to fail/warn when descriptor network and krux user's choice don't match... with the downside (I'll argue upside) of not being able to load a mainnet xpub while setup for testnet and seeing testnet addresses instead of an error "descriptor is not for this network".

odudex: In my opinion, this feature should focus on the descriptor only. That's why I would call it "Descriptor Addresses".
Krux will use data within the descriptor to export addresses.
Incomplete descriptors, without enough data to derive addresses, should not be allowed.
Assembling wallets from pieces of information from multiple sources doesn't seem like a good idea to me.

odudex: Initially, I was only considering the new feature in isolation. I hadn't realized that you were also addressing the scenario of descriptor loading with an already loaded key, and how these two situations are naturally connected in code. Your proposed approach seems ideal to me.

jdlcdl:
Attached: a compressed tar file that will create a "./sd-wallet_sans_key/" directory wherever it is unpacked.  This directory holds:
* seeds.json w/ 3 mnemonics w/o passphrases for, "Grampa", "Grampa85c0" and "Grampa85c1".  The key is "abc".
* settings.json for settings storage on sdcard and user-interface in english,
* MAINNET directory with subdirectories for krux-supported script types, each containing usefully-named text-files that can be loaded in this branch as wallet descriptors for verifying/viewing addresses.
* TESTNET directory with same as above, for testnet.

You may put the contents of "./sd-wallet_sans_key/" onto an sdcard for testing on a krux device, or into the "~/krux/simulator/sd" directory for testing via the simulator.

The following gist was used in testing this branch and creating these files.
https://gist.github.com/jdlcdl/8675f0f1e748d06274d31c27cc7474d0

[sd-wallet_sans_key.tar.gz](https://github.com/selfcustody/krux/files/15503304/sd-wallet_sans_key.tar.gz)

jdlcdl:
Commit [ca84389](https://github.com/selfcustody/krux/pull/388/commits/ca84389b65ea6cf28a71502a95e36208439fad82) corrects a bug found by Odudex where bech32 addresses were incorrect for testnet (they were regtest bech32 addresses instead) when derived from a wallet descriptor only.

It was about an assumption made in wallet.which_network() that different implementations of python might sort dictionary keys similarly (main, test, signet, regtest), but on k210, "regtest" came first and this branch was unable to determine between testnet/signet/regtest based on the extended-public-key version alone, assuming regtest gave "bcrt" addresses instead of "tb" addresses.

The fix includes:
* what appears to be more looping but it's not, it's just slightly easier to read w/o comprehensions,
* a hard coded list of supported embit.networks.NETWORKS keys ("main", "test")
* a hard coded list of extended-public-key version keys ("xpub", "zpub", "ypub", "Zpub", "Ypub")

This bug would have appeared only for testnet, only for bech32 addresses, and the wrong addresses would not have been valid for the wrong network.  Still, my sloppy mistake is solely mine.

Thank you Odudex!

src/krux/pages/home_pages/wallet_descriptor.py
```diff
@@ -39,41 +39,41 @@ def __init__(self, ctx):
...
                     title + ":",
                     DESCRIPTOR_FILE_EXTENSION,
                     save_as_binary=False,
                 )
+
         return MENU_CONTINUE
 
     def _load_wallet(self):
         wallet_data, qr_format = self.capture_qr_code()
+        persisted = None
```
tadeubas, line 76: I would change to `persisted = False`

src/krux/wallet.py
```diff
@@ -36,7 +49,9 @@ def __init__(self, key):
         self.descriptor = None
         self.label = None
         self.policy = None
-        if not self.key.multisig:
+        self.persisted = None
```
tadeubas, line 52: I would change this initialization to `self.persisted = False`

src/krux/pages/home_pages/wallet_descriptor.py
```diff
@@ -84,44 +84,49 @@ def _load_wallet(self):
                 _, wallet_data = utils.load_file(
                     (DESCRIPTOR_FILE_EXTENSION, JSON_FILE_EXTENSION)
                 )
+                persisted = True
             except OSError:
                 pass
+        else:
```
tadeubas, line 90: If you made the above change to initialize `persisted as False`, you can remove this `else:`

tadeubas (commented):
Hi @jdlcdl, amazing work! I just made a few suggestions because it doesn't make sense for a Boolean variable to take `None` as a value. It was either persisted or NOT (if it was created recently, it was NOT)

jdlcdl:
> Hi @jdlcdl, amazing work! I just made a few suggestions because it doesn't make sense for a Boolean variable to take `None` as a value. It was either persisted or NOT (if it was created recently, it was NOT)

I was using 3 values for: as None=i don't know if it's persisted or not, False=doesn't appear to be persisted, and True=Im sure it's already persisted.

But None and False are effectively the same: "I'm not sure that it's persisted" (because even if we set it False, it could be persisted as another file on sdcard or elsewhere... and we'll set it True if/when we load or save it.

Thank you Tadeu, I'll issue a pr for your suggestions.

src/krux/pages/home_pages/wallet_descriptor.py
jdlcdl: I believe that odudex already caught this when breaking-up the loading of descriptors from in-serial (via camera, then from sdcard) to parallel choices of how to load.

PR #390, opened 2024-05-14
PSBT preview now shows fee in percentage of the value spent

tadeubas:
- Added % to fee in psbt preview
- Added warning to psbt when fees greater than 30%

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other


src/krux/pages/home_pages/home.py
odudex, line 1: Once you have the fee calculated, you could show it in the warning message, ex:
```t("Warning:") + "\n" + t("High fees!") + "\n" + t("%s% of spent") % fee_percent```

src/krux/psbt.py
odudex, line 1: you could optimize calc and remove the dependency of `math` here with:
```fee_percent = ((fee * 1000 // spend_amount) + 9) // 10 / 10```

src/krux/format.py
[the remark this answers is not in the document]
tadeubas: I didn't know that, I'm going to change it... in the future we will have to change other functions that already have `get_` in their name too 😆

src/krux/psbt.py
tadeubas: I made the change, but I think it's unreadable now, before any dev knew what was going on and we already use the math library in several places (including for the ceil function)

jdlcdl: Is this all about round() not giving what a user would expect? `fee_pct = round(fee / spend * 100, 1)` is so much easier to read... but round(0.5) is 0, so I can see reason in displaying the ceiling to be conservative.

jdlcdl:
This is working for me and I like it, including the hardcoded warning that might someday become a preference... but 30% seems sane to me.

I note that self transfers wont show any fee pct... but that's just because there is no spend.  it's handled well (no zerodivision or name errors).

tadeubas:
Thank you @jdlcdl and @odudex for the review and tips! Any other suggestions or changes feel free to submit a PR to my branch or make another comment on this PR 😉

src/krux/psbt.py
odudex: It looks readable to me, but you could add and easier alternative as a comment as an explanation :)
I've reduced the number of math occurrences in the code, although it's still used in several places. This change is part of the many optimizations that have been enhancing Krux's performance. Coming from a background in programming 8-bit microcontrollers, I tend to favor more primitive and constrained equations.

odudex:
Thank you!

PR #395, opened 2024-05-19
Sats/vB in PSBT preview

tadeubas:
- PSBT now shows a good estimation for ~sat/vB

- [ ] Bug fix
- [X] New Feature
- [ ] Documentation update
- [ ] Other


odudex:
If the resource cost were lower, I would think: Why not?
However, given that the total fee can already be compared with the coordinator, is it justified to add seven functions and 170 lines of code to recalculate this indicator?

tadeubas:
It's actually not much, most of the lines are tests and comments, the logic itself is simple, just consts and a little math (multiplications and additions), it requires almost nothing from the device's CPU or memory and creates confidence to the user that the coordinator and the HW are working as expected. Additionally, some other users already want this feature.

tadeubas:
I was thinking about what you said about the "resource cost" to be "justified to add seven functions and 170 lines". Opinions are just opinions, and lines and functions are just compiled... to me the real measurable cost is just the size of the firmware (I think we can't even measure the delay on the calculation, if any nano seconds exists).

firmware.bin size for yahboom, from `wallet_customization` branch:
1.944.064 bytes

firmware.bin size for yahboom, from `satsvb` branch:
1.946.112 bytes

Is this all the cost you mean? The real point is, I've added 2.048 bytes of code with this feature. If we come to a limit and need to remove things to make firmware smaller, we would have much more unnecessary things to remove that really impact the size. For example: Why not remove the stackbit 1248 that is something just for Brazilians (just the `stack_1248.py` has 463 lines of code)? Why not remove Tiny Seed also? I think these features add much more "resource cost", "functions and lines"... we could also check if we have users that really uses all the translations we current have and so on...

odudex:
> I was thinking about what you said about the "resource cost" to be "justified to add seven functions and 170 lines". Opinions are just opinions, and lines and functions are just compiled... to me the real measurable cost is just the size of the firmware (I think we can't even measure the delay on the calculation, if any nano seconds exists).
> 
> firmware.bin size for yahboom, from `wallet_customization` branch: 1.944.064 bytes
> 
> firmware.bin size for yahboom, from `satsvb` branch: 1.946.112 bytes
> 
> Is this all the cost you mean? The real point is, I've added 2.048 bytes of code with this feature. If we come to a limit and need to remove things to make firmware smaller, we would have much more unnecessary things to remove that really impact the size. For example: Why not remove the stackbit 1248 that is something just for Brazilians (just the `stack_1248.py` has 463 lines of code)? Why not remove Tiny Seed also? I think these features add much more "resource cost", "functions and lines"... we could also check if we have users that really uses all the translations we current have and so on...

Fair enough! Let's add it.
Please make it fully isolated, and loaded on demand as the modules you mentioned. Also, please make constants upper case and create constants for magic numbers.

tadeubas:
Hey @odudex , I think the sats_vb is now fully isolated, and loaded on demand. If I miss something plz let me know 😉

odudex:
Great!
Can I ask one last thing? Can you pull recent changes, extend and adapt the Sats/vB tests to also cover the recently added Legacy and Taproot tests?

tadeubas:
Sure, will do ASAP

odudex:
Was going to run RAM tests on this, but it failed to sign even small transactions. Seems to have a bug:
`Error: KeyError('p2wpkh',)`

tadeubas:
> Was going to run RAM tests on this, but it failed to sign even small transactions. Seems to have a bug: `Error: KeyError('p2wpkh',)`

My bad! Now it is fixed. Thx 👊

jdlcdl:
ACK tested as of 0e95ae9

Tested on amigo for:
* 6 native segwit inputs to 5 outputs of assorted types,
* signing on utxo inputs of different types (legacy, native segwit, nested segwit, taproot, multisig native segwit)
and in all cases, pre-signature estimates on krux psbt review screen were within a fraction of a satoshi per vbyte compared to what sparrow/mempool.space reported for the finalized transaction.

PR #396, opened 2024-05-21
Docs: info about anti-glare and small changes

tadeubas:
- Added more info about anti-glare
- Added info about brightness control
- Other small changes

- [ ] Bug fix
- [ ] New Feature
- [X] Documentation update
- [ ] Other


src/krux/display.py
odudex, line 1: Line endings?

tadeubas: This is not a line ending issue, just that `pwm_value` is defined inside the `if` clause, so there is no point in using this value outside the `if`, this could cause an exception as `pwm_value` would not be defined

odudex: Ops thank you, I was reviewing on the phine and this was not visible.

PR #405, opened 2024-06-17
Show sorted directories first in SD filemanager

tadeubas:
- SD file listing now shows sorted folders first than files for better navigability
- Sorting ignores case (like when sorting a folder in Windows)

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other


odudex:
Thanks!

PR #408, opened 2024-06-22
Bugfix: load legacy PSBTs from SD card in compressed mode

odudex:
### Description

Noticed legacy PSBTs would fail to load from SD card with error: "Unable to get policy".
Reason seems to be, when loaded from file by Embit in compressed mode, the scriptpubkey from the UTXO of non-witness inputs is not retained, preventing the policy from being determined.
The use of compressed mode when loading from SD card is a recent change.
Compressed mode is not used when loading from QR codes.

To address this and other potential problems associated with PSBT verification and the use of Embit's compressed mode when loading PSBTs from SD card files, I have segregated the PSBT validation code. This allows for a second attempt at PSBT loading and verification, this time without employing compressed mode for loading.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Other


tadeubas:
Good catch 👏 ! It would be nice if we had a single test case using this legacy PSBT to ensure it will continue to work with future changes.

odudex:
PSBT alone did not suffice to replicate the problem; however, I have now added tests to encompass all the involved variables.

PR #420, opened 2024-07-16
Adds "< " to "Back" menu items

jdlcdl:
### Description

Perhaps with another "arrow" glyph, this will look nicer, but "< Back" is still very identifiable as a special menu item.

Also changed some tuples' 2nd item to the more standard "lambda: MENU_EXIT" instead of "lambda: None"

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other

### DIFF
--- a/src/krux/pages/__init__.py
+++ b/src/krux/pages/__init__.py
@@ -912,6 +912,6 @@ def choose_len_mnemonic(ctx):
     return num_words
 
 
-def cta_back(status=MENU_EXIT, tlabel="Back"):
+def cta_back(status=MENU_EXIT, label=t("Back")):
     """Reusable 'call-to-action: go back'.  Currently a menu item tuple"""
-    return ("< " + t(tlabel), lambda: status)
+    return ("< " + label, lambda: status)
--- a/src/krux/pages/__init__.py
+++ b/src/krux/pages/__init__.py
@@ -109,7 +109,7 @@ class Page:
                     t("Load from SD card"),
                     None if not self.has_sd_card() else lambda: None,
                 ),
-                ("< " + t("Back"), lambda: MENU_EXIT),
+                ("< " + t("Back"), lambda: None),
             ],
         )
         index, _ = load_menu.run_loop()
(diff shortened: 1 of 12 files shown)


odudex:
The `lambda: MENU_EXIT` didn't change behaviors, right?

tadeubas:
Hi @jdlcdl ! Can we create a constant for Back and reuse it everywhere?

src/krux/pages/__init__.py
```diff
@@ -109,7 +109,7 @@ def load_method(self):
                     t("Load from SD card"),
                     None if not self.has_sd_card() else lambda: None,
                 ),
-                (t("Back"), lambda: None),
+                ("< " + t("Back"), lambda: MENU_EXIT),
```
jdlcdl, line 112: 2nd item should return None to express that it's ignored.

src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -199,7 +199,7 @@ def sign_message(self):
                     t("Sign to SD card"),
                     None if not self.has_sd_card() else lambda: None,
                 ),
-                (t("Back"), lambda: None),
+                ("< " + t("Back"), lambda: MENU_EXIT),
```
jdlcdl, line 202: 2nd item should be None to express that it's ignored.

jdlcdl:
> The `lambda: MENU_EXIT` didn't change behaviors, right?

I was trying to standardize exiting menus with this MENU_EXIT status, but often it is ignored and doesn't matter.  I wasn't intending to alter behavior.  I'll move these back to the way they were because I'm not sure if they're inspected elsewhere.

jdlcdl:
> Hi @jdlcdl ! Can we create a constant for Back and reuse it everywhere?

Still not ready, but working in that direction.

latest commit tries to implement a cta_back() reusable in pages/__init__.py but to pass tests, couldn't use it in one place, and to pass i18n checks, had to feed it i18n food at least once.  Still a work in progress, and have not tested it on device (only limited testing in simulator so far).  It's leading me to believe that "Menu" class might take an optional cta_back param and set this up by default so that every menu doesn't always need it; then it would be easier to move from Menu item to Page later.

jdlcdl:
Altering to ready-for-review.

* i18n food kluge was fixed above, ty to tadeubas.

* assert_has_calls() was failing because cta_back() was returning a callable that returned MENU_EXIT, optionally could pass in a different status and it would return a callable that returns it) but this is not always the case, in the settings menu, we need to pass a callable to cta_back() which will be returned (instead of a callable that returns what was passed in).

odudex:
Reviewed, tested and merged. Thank you!

PR #421, opened 2024-07-16
Directories suffixed with "/" in file_manager

jdlcdl:
### Description

Suffixes directory names with "/" (display only) in the file_manager (also for the special "../" entry)

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other


src/krux/pages/file_manager.py
```diff
@@ -97,7 +97,10 @@ def select_file(
                         or SDHandler.dir_exists(path + "/" + filename)
```
tadeubas, line 97: Hi @jdlcdl ! Here at line 97 we check if it is a directory using `SDHandler.dir_exists(path + "/" + filename)`, then below on line 101 you made the check using `if filename in directories`. Can we create a variable `isDirectory` to reuse?

jdlcdl: Maybe it's best to do the os.stat directory check higher-up (if it's not a file), and then set isDirectory via "in directories" instead of doing another os.stat for filenames we already know are files?
Or maybe we can avoid an extra os_stat call, since it's in our os.listdir() output, can we assume it's a directory if it's not a file?

tadeubas: I don't know which is faster, I tend to think that `SDHandler.dir_exists(path + "/" + filename)` is faster than an `in list` check, because the list check is O(n), but the function dir_exists appears to be O(1). But I'm not worried about performance in this case, just code simplification and reuse.

odudex:
looks good to me!

jdlcdl:
I setup to test as of [7be968a](https://github.com/selfcustody/krux/pull/421/commits/7be968a6d607672a8592e66387837c707e75c9ab) via the simulator, w/:
* subdir_1 having 1 file, 1 subdir, 1 soft link, and 1 broken soft link,
* and did the same for subdir_10 and subdir_100 having 10 and 100 of each respectively.
Performance was great in the simulator, however broken links were visible leading me to believe that I should NOT assume that simply checking file_exists() was enough, I needed to check dir_exists() as well.

But as I tested the sipeed amigo, I realized that links (working, soft, broken) were not even supported on an sdcard formatted for krux.  Further, checking for existence of files and directories was drastically slower on the device than in the simulator.  The subdir with 100 files and 100 subdir's was taking around 2.7s.   I tried checking only for existence of files and if not assuming directory, as well as for existence of directory and if not assuming file... and performance was exactly the same.  100 files and 100 directories per subdir takes 1.2s in the latest commit above.

src/krux/pages/file_manager.py
```diff
@@ -78,8 +78,11 @@ def select_file(
 
                 del dir_files
 
-                # show sorted folders first than sorted files
-                for filename in directories + files:
+                # show sorted folders first then sorted files
+                for filename, is_directory in [(x, True) for x in directories] + [
```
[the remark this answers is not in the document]
jdlcdl: I'm understanding this as a memory efficiency difference:

While iterating over `[(x, True) for x in directories] + [(x, False) for x in files]:`, that expression will create a new list in memory as big as the other 2 existing lists (2x more memory than needed).

Whereas enumerate() will return an efficient iterator/generator that step through the 2 existing lists without needing to create a new one.  The extra step of setting is_directory is negligible because len(a_list) is not actually computing the number of elements in the list, it's returning the internal counter efficiently.

tadeubas:
Have you tested with a loooong folder name? I think it will remove the forward slash, you need to change the code below when filename doesn't fit on the screen, you need to change to use display_filename so the forward slash is taken into account

jdlcdl:
The 2 last commits are:
* a bug fix for long filenames, which was incomplete (shame on me)
* the bug fix to the above bug fix w/ tests that hopefully cover sorting of dirs before files as well as long filenames.

odudex:
Thank you!

PR #426, opened 2024-07-23
Descriptor urtype account

jdlcdl:
### Description
I'm not sure if this belongs, but I found this while trying to scan an xpub exported by seedsigner for sparrow, while testing/inspecting the seedsigner release candidate for v0.8.0.

Conveniently, all 4 of the coordinators supported by both seedsigner/krux deal well with the xpubs exported for sparrow by seedsigner, and they're often scanned with better success, especially with a bad webcam, than the xpubs exported specifically for that coordinator.  (For singlesig types, they're full descriptors with script wrapper and checksum, always with the mainnet "xpub" version bytes).  In coordinators that support testnet mode... they force the keys to testnet and the resulting import is a "tpub" (this is not the case for krux, because krux respects the xpub exactly in watch-only mode, and would show a key mismatch if a wallet were loaded for testnet).

My reasoning for this code, is simply to verify between my only two signers... but it's more realistic that bitcoiners will be verifying descriptors from their coordinator (which works flawlessly whether setup by krux or seedsigner)

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other

### DIFF
--- a/src/krux/wallet.py
+++ b/src/krux/wallet.py
@@ -207,6 +207,15 @@ def parse_wallet(wallet_data, allow_assumption=None):
         except:
             pass
 
+        # Try to parse as a Crypto-Account type
+        try:
+            account = urtypes.crypto.Account.from_cbor(
+                wallet_data.cbor
+            ).output_descriptors[0]
+            return Descriptor.from_string(account.descriptor()), None
+        except:
+            pass
+
         # Treat the UR as a generic UR bytes object and extract the data for further processing
         wallet_data = urtypes.Bytes.from_cbor(wallet_data.cbor).data
 


odudex:
If the modification enables Krux to accurately load a descriptor from SS and display the correct receive and change addresses, then we could certainly implement it.

tadeubas:
Nice addition Jean! Plz add a simple test to cover this use case and I think it is good to merge!

jdlcdl:
> If the modification enables Krux to accurately load a descriptor from SS and display the correct receive and change addresses, then we could certainly implement it.

Yes, that is the intention of this pr... but it will only work for single-sig and mainnet (because seedsigner is currently exporting for mainnet, and for multisig, like krux, will only export a fraction of the cosigners at a time.)

jdlcdl:
> Nice addition Jean! Plz add a simple test to cover this use case and I think it is good to merge!

Thank you for reviewing, tests added to check for UR:Output and UR:Account at the least.

Tests for UR:Bytes would also be nice (but I don't know of a coordinator or signer, yet, for that use case... even though the code does exist).

jdlcdl:
> I know Specter-DIY uses UR:Bytes somewhere, maybe for PSBT

It looks like the Specter Desktop wallet-descriptors are being qr-encoded as an alphanumeric dictionary, rather than UR:Bytes.   Whenever I stumble on a coordinator that is exporting UR:Bytes as their wallet backup, then I'll add tests for that.

We already have quite thorough UR:Bytes testing for PSBTs, it's in the ~/krux/tests/test_qr.py tests.

odudex:
Thank you!

PR #430, opened 2024-07-26
About shows board hardware

tadeubas:
Now about shows hardware alongside version

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other

### DIFF
--- a/src/krux/pages/login.py
+++ b/src/krux/pages/login.py
@@ -723,11 +723,16 @@ class Login(Page):
     def about(self):
         """Handler for the 'about' menu item"""
 
+        import board
         from ..metadata import VERSION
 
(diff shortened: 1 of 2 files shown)


jdlcdl:
This worked for me.

I imagine it's mostly useful for the case that a device was setup by another, handed to a newbie, and the new user wants to be sure that they'll reflash the correct firmware next time?

tadeubas:
Yes, it can be useful when the device is gifted to someone already with Krux installed, or to check if you have installed the correct firmware (some versions may partially work on different devices). If you have any idea of ​​any additional information that could be added, plz tell.

odudex:
Thank you!

PR #436, opened 2024-08-03
Uses grid_type to label TinySeed page ui

jdlcdl:
### Description

With my jaw-on-the-floor, while reviewing @3rdIteration's TinySeed improvements to support OneKey-KeyTag and Binary-grid formats, I noticed that once a grid is scanned and user has a chance to edit, that "Tiny Seed" as the title of the page might be confusing (even though krux is not confused about which grid_format to use).

This pr alters TinyScanner's binary_grid_settings for the default key from "TinySeed" to "Tiny Seed" and then passes the grid_type into TinySeed page so that ui/ux labelling can be appropriate for the current mode when editing.  note: Only the "Binary Grid" key is localized, while the other two keys retain their trademark english names.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other

### DIFF
--- a/src/krux/pages/login.py
+++ b/src/krux/pages/login.py
@@ -88,7 +88,7 @@ class Login(Page):
             self.ctx,
             [
                 (t("QR Code"), self.load_key_from_qr_code),
-                ("Tiny Seed", lambda: self.load_key_from_tiny_seed_image("TinySeed")),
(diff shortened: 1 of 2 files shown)


odudex:
Thank you!

PR #437, opened 2024-08-05
adjusted test to cover bugfix in commit 3eba4e4

jdlcdl:
Prior to commit 3eba4e4 and related to signing messages, if a message contained a ":" colon character, it was converted to a " " space character.  This pr adjusts the existing unit-test so that it would fail with prior versions and succeeds with 3eba4e4's fix.

User's who need to sign on a message using a previous release of krux might note the conversion of ":" to " ", which is visible in-device while signing. (ie: In Sparrow's "Sign/Verify Message" screen, they'd convert any ":" characters in the "message" text-area to " "... and then the krux signature would verify as valid) 

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other: Adjusted unit test


odudex:
Thank you Jean!

PR #442, opened 2024-08-12
embit remove unused  py_ripemd160.py

tadeubas:
removing from Docker as it is used only in Python, micropython uses `ripemd160` from `hashlib`. Also `ripemd160` is only used in embit for miniscript.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


odudex:
Thank you!

PR #443, opened 2024-08-13
Docs: remove mention of 12v for printer power supply

tadeubas:

- [ ] Bug fix
- [ ] New Feature
- [x] Documentation update
- [ ] Other


odudex:
12V printers are compatible with Krux, and users have the option to choose them. This is not a plug-and-play feature and requires basic electronics knowledge, the understanding of specs, allowed voltage ranges, etc. It's important not to remove this information to avoid misleading users into thinking that 12V models cannot be used.

PR #445, opened 2024-08-16
Simplified translations

tadeubas:
### Description

Removing unnecessary translations, merging others to simplify the translation process and also to reduce a little the firmware size

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


src/krux/pages/tiny_seed.py
```diff
@@ -179,7 +179,7 @@ def print_tiny_seed(self):
         pad_y = 8  # 2mm*8px
         self.ctx.display.clear()
         self.ctx.display.draw_hcentered_text(
-            t("Printing") + " ...", self.ctx.display.height() // 2
+            t("Printing") + "..", self.ctx.display.height() // 2
```
jdlcdl, line 182: I like this pattern (if it works in all languages "..")

jdlcdl:
This pr has:
* -654 fewer lines
* +303 replacing lines

Each translator file has:
* -27 fewer lines
* +11 replacing lines

green checks

Wow!

src/krux/pages/tiny_seed.py
tadeubas: We have been using this 2 dots pattern since first version of Krux!

https://github.com/selfcustody/krux/blob/v22.03.0/i18n/translations/en-US.json

jdlcdl: I meant the `t("In the act of doing") + ".."` vs `t("In the act of doing..")` pattern.

tadeubas: But Jeff's file (first version of Krux) doesn't distinguish between going to do or doing, see examples below:
```
Checking receive address %d for match..
Loading..
Shutting down..
Updating bootloader..
Upgrading firmware..
Backing up bootloader..
```

tadeubas: Sorry! I saw what you mean now... it is the separation of the ponctuation from the word... I think we can't do that. I will change the above

odudex:
Thank you!

PR #447, opened 2024-08-27
Removing pl-PL translation until we have a maintainer

tadeubas:
### Description

- Removes `pl-PL` translations. If a user had selected that locale, it defaults to `en-US`.
- Created checks for `Category and Number Settings` to ensure that wrong stored configs doesn't broke functionalities.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [x] Other


odudex:
Thank you!
Great work on settings and tests!

PR #452, opened 2024-09-07
draw_ram_indicator() easier numbers to read, includes RAM just collected

jdlcdl:
### Description

The idea here was twofold:
* easier free RAM numbers to read (between 1 and 1023) usually 3 digits of precision with K or M to see the big picture,
* remember amount of free RAM before gc.collect() frees it so that we can see how much was recently collected... but this will not always be true if there is a habit of using gc.collect() within routines that are memory-costly.

Since there will be loops that absolutely require gc.collect(), perhaps we can make a habit of doing it like:
```python
pre_collect = gc.mem_free()
gc.collect()
print("%s has collected %d bytes" % (__name__, gc.mem_free() - pre_collect))
```
...so that we can at least see how much memory had been used and that gc.collect() freed it for x times for this routine?  Otherwise, a bad habit of collecting memory even when it's not necessary may become the norm?

Alternately, I wonder if we might never gc.collect() except where we are absolutely sure it's necessary... and then more generally use try/except "MemoryError", then do gc.collect() within the except block, and continuing on?

Debugging RAM usage.

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other

### DIFF
--- a/src/krux/pages/__init__.py
+++ b/src/krux/pages/__init__.py
@@ -639,9 +639,36 @@ class Menu:
     #     self.draw_ram_indicator()
 
     # def draw_ram_indicator(self):
-    #     """Draws the amount of free RAM in the status bar"""
+    #     """Draws the amount of free RAM in the status bar +recently-collected"""
+    #     def strnum(_in):
+    #         large_units = ("","K","M")
+
+    #         value = _in
+    #         for i in range(len(large_units)):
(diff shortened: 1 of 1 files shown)


odudex:
Thank you Jean! Your indicator improvements will help answer these questions we all have.

PR #459, opened 2024-09-26
fixed verifying command in docs steps on windows

qlrd:
### What is this PR for?

A user on KruxBR found an issue on integrity verification step on Windows, more specifically, in powershell's `Get-FileHash` and `Get-Content` commands. 

Since the `Get-FileHash` output a hash in upper case,when comparing with the output of `Get-Content` in lower case, the verification will result in a `False` value. 

But this is a Negative False. 

To fix this, we expanded the power shell command as well adding steps for a scrutiny procedure.

- [ ] Bug fix
- [ ] New feature
- [x] Docs update
- [ ] Other

### DIFF
--- a/docs/getting-started/installing/from-gui/windows.en.md
+++ b/docs/getting-started/installing/from-gui/windows.en.md
@@ -11,7 +11,34 @@ verify-the-integrity-explain.en.txt
 ----8<----
 
 ```pwsh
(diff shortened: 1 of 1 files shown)


odudex (approved):

odudex:
Thank you!

PR #464, opened 2024-10-04
From English to Japanese - Krux translation

BitCoisas:
### Description

This PR introduces the translation of Krux from English to Japanese. It guarantees that both the user interface and instructional content are fully localized for Japanese speakers, improving overall usability and accessibility.

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


i18n/translations/ja-JP.json
```diff
@@ -0,0 +1,295 @@
+{
+    "% of the amount.": "% の金額.",
+    "%d of %d multisig": "%d の%d マルチシグネチャ",
+    "%d to %d": "%d へ %d",
+    "%s removed.": "削除されました.",
```
jdlcdl, line 5: Is this translation missing a "%s" for a replacement string?

jdlcdl:
I am not capable to verify any of these Japanese translations, however I have spot-checked a few using translate.google.com and they appear to make sense from Japanese back to English.

Besides a single case of a missing "%s" replacement (mentioned on line 5) all other replacement flags appear to be included in the Japanese translations.

Font work will be required before krux devices can use this translation, but thank you for submitting this work.

jdlcdl:
This note is more for myself than for BitCoisas, but Im storing it here so that I can find it in the future, and while I work-out the process of adding support for another language,

TODO list for supporting Japanese (maybe other languages with font glyphs not already used in krux):

[X] Double-check for (and fill) any missing "%" replacements in translated strings

[ ] Check to see which source font files might already contain all of the glyphs needed.  I did it like this:
```python
import os
import json

l10nfile = './i18n/translations/ja-JP.json'
l10ndict = json.loads(open(l10nfile).read())

chars = []
for translation in l10ndict.values():
    for char in [x for x in translation if x >= ' ']:
        if char not in chars:
            chars.append(char)
chars.sort()
print('\nTranslation file %s has %d unique symbols in its translations' % (l10nfile, len(chars)))
print('lowest unicode point: %s (or %d), highest unicode point: %s (or %d)' % (
    hex(ord(min(chars))), ord(min(chars)),
    hex(ord(max(chars))), ord(max(chars))
))


fontdir = './firmware/font'
for basename in [x for x in os.listdir(fontdir) if x.endswith('.bdf')]:
    bdf = '/'.join([fontdir, basename])
    print('\nChecking font file %s for existing/missing glyphs...' % bdf)
    bdf_contents = open(bdf, 'r').read()
    existing, missing = [], []
    for char in chars:
        if '\nENCODING %d\n' % ord(char) in bdf_contents:
            existing.append(char)
        else:
            missing.append(char)
    print('  existing: %s\n  missing: %s' % (''.join(existing), ''.join(missing)))
    if len(missing):
        print('  missing %d unicode points between: %s (or %d) and %s (or %d)' % (
            len(missing),
            hex(ord(min(missing))), ord(min(missing)),
            hex(ord(max(missing))), ord(max(missing))
        ))
```
...which shows that we'll need to find a 14pt high font file with Japanese characters, our 16pt and 24pt high fonts seem sufficient at a glance.

[ ] Glyphs that are not 2x high as they are wide, like roman and are 1:1 need to be part of the wide-glyphs set.  In `firmware/fonts/bdftokff.py` add the new locale to the `wide_glyphs` list variable.

[ ] `src/krux/display.py` needs to know if the current locale uses wide-fonts (to short-circuit around an unnecessary call to 
`lcd.string_has_wide_glyph()` so add the wide-glyph locale (ie: "ja-JP") to the condition protecting that call.

[ ] submodule `firmware/MaixPy` defines a range of min/max unicode-codepoints for wide-glyphs.  Currently it's the first Chinese character to the last Korean character -- but with Japanese, I have a feeling it will be the first Japanese punctuation character around 0x3000 and the last full-width-roman-and-half-width-katakana character at 0xffef.  Constants for this range are in `firmware/MaixPy/components/micropython/port/src/omv/img/include/font.h` and each project uses these in its `projects/BOARDNAME/compile/overrides/components/micropython/port/src/omv/img/font.c` file

... (more to do I'm sure, because the above isn't working for me yet ;)

odudex:
Really nice script! We should add it to the i18n validation workflow or to the firmware/fonts, to be run before important commits and PRs

i18n/translations/ja-JP.json
BitCoisas: You're right, i don't know if i can change it here, but i think @odudex  can edit it before the final version! Thx so much for review it.

BitCoisas:
@jdlcdl About using translate.google.com, as I'm not a fluent Japanese speaker, I've verified some phrases using AI or Google Translate to check if what I wrote makes sense. However, I have double-checked the entire translation with two native Japanese speakers, and they said it's okay. Maybe you are referring to the sentences in katakana, but katakana is mainly used for foreign words, names, and onomatopoeia in Japanese. Its usage is specific and distinct from hiragana or kanji, and while it may appear different, it's essential for expressing certain terms in the language. If there are particular concerns about the use of katakana, feel free to let me know, and I'd be glad to clarify.

i18n/translations/ja-JP.json
odudex: @BitCoisas If you update your fork with the correction it will automatically update this pull request. I would be thankful if you could do that, just to ensure we won't forget that later.

BitCoisas: woahhhh, i didnt know that, thanks guys!

BitCoisas: Did it!

jdlcdl:
[BitCoisas](https://github.com/BitCoisas), thank you for elaborating on the translation, also, thank you for the new commit adding the only missing replacement.

When I'd mentioned that I cannot verify the translations, it was just to leave record that I couldn't do that while trying to verify what I'm capable of towards a review.

As it stands now, I think we still need to find 14pt high fonts.
For at least the 24pt fonts, which I use in the Amigo device, there exists a problem that needs solving in the bdf to kff conversions (a dev task) because some of the wide fonts are expressed in the bdf as 16 bits wide and for some reason they're being ignored during conversion... other glyphs that are expressed as 24 bits wide appear to be working.  So, I have work to do on that front.  More soon.

jdlcdl:
Correction, it wasn't so much a broken conversion as it was me forgetting that another converter step also needs to know about another range of unicode code points (the hextokff step) so it appears to be displaying characters on my Amigo, and we're back to searching for 14pt fonts in japanese (hopefully that serve for chinese and korean too), and I'll othewise just edit my "how to add support for another asian language" above.

jdlcdl:
I've finally made adjustments for 14pt Japanese fonts (borrowed from FusionPixelFont's Galmuri11 set).  pr #465 which relies on this one is expected to work for all devices (however I can only test on the large Amigo).  Will work with other devs to get this ready whenever possible (likely after the pin_and_anti-tamper branch has settled).

odudex:
Thank you @BitCoisas !
@jdlcdl has created the necessary font glyphs support and Krux is ready to receive your Japanese translation.

PR #469, opened 2024-10-14
Avoid exiting when user misclick and cancel the load of a file

tadeubas:
### What is this PR for?
- Fixed bug when signing message with empty file
- Avoid exiting when user misclick and cancel the load of a file

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


odudex:
Thank you!
It would be nice to add a test to cover new code and meet the patch requirements.

odudex:
Thank you!

PR #471, opened 2024-10-15
Tamper Detection 

odudex:
### What is this PR for?
Add a tamper detection mechanism called "Tamper Code Flash Hash" or "TC Flash Hash".
The mechanism combines cryptographic hashes, a Tamper Check Code, and camera-generated entropy to create a tamper indicator that is unique to each device, represented by a memorable image and four words.
A side feature is the "Flash Map", which is a visual tool to check the areas of the flash memory that contain data.
This PR also covers the addition of BIP85 base64 passwords.

- [ ] Bug fix
- [X] New feature
- [ ] Docs update
- [ ] Other


jdlcdl (commented):
as of 108th commit 4c91316
* have followed this branch since the beginning,
* up to date on visual code review
* still need to test beta10 on Amigo

PR #480, opened 2024-11-28
Fix incorrect settings in SD changing user defined values in flash

tadeubas:
### What is this PR for?
When user had an old settings file in the SD (like when updating from an old version), change in settings would not show up when the device restarts (appeared as not persisted)

Below attached a settings file to help testing with your SD card:
[settings.json](https://github.com/user-attachments/files/17942197/settings.json)

- [X] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/settings.py
+++ b/src/krux/settings.py
@@ -132,39 +132,45 @@ class Store:
 
     def __init__(self):
         self.settings = {}
-        self.file_location = "/" + FLASH_PATH + "/"
+        self.file_location = Store.get_vfs_location(SD_PATH)
         self.dirty = False
 
         # Check for the correct settings persist location
-        try:
-            with open(self.file_location + SETTINGS_FILENAME, "r") as f:
-                self.settings = json.loads(f.read())
-        except:
-            pass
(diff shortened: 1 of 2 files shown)


src/krux/settings.py
```diff
@@ -210,17 +216,15 @@ def delete(self, namespace, setting_name):
...
-            try:
-                if os.stat(self.file_location + SETTINGS_FILENAME):
-                    os.remove(self.file_location + SETTINGS_FILENAME)
-            except:
-                pass
-            self.file_location = "/" + location + "/"
+        """Delete any unused persistent settings when set location"""
+        self.file_location = Store.get_vfs_location(location)
+        try:
+            if FLASH_PATH in self.file_location:
```
odudex, line 222: Why the data transformations then search withing a string when you can compare directly?
`if location == FLASH_PATH:`

odudex: Nevermind! Noted you're also updating the "file_location" variable

odudex:
Thank you!

PR #487, opened 2024-12-06
Hide mnemonic when wallet loading if the Hide mnemonic setting is True

btcinsaneguy:
### What is this PR for?
When `Hide mnemonic` setting is True, it will hide the mnemonic while login

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/pages/login.py
+++ b/src/krux/pages/login.py
@@ -254,8 +254,10 @@ class Login(Page):
 
         from .mnemonic_editor import MnemonicEditor
 
-        mnemonic_editor = MnemonicEditor(self.ctx, mnemonic, new)
-        mnemonic = mnemonic_editor.edit()
+        # If the mnemonic is not hidden, show the mnemonic editor
+        if not Settings().security.hide_mnemonic:
+            mnemonic_editor = MnemonicEditor(self.ctx, mnemonic, new)
+            mnemonic = mnemonic_editor.edit()
         if mnemonic is None:
             return MENU_CONTINUE
         self.ctx.display.clear()


tadeubas:
And how to confirm the mnemonic that was loaded? Just by fingerprint?

jdlcdl:
I agree with this, even to the degree that bip85 child mnemonics should also not be shown.

My original thoughts on this were:
* I create my mnemonic once, and while I may want to back it up again, or confirm it, or learn to remember my mnemonic with the help of krux (in the case of an encrypted mnemonic), most of the time I use krux I would NOT want this to be available.
* I might login to krux but then I put my mnemonci backup away immediately, then my focus shifts to my coordinator wallet. I could get distracted and I don't want those living with me, or untrusted friends who were invited into the home to have a chance to find a logged-in signer.
* In the event that SeedXOR ever becomes available, while a bad idea to do the following, trusted associates like husband and wife might share a mnemonic by keeping a mnemonic each to themselves and funding the seedXOR result;  If the resulting mnemonic were visible then the first party who entered their mnemonic would lose it to the one who now knows the final result and can simply remove theirs to learn the other... and "hide mnemonic" doesn't help unless it's truly hidden.
* While I love that krux helps teach you to remember your mnemonic every time you load it, it could be against what the user truly wants... the ability to eat/burn their mnemonic backup in a pinch (knowing they can jack-hammer the concrete patio to recover a steel backup).  It should be plausible to attackers that we don't actually remember our backup, and if hide-mnemonic is enabled in settings and we rarely see mnemonic, then this is plausible.

I do however believe strongly that the naked (passphrase-less) bip32 fingerprint should replace the mnemonic when hidden, and even that naked fingerprint should be visible with the mnemonic during edit.

btcinsaneguy:
Thank you for all the feedback and for maintaining this great project. 

I created this PR because I've been personally using a custom build with this feature, as there are many unexpected CCTVs in Asian countries where I live.

jdlcdl (approved):
I  <3 this pr!  Thank you btcinsaneguy!

Because this was pr'd back in November, I had to fetch this pr, then rebase it to current develop.  All tested well, I was a big fan of hide-mnemonic but now I'm a bigger fan of it.

odudex:
Thank you guys!

btcinsaneguy:
Thank you for the kind words, but I should be the one thanking you for creating such an amazing open source Bitcoin wallet project! I've been using it regularly and really appreciate it. I'll definitely contribute again in the future if I find more ways to help improve it!

PR #488, opened 2024-12-06
Simplifying messages / translations + small changes

tadeubas:
### What is this PR for?

- Simplified the below translations to just "Failed to load" :

```
"Failed to load PSBT",
"Failed to load address",
"Failed to load key",
"Failed to load message",
"Failed to load mnemonic",
"Failed to load output descriptor",
"Failed to load passphrase",
```

- Removed translations for extreme rare error messages (some in `firmware.py`):
```
"Maximum length exceeded (%s)",
"invalid public key",
"Invalid bootloader" ,
"Firmware exceeds max size",
```

- Reuse of function `status_text` on `firmware.py`
- Added `try / except` block around function that could raise an exception without informing the user in `firmware.py`
- Fixed very long key / passphrase not showing properly at the proceed screen on m5stickv
- Fixed key / passphrase not checking for too long input using keypad
- Usage of `ValueError` instead of `flash_error` when loading too long input for key / passphrase
- Usage of `flash_error` instead of `flash_message` for error messages on bkp encrypted mnemonic
- New poetry poe task alias for wonder-mv: `simulator-mv`

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other


jdlcdl (approved):
I've read all changes here and have done random testing of these in simulator and on amigo.  All seems ok to me.

---

Just a thought for a future commit (or future pr as it might get out of hand).

* multiple display.flash_text() calls in firmware.py are for errors rather than informative text, so perhaps display might appreciate a method like "display.flash_error()" as exists in pages.__init__ so that it's obvious that firmware upgrade failed (before message disappears).

tadeubas:
I agree, we should use the error color when an error appear in `firmware.py`. And sure, this can be done in another PR

odudex (approved):
Great clean up and refactor! Congrats and thank you!

PR #491, opened 2024-12-09
Fix firmware.py opening files and not closing

tadeubas:
### What is this PR for?
Files were being opened without being close, this could lead to errors in another open operation on the same file that was left open

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other


src/krux/firmware.py
```diff
@@ -258,17 +259,19 @@ def status_text(text):
...
         ),
         new_address,
-        open(firmware_path, "rb", buffering=0),
+        firmware_file,
         new_size,
         65536,
         True,
         firmware_with_header_hash,
     )
+    firmware_file.close()
```
jdlcdl, line 274: Might this need to be within the "finally" branch of try/except, in case of a write failure? or within the context of "with open()" so that it's automagically closed after the write_data() call?

tadeubas: Will change that!

Issue #486, opened 2024-12-04
[Bug] Multisig Bluewallet Wallet Descriptor not recognised


TimYorke:
Krux 24.11.1 running on yahboom HW and Bluewallet v7.0.6 (also reproduced on v6.5.5)

I see the error "Invalid wallet: ValueError('invalid wallet format',)" when I try to import the wallet descriptor from BlueWallet (Android). I see the progress bar increase as it recognises the QR codes, but then just displays a black screen with this error message. 

I can reproduce it using the QR code animation displayed in "Export Coordination Setup" in BlueWallet, and via "Tools->Descriptor Addresses->Load from camera" in Krux

It's a 2/2 native segwit "Multisig Vault" and derivation path "m/48'/0'/0'/2'"

I'm happy to try and debug locally if someone can give me some pointers.

jdlcdl:
Would you be able to confirm if you see the cosigner keys as "Zpub" instead of "xpub" in your BlueWallet backup file?

odudex:
Your report helped us identify incompatibilities with BlueWallet multisig descriptors and PSBT signing in Krux. We’ve released a fix in [v25.01.beta3](https://github.com/odudex/krux_binaries), which should allow you to successfully load your descriptor and sign PSBTs.

A stable release containing this fix is planned for next month. After not been testing with Blue recently, we will now do more tests and continue to improve compatibility. In the meantime, if you prefer to stick with the current stable release, you can try recreating your wallet in BlueWallet by exporting only the xpubs. This approach should help you avoid the issue until the stable fix is available.

TimYorke:
Thank you. The xpub-only workaround did the trick

odudex:
Latest release should have it fixed, can you confirm?

PR #492, opened 2024-12-10
fix: assuming issue 486 is Zpub instead of xpub

jdlcdl:
### What is this PR for?

Assuming that issue #486 is about cosigners expressed as "Zpub" instead of "xpub", this may be a fix.

Note: Need to verify that this will not cause problems when signing a BW PSBT w/ wallet descriptor loaded.

- [X] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/wallet.py
+++ b/src/krux/wallet.py
@@ -268,10 +268,14 @@ def parse_wallet(wallet_data, allow_assumption=None):
 
             keys = []
             for i in range(len(key_vals)):
-                xpub = key_vals[i]
-                if xpub.lower().startswith("xpub") or xpub.lower().startswith("tpub"):
+                kv = key_vals[i]
+                kv_prefix = kv[:4].lower()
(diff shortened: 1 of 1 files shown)


odudex:
Tested this on an on-chain, 2 of 3 wallet, with 2 Zpus and a one xpub as cosigners.
Successfully loaded the wallet descriptor and signed, with both Zpub and xpub respective keys.

src/krux/wallet.py
```diff
@@ -294,6 +298,7 @@ def parse_wallet(wallet_data, allow_assumption=None):
             )
             return descriptor, label
     except:
+        raise
```
odudex, line 301: On my tests i removed this duplicated raise

odudex:
To fix pylint issue, we could put code below
```
# Try to parse as a key-value file
try:
```
inside a dedicated function: `parse_key_value_file`

odudex:
We can close this one, right?

PR #493, opened 2024-12-11
descriptors w/o fingerprints show "@ unknown"

jdlcdl:
### What is this PR for?

This PR will show "@ unknown" when loading a descriptor w/o origin/master fingerprint information, instead of showing the hd-key's own fingerprint.

Background: All bip32 extended keys have their own fingerprint, and most have a parent fingerprint.  These fingerprints are contained within or can be derived from the base58 xprv/xpub.  But both of these are NOT what users think of "fingerprint" except for the root level bip32 master key... which is not available via xprv/xpub except for the master's depth=1 children.  For this reason, the "origin info", containing "[fingerprint/derivation/path]" is included in many descriptors.

But since not all descriptors include the optional origin information, namely BlueWallet single-sig wallet backups, it is incorrect (or at least confusing to the user) to show a key's own fingerprint when they're actually expecting to see the origin/master fingerprint instead.  If there is no origin information, it is better for the user to see "@ unknown" instead of a fingerprint they don't recognize.  

Thank you Tadeubas for pounding this concept into my head, and for being tolerant enough till I understood.  Because of this, I'm going to do better by using the terms "bip32 master fingerprint" or "origin fingerprint" instead of the more vague term "fingerprint" in future discussions/explanations.

- [*] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other


odudex:
Thank you @jdlcdl and @tadeubas !

PR #498, opened 2024-12-19
Taproot Miniscript

odudex:
### What is this PR for?
After having P2WSH Miniscript compatibility added to develop branch this PR adds compatibility for Taproot Miniscripts, first tests done targeting Liana as coordinator.

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other


odudex:
At this point I'm prioritizing SD card signing for miniscript, as Liana does not support QR codes.
I'm rethinking SD card to flow, also having multisig in mind. On current commit I won't trim any data of PSBT, this way, in a multi-device setup you can:
Create PSBT on PC -> Sign in device 1 -> Sign in device 2 -> Broadcast on PC.
(Without needing to load device 1 signatures on PC prior to sign with device 2)

jdlcdl:
Recording thoughts after testing this at commit d467330:

Without loading a mnemonic, using tools / descriptor addresses, to load a wallet descriptor from liana
* WORKS: to load and find correct addresses for miniscript-wsh as well as miniscript-tr
* TODO: however it shows "TR Miniscript" in both cases, instead of showing WSH when that's the case.
... if doing the same when a mnemonic is loaded:
* still works to find correct addresses just like above, still needs a correction on the title to reflect WSH when applicable,
* TODO: perhaps a mismatch check needs to be added, so that when krux is expecting miniscript WSH and instead loads a TR descriptor, a warning is displayed.

When signing a PSBT from liana w/ wallet descriptor:
* WORKS: to raise a policy-mismatch if wrong wallet descriptor is loaded,
and without a wallet descriptor loaded:
* WORKS: to show policy "p2tr" and list of key-origin fingerprints, for a psbt that is miniscript-p2tr,
* WORKS: to show policy "p2wsh" w/o list of keys but could be improved to include key-origin fingerprints -- for a psbt that is miniscript-wsh,
and overall:
* WORKS: to display what is being spent, whether it's change/self-transfer and what the fees are
* appears to create valid signatures, however I had to `finalizepsbt` and `sendrawtransaction` on a signet node because liana didn't show me the "broadcast" button.... but odudex has explained how to get around this for next time (save first, then sign, then update, then can broadcast)

Learned from odudex that we cannot count on specific derivation paths, different wallets (nunchuk's m/87'/0'/0' tap multisig) may use their own preferred paths...and krux may need to support `custom derivation path` to create/recover for different wallet softwares. 

Also noted that `liana` conveniently uses the bech32 checksum, which is the suffix on the wallet descriptor, as a wallet identifier, and this is very nice... perhaps something for krux to adopt, to display when loading a descriptor.

Also todo: review/tests for embit changes, tests for krux too.

src/krux/pages/home_pages/wallet_descriptor.py
```diff
@@ -177,7 +178,16 @@ def display_loading_wallet(self, wallet):
                 key_origin_str = str(key.origin)
                 key_fingerprint += key_origin_str[:8]
             else:
-                key_fingerprint += t("unknown")
+                if (
```
jdlcdl, line 181: Does this need a wallet-sans-key condition for reviewing descriptor w/o loading wallet first?  Else wallet.key is None and won't have policy_type or script_type attributes?

Asking because I tried loading Gui's TR descriptor via tools/wallet-descriptor... and saw "NoneType has no policy_type", but not sure if it's from this line or the elif at line 216

odudex: Thank you!
Fixed by getting this information from descriptor instead of loaded key (the right way)

src/krux/psbt.py
```diff
@@ -504,35 +519,43 @@ def _fill_zero_fingerprint_scope(self, scope):
...
-            # Check for P2WSH (SegWit multisig) and copy witness_script if present
-            if hasattr(inp, "witness_script"):
+            # Preserve witness_script for P2WSH multisig
+            if inp.witness_script:
                 trimmed_psbt.inputs[i].witness_script = inp.witness_script
 
+            # Preserve taproot_key_sig for Taproot inputs
+            if inp.taproot_key_sig:
+                trimmed_psbt.inputs[i].taproot_key_sig = inp.taproot_key_sig
+
```
jdlcdl, line 558: `inp.taproot_script_path_sigs` too?

I've seen this field (and others) in the "sign to sdcard" version but it's not showing up in the "sign to qr" version.

odudex: Indeed, `inp.taproot_sigs` were missing for QR code exported signed PSBT. Fixed!
Now we probably still need to add some more fields so the signed PSBT can be broadcasted as a standalone verifiable and broadcastable PSBT (not only for  updating a full PSBT on the coordinator)

src/krux/wallet.py
```diff
@@ -382,17 +444,29 @@ def parse_address(address_data):
...
 
-    try:
-        sc = address_to_scriptpubkey(addr)
-        if not isinstance(sc, Script):
+    if addr == addr.upper():
+        # bip173 suggests bech32 in uppercase for compact QR-Code
+        try:
+            sc = address_to_scriptpubkey(addr.lower())
+            if isinstance(sc, Script):
+                addr = addr.lower()
```
jdlcdl, line 459: refactor?  If we get here, we might as well just `return addr.lower()` and avoid re-setting this var, and more isinstance() checks?

src/krux/wallet.py
```diff
@@ -168,14 +190,51 @@ def load(self, wallet_data, qr_format, allow_assumption=None):
                 "n": n,
                 "cosigners": cosigners,
             }
-        elif self.descriptor.miniscript is not None:
+        elif self.descriptor.miniscript is not None or self.descriptor.taptree:
+            if self.descriptor.taptree:
+                if not descriptor.keys[0].origin:
+                    import hashlib
+                    from embit.ec import PublicKey
```
jdlcdl, line 197: could also import NUMS_PUBKEY directly from embit.ec here, and use it below instead of redefining BIP_341_NUMS_EXAMPLE at line 45 above.

Note: embit.ec.NUMS_PUBKEY has comment pointing to bip341... but your added comments above might be kept somewhere here.

odudex: Great finding! Let's use it!

PR #506, opened 2025-02-21
Keeps the ABC key in the same position for all keypad types

rough-street:
This is a fix for one of my main frustrations when entering my passphrase - the moving "ABC" key!
As I looped through the keypads, the ABC key changed position on the numeric keypad and made me hit the delete key, meaning I had to loop through even more!

I have fixed this issue by changing the order of the keys so that the ABC key is in the bottom right of the screen for all keypad types in screens that use it.

### DIFF
--- a/src/krux/pages/keypads.py
+++ b/src/krux/pages/keypads.py
@@ -69,13 +69,13 @@ class Keypad:
     def more_index(self):
         """Returns the index of the "More" key"""
         if len(self.keysets) > 1:
-            return self.del_index - 1
+            return self.go_index + 1
         return None
 
     @property
     def del_index(self):
         """Returns the index of the "Del" key"""
-        return len(self.keys) + self.empty_keys + (1 if len(self.keysets) > 1 else 0)
+        return len(self.keys) + self.empty_keys
 
     @property
     def esc_index(self):
--- a/src/krux/pages/keypads.py
+++ b/src/krux/pages/keypads.py
@@ -69,13 +69,13 @@ class Keypad:
     def more_index(self):
         """Returns the index of the "More" key"""
         if len(self.keysets) > 1:
-            return self.go_index + 1
+            return self.del_index - 1
         return None
 
     @property
     def del_index(self):
         """Returns the index of the "Del" key"""
-        return len(self.keys) + self.empty_keys
+        return len(self.keys) + self.empty_keys + (1 if len(self.keysets) > 1 else 0)
 
     @property
     def esc_index(self):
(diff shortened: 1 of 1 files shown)


odudex:
I liked the idea. All the possible ideas I've previously thought to solve this issue had terrible trade-offs, like reducing buttons surface area and the optimal use of the display.
I will create a pool on our [Telegram group](https://t.me/SC_Krux) proposing the change to get some users' feedback.

odudex:
Just another note that makes this PR quite labor intensive:
It's important to check the impact of the changes on:
- Other parts of the UI
- On tests that involve navigation `poetry run poe tests`
- Screen shot generation routines for docs that also involve navigating through keypads 
`cd simulator && ./generate-all-screenshots.sh`

rough-street:
Thanks. Yes, you're right about the tests - I can see some are now broken. I'll get them updated.

odudex:
Wait a bit, let me check another idea: Having a fixed column count may not be too bad.

tadeubas:
Hey @odudex , do you know why the workflow in this PR is not showing the tasks like the test that is broken?

Yesterday I've opened a PR that executed all github actions fine

odudex:
> Hey @odudex , do you know why the workflow in this PR is not showing the tasks like the test that is broken?
> 
> Yesterday I've opened a PR that executed all github actions fine

I have to manually approve the run

odudex:
How about fixed column count? Users wouldn't need to get used to a new position, and tests don't need to change.
image

rough-street:
Not having to change the tests is appealing to me :-) This could be less work after all - I'll go ahead with trying to implement this instead with some new commits.

odudex:
> Not having to change the tests is appealing to me :-) This could be less work after all - I'll go ahead with trying to implement this instead with some new commits.

Right, for tests I've created a variable at `__init__`
```
        self.max_keys_count = max([len(keyset) for keyset in keysets])
        self.max_keys_count += FIXED_KEYS + (1 if len(keysets) > 1 else 0)
```
then used it instead of `self.total_keys` in `width` and `height` keypoard properties

rough-street:
Thanks @odudex. That seems to work great.

odudex:
Thank you for contribution, hopefully first of many!

tadeubas:
Hey @odudex I think you forgot to run the github action tasks... lint is broken:
```
Poe => pylint src
************* Module krux.pages.keypads
src/krux/pages/keypads.py:43:0: R0902: Too many instance attributes (12/11) (too-many-instance-attributes)
src/krux/pages/keypads.py:50:30: R1728: Consider using a generator instead 'max(len(keyset) for keyset in keysets)' (consider-using-generator)
```

odudex:
I didn't. I'll do a bigger refactor that will cover this

PR #508, opened 2025-02-21
Removed the unnecessary NUM_SPECIAL_3 

tadeubas:
As multiple keyboards are now fixed in size it doesn't make sense to separate numbers from special chars for TC code anymore

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other


odudex:
Great, thank you!

Issue #507, opened 2025-02-21
[Enhancement] Indicate to user that they can swipe on keypads


rough-street:
Currently, it's not at all obvious that when entering general text (e.g. passphrase) that you can swipe the touchscreen to move between keypads, which is much faster than looping through the keypads with the ABC button.

I can see two screens where general text keypads are used:
- enter passphrase
- enter text for QR code

At the top of each screen, it simply says "Passphrase" or "Text". This seems a bit redundant and the space could be used to display a hint like "<< swipe >>".

I'm happy to make a PR with this change, or maybe you have some other ideas, considerations?

rough-street:
@tadeubas 

They're good points. 

1. sure - I'll use @odudex's predicate above which should take care of that
2. [Transcribing QR codes](https://selfcustody.github.io/krux/getting-started/features/QR-transcript-tools/) already has a hint (could be improved separately to follow the format settled on for the keyboard). The [settings page](https://selfcustody.github.io/krux/getting-started/settings/) and [files navigation](https://selfcustody.github.io/krux/getting-started/features/tools/#check-sd-card) use vertical scrolling of a list which is already fairly intuitive based on mobile phone behaviour. But if it's not obvious there's more to see, then the best improvement for these would be to add a scrollbar to indicate, or make the last item show "more..", or something along those lines if they don't already. Best done as separate piece of work I think.
3. yes I think the icons look smart and separate the UI elements nicely - I like the idea of using that format for the hint.

rough-street:
Aha -  I see what you mean by custom icon now - a custom font character. Yes, it looks like there's plenty of suitable existing ones to choose from. I'd say we're spoilt for choice.

rough-street:
I'm going to implement this with some extra filter logic so as to not show the swipe hint until the user of a touch screen device hits the ABC key. Then I'll display the hint in place of the title for the entry screen ("Passphrase", "Text" etc.) until they swipe, at which point the hint won't be shown again for the remainder of the screen.

That way they'll be be given the hint until they actually try swiping, and then as long as they swipe from then on, they'll never see the hint again.

PR #512, opened 2025-02-22
Swipe hint for touchscreen keypads

rough-street:
Solution to issue https://github.com/selfcustody/krux/issues/507

For HW with touchscreens, a swipe hint will be displayed in place of the title text each time the ABC key is tapped on a text-entry screen, until the user tries swiping. 

- [ ] Bug fix
- [*] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/pages/__init__.py
+++ b/src/krux/pages/__init__.py
@@ -33,6 +33,8 @@ from ..input import (
     BUTTON_TOUCH,
     SWIPE_DOWN,
     SWIPE_UP,
+    SWIPE_LEFT,
+    SWIPE_RIGHT,
     ONE_MINUTE,
 )
 from ..display import (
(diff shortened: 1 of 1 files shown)


rough-street:
It seems there aren't any existing tests for the title of the keypad screens. Some advice on how I should to pass the code cov threshold would be appreciated :-)

tadeubas:
Codecov is just saying that new lines of code were added on `src/krux/pages/__init__.py` and those lines were not executed by any test case:

```
src/krux/pages/__init__.py#L184-L186

Added lines #L184 - L186 were not covered by tests
```

You have to see on our tests folder where we test the `__init__` page (that is used too by other pages that are subclasses) and add a test case that would cover those lines (execute such lines).

Another thing is that we use i18n so every text needs to be translated, in this case "swipe" needs to be enclosed by the `t()` function (only the text, not the icons). This will also require you to fill all the necessary translations (we have some poetry tasks to help with that, one called fill will automatically use google translate to give you the translated texts to copy and paste on the corresponding locale file).

I've tested on our simulator and it is appear to be working as expected, just missing tests and translations.

odudex:
Complementing what @tadeubas said:

**Tests**
A test the runs the added code and triggers the hint should be added to `/tests/pages/test_page.py` file. Try to get inspiration from a similar keypad capture test, that will press the keypad toggle button and activate the hint.
Also use the swipe term inside translation function like `swipe_hint = "« {} »".format(t("swipe"))`

**Translations**
Finally, cd into i18n and perform the translations checks and corrections, like:
```
poetry run python i18n.py validate
poetry run python i18n.py clean
poetry run python i18n.py fill  # Then manually review and copy translations to respective files
poetry run python i18n.py bake
```
then cd .. and run `poetry run poe format` format final translations files

rough-street:
Thanks for the pointers - very clear. 

This is the first python project I've worked on. I'm a compiled language guy (C++, C#, Rust) so I'm used to relying on the compiler to check for language-level errors rather than exercising the code via a test.

rough-street:
Now I can see how the test/mocking framework works, I'm happy to add some behavioural tests for this if you'd like them.

odudex:
Nice work! Although not all of our tests do so perfectly, it would be good to add behavioral tests, like checking if "« swipe »" is written on the display.

rough-street:
If you're happy with this, then I think I'm done. Swiping the keypad is now for everyone! No longer a solely a power of those who read all the docs :-)

odudex:
Great work! Will just update the glyphs on the firmware to have "« »"
Thank you!

PR #513, opened 2025-02-23
Fix rare error when loading mnemonic word numbers

tadeubas:
### What is this PR for?
Fix errors when `Loading Mnemonic > Via Manual Input > Word Numbers` and input is empty and user click on "GO" btn:
- ValueError: invalid literal for int() with base 10: ''
- ValueError: invalid literal for int() with base 16: ''
- ValueError: invalid literal for int() with base 8: ''

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other

### DIFF
--- a/src/krux/pages/login.py
+++ b/src/krux/pages/login.py
@@ -632,9 +632,10 @@ class Login(Page):
             return None
 
         def to_word(user_input):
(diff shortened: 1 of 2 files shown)


odudex:
Nice catch!

PR #515, opened 2025-02-25
Fixed keypad touch index stuck after "ESC" > "No"

tadeubas:
### What is this PR for?
- Using touch only, when on any keypad and hit "ESC" then "No", then the keypad would stuck responding only index 0,0 everytime, regardless of the place touched

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other

### DIFF
--- a/src/krux/pages/__init__.py
+++ b/src/krux/pages/__init__.py
@@ -176,6 +176,10 @@ class Page:
                     if esc_prompt:
                         if self.esc_prompt() == ESC_KEY:
                             return ESC_KEY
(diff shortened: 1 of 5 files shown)


jdlcdl:
I confirm that this behavior exists currently.  Not sure how I haven't stumbled on getting stuck in that loop.

Tested this PR which is working well on Amigo.

odudex:
Good catch! I introduced this bug when optimized the keypad to not remap on keyset switch.

odudex:
Thank you!

PR #518, opened 2025-02-26
add qr brightness hint

odudex:
### What is this PR for?
Inform users animated and QR codes brightness can be toggled with button presses.
Refactor animated QR logic to avoid in-loop unnecessary calculations and screen drawings.

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other


jdlcdl:
Tested this on Amigo.

Now that I see where this functionality is available, taking notes of how and where we can change qr brightness:

* Via on screen hint: "page to toggle brightness":
  * login: tools / print test qr (would this alter the brightness of the print?)
  * home: backup mnemonic / plaintext qr
  * home: wallet / wallet descriptor

* Via a "toggle brightness" option in the next menu:
  * login: tools / create qr code
  * home: backup mnemonic / compact seedqr
  * home: backup mnemonic / seedqr
  * home: backup mnemonic / encrypted qr
  * home: extended public key
  * home: address / recv/chg address
  
When convenient and the qrcode is meant to be read by another camera, I prefer the on-screen hint.

I'm not sure how necessary toggling brightness is for manual transcription but I'll note that in these cases, "Page next" is already used for transcription modes.

For extended public key and address-list qrs, it might be convenient to alter brightness while viewing the qr-code (since "page next" button is available... and it doesn't hurt to also have it on the next menu (but it might not be necessary if it can be done while viewing).

odudex:
PAGE now changes brightness when transcript disabled on QR Viewer.
But @jdlcdl notes reminded me maybe qr_viewer.py and `display_qr_codes` from pages.__init__.py could be merged to reuse code and better standardize behavior. A mission to start after next release.

PR #521, opened 2025-03-04
Text with highlight colours

tadeubas:
### What is this PR for?
Adjusting screen texts to use highlight colors:
- The fingerprint on various screens
- PSBT preview (matching highlight color for prefixes, like message sign to addresss)
- File manager prefixes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -535,7 +535,7 @@ max-bool-expr=5
 max-branches=22
 
 # Maximum number of locals for function / method body.
-max-locals=26
+max-locals=27
 
 # Maximum number of parents for a class (see R0901).
 max-parents=7
--- a/src/krux/display.py
+++ b/src/krux/display.py
@@ -375,6 +375,7 @@ class Display:
         bg_color=theme.bg_color,
         info_box=False,
         max_lines=None,
+        highlight_prefix="",
     ) -> int:
         """Draws text horizontally-centered on the display, at the given offset_y"""
         lines = (
@@ -406,14 +407,28 @@ class Display:
                     color,
                     bg_color,
                 )
+                if highlight_prefix:
+                    prefix_index = line.find(highlight_prefix)
+                    if prefix_index > -1:
+                        self.draw_string(
+                            offset_x,
+                            offset_y + (i * (FONT_HEIGHT)),
+                            line[: prefix_index + len(highlight_prefix)],
+                            color=theme.highlight_color,
+                        )
+
         return len(lines)  # return number of lines drawn
 
-    def draw_centered_text(self, text, color=theme.fg_color, bg_color=theme.bg_color):
+    def draw_centered_text(
+        self, text, color=theme.fg_color, bg_color=theme.bg_color, highlight_prefix=""
+    ):
         """Draws text horizontally and vertically centered on the display"""
         lines = text if isinstance(text, list) else self.to_lines(text)
         lines_height = len(lines) * FONT_HEIGHT
         offset_y = max(0, (self.height() - lines_height) // 2)
-        self.draw_hcentered_text(text, offset_y, color, bg_color)
+        self.draw_hcentered_text(
+            text, offset_y, color, bg_color, highlight_prefix=highlight_prefix
+        )
 
     def flash_text(self, text, color=theme.fg_color, duration=FLASH_MSG_TIME):
         """Flashes text centered on the display for duration ms"""
--- a/src/krux/pages/__init__.py
+++ b/src/krux/pages/__init__.py
@@ -292,6 +292,10 @@ class Page:
         header = "BIP39 {}{}".format(suffix, fingerprint)
         self.ctx.display.clear()
         self.ctx.display.draw_hcentered_text(header)
+        if fingerprint:
+            self.ctx.display.draw_hcentered_text(
+                fingerprint, color=theme.highlight_color
+            )
         lines = self.ctx.display.to_lines(header)
         starting_y_offset = DEFAULT_PADDING // 4 + (
             len(lines) * FONT_HEIGHT + FONT_HEIGHT
@@ -305,6 +309,10 @@ class Page:
                 self.ctx.input.wait_for_button()
                 self.ctx.display.clear()
                 self.ctx.display.draw_hcentered_text(header)
+                if fingerprint:
+                    self.ctx.display.draw_hcentered_text(
+                        fingerprint, color=theme.highlight_color
+                    )
                 for i, word in enumerate(word_list[12:]):
                     self.ctx.display.draw_string(
                         DEFAULT_PADDING, starting_y_offset + (i * FONT_HEIGHT), word
--- a/src/krux/pages/file_manager.py
+++ b/src/krux/pages/file_manager.py
@@ -166,31 +166,51 @@ class FileManager(Page):
         """Display the file details on the device's screen"""
         import uos
         import time
+        from ..display import DEFAULT_PADDING, FONT_HEIGHT
+        from ..themes import theme
 
         stats = uos.stat(file)
         size_KB = stats[6] / 1024
         size_KB_fraction = str(int(size_KB * 100))[-2:]
         created = time.localtime(stats[9])
         modified = time.localtime(stats[8])
-        format_datetime = " %s-%02d-%02d %02d:%02d"
+        format_datetime = "%s-%02d-%02d %02d:%02d"
         file = file[4:]  # remove "/sd/" prefix
 
         self.ctx.display.clear()
-        self.ctx.display.draw_hcentered_text(
-            file
-            + "\n\n"
-            + t("Size:")
-            + " "
-            + generate_thousands_separator(int(size_KB))
-            + render_decimal_separator()
-            + size_KB_fraction
-            + " KB"
-            + "\n\n"
-            + t("Created:")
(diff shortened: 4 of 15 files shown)


.pylintrc
```diff
@@ -535,7 +535,7 @@ max-bool-expr=5
 max-branches=22
 
 # Maximum number of locals for function / method body.
-max-locals=26
+max-locals=27
```
odudex, line 538: We should stop doing this, we've gone too far from 15

src/krux/pages/file_manager.py
```diff
@@ -166,31 +166,51 @@ def display_file(self, file):
...
+            self.ctx.display.draw_hcentered_text(
+                file
+                + "\n\n"
+                + t("Size:")
+                + " "
+                + generate_thousands_separator(int(size_KB))
+                + render_decimal_separator()
+                + size_KB_fraction
+                + " KB"
+                + "\n\n",
```
odudex, line 192: instead or printing new lines at the end, you can just increment offset with `2 * FONT_HEIGHT`

.pylintrc
tadeubas: the problem is the import statement, every import inside a function is a "new local variable"

tadeubas:
> I'm noticing that ":" is included in changes to translation strings, but did we not change this pattern to be more like:
> 
> old: `t("label:") + " " + value` had become recent: `t("label") + ": " + value`
> 
> But maybe this is necessary for new use of "highlight_prefix" ???

It is not necessary to have the `:` inside the `t()` function. But if we just use the word with `:` I think it is better for translations to have this information because it add context and can make easy to translate / verify. Some languages might translate different because of this punctuation (I know fr for example that place a THIN_SPACE before `:`, Chinese and other Asian lang was using another char `：` to represent `:` (for better readability / centralization with Asian symbols).

tadeubas:
Converted this to draft until I finish fixing tests

.pylintrc
tadeubas: changed the function and returned the original value on `.pylintrc`

src/krux/display.py
```diff
@@ -365,6 +365,9 @@ def draw_string(self, x, y, text, color=theme.fg_color, bg_color=theme.bg_color)
             x = max(0, x)
         lcd.draw_string(x, y, text, color, bg_color)
 
+    def _get_text_lines(self, text, max_lines=None):
```
odudex, line 368: To avoid having two functions with same purpose, couldn't `to_lines` have this at the beginning?
```
if isinstance(text, list):
    return text
```

src/krux/pages/__init__.py
```diff
@@ -78,6 +78,8 @@
 
 EXTRA_MNEMONIC_LENGTH_FLAG = 48
 
+SWIPE_SYMBOL = "»"
```
odudex, line 81: should have left and right symbols

src/krux/pages/login.py
```diff
@@ -334,7 +321,7 @@ def _load_key_from_words(self, words, charset=LETTERS, new=False):
             )
             if not derivation_path:
                 derivation_path = key.derivation
-            wallet_info = key.fingerprint_hex_str(True) + "\n"
+            wallet_info = "\n"
```
odudex, line 324: Please use offset instead of "\n" to change position of text

src/krux/pages/settings_page.py
```diff
@@ -355,11 +357,17 @@ def category_setting(self, settings_namespace, setting):
                     DEFAULT_PADDING,
                     t("Right"),
                 )
-            self.ctx.display.draw_centered_text(
-                settings_namespace.label(setting.attr) + "\n" + str(current_category),
+            lines = self.ctx.display.draw_centered_text(
+                "\n\n" + str(current_category),
```
odudex, line 361: Please use offset instead of "\n" to change position of text

odudex:
Really liked new UI looks!
The use of `string.find()` for worries me a bit on resource usage, but it probably pays of!
And on minor changes, I think it would be better to not use "\n"s at the beginning of strings to position text, always use offset instead.
Tests are great too!

tadeubas:
Thx a lot for all the review! I'm looking into everything you mentioned and will be making commits soon. The `string.find()`  is used just on small portions of text that are presented statically, so I didn't feel or notice lag or bad UX perception. And thx again for the advice of the use of "\n" at the start, lol I forgot that we can use y_padding :man_facepalming:

src/krux/pages/__init__.py
tadeubas: Wow! What huge mistake :see_no_evil:

src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -233,7 +235,8 @@ def sign_standard_message(self, data):
 
         self.ctx.display.clear()
         self.ctx.display.draw_centered_text(
-            "SHA256:\n%s" % binascii.hexlify(message_hash).decode()
+            "SHA256:\n\n%s" % binascii.hexlify(message_hash).decode(),
+            highlight_prefix="SHA256:",
```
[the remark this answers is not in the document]
tadeubas: It is not needed, you think it is faster with ":" only?

src/krux/pages/home_pages/wallet_descriptor.py
```diff
@@ -105,7 +109,12 @@ def _load_wallet(self):
 
                 utils = Utils(self.ctx)
                 _, wallet_data = utils.load_file(
-                    (DESCRIPTOR_FILE_EXTENSION, JSON_FILE_EXTENSION), prompt=False
+                    (
+                        DESCRIPTOR_FILE_EXTENSION,
+                        JSON_FILE_EXTENSION,
+                        PUBKEY_FILE_EXTENSION,
```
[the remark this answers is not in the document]
tadeubas: I was thinking about this, we export the XPUB for singlesig as `.pub`. I tried to import it and it didn't appear, so I added the `.pub`, what do you think?

src/krux/pages/login.py
tadeubas: Can't use offset here because the text uses ` info_box=True` on `draw_hcentered_text`. The box looks broken  on the screen

src/krux/pages/new_mnemonic/dice_rolls.py
```diff
@@ -304,11 +306,14 @@ def delete_roll(buffer):
                 "".join(self.rolls) if self.num_sides < 10 else "-".join(self.rolls)
             )
             self.ctx.display.clear()
-            rolls_str = t("Rolls:") + "\n\n%s" % entropy
+            rolls_str = "\n\n%s" % entropy
```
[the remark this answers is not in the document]
tadeubas: Can't use offset here because the text uses ` info_box=True` on `draw_hcentered_text`. The box looks broken  on the screen

src/krux/pages/home_pages/wallet_descriptor.py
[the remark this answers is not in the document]
odudex: IMO we should not mix extended public keys with descriptors. Even if someone wants a singlesig descriptor, it should come from coordinator, as descriptor, just to ensure good practices.

src/krux/pages/__init__.py
odudex: That's minor!

src/krux/pages/home_pages/sign_message_ui.py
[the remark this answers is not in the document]
odudex: Lighter for sure

src/krux/pages/login.py
odudex: Ok, maybe then `info_box=True` doesn't need to be set (and an extra rectangle doesn't need to be drawn) right below when drawing the fingerprint over it

src/krux/pages/settings_page.py
tadeubas: This call is for `draw_centered_text`, it is more complicated as it calculates the offset based on the text size in lines and display height... I think if I try to optimize this, it will turn out to be slower

src/krux/pages/login.py
tadeubas: Yes, this can be done!

src/krux/pages/settings_page.py
odudex: Since drawing was separated, the best then is to change to `draw_hcentered_text`

src/krux/pages/new_mnemonic/dice_rolls.py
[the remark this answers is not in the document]
tadeubas: Called `info_box=True` only once as suggested on the other conversation, thx!

src/krux/pages/settings_page.py
tadeubas: Changed category settings to use `draw_hcentered_text` :+1:

jdlcdl:
Caught up with code review as/of 79cc8c5

These color changes look great!

I especially love the passphrase-less bip32 master fingerprint viewable while reviewing the mnemonic -- it's so important as a checksum and to help users to learn and associate it with the mnemonic.

I also like the extra vertical spacing while in Settings.

PR #524, opened 2025-03-06
Check Encryption Key Strength

odudex:
### What is this PR for?
Add a simple check and warning on encryption key strength.

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other


src/krux/pages/encryption_ui.py
```diff
@@ -59,9 +102,22 @@ def encryption_key(self):
         if key in (ESC_KEY, MENU_CONTINUE):
             return None
 
+        if creating:
```
tadeubas, line 105: I think this line has to go inside the `if key` and inside the `if creating`

jdlcdl:
This is working for me on simulator and Amigo.

src/krux/pages/encryption_ui.py
odudex: Can you elaborate? What `if key`? What's the current error/issue? How does this solve it?

odudex: Got it, and refactored. Thank!

PR #531, opened 2025-03-08
Code refactor for board configs

tadeubas:
### What is this PR for?
**PR EDITED**
- New class `kboard` to cache configs and promote code reuse

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other


odudex:
Won't it overlap with the progress bar? How about filling flash with camera entropy?

tadeubas:
> Won't it overlap with the progress bar? How about filling flash with camera entropy?

Didn't think about that :disappointed: - will came up with a solution soon, sorry :smiley:

odudex:
I believe the best is to abbreviate the translations

tadeubas:
I saw the simplified translation on other PR, but regarding  the questions about the title at bottom on Amigo, I've checked that:
- The PAGE to toggle mode is **not enabled** on "How about filling flash with camera entropy?", so this hint txt does not appear on the screen, so it is not related.
- The toggle mode option is just a hint, in other devices it appears and disappears. On Amigo I've noticed that if you press page and change mode until zoom the hint will also disappear. This makes sense because it is just a hint, like << swap >> on keypad that also disappear after a swap. So when the progress bar appears scanning a PSBT for example, the hint can disappear too because the user already had the chance to read the text and is already performing successfully the scanning operation.

What do you think?

odudex:
Can you separate this PR into two so changes can be individually discussed?

tadeubas:
Sure!

tadeubas:
Edited to just code refactor for board configs

src/krux/camera.py
```diff
@@ -87,8 +87,8 @@ def initialize_sensor(self, mode=QR_SCAN_MODE):
         """Initializes the camera"""
         sensor.reset(freq=18200000)
         self.cam_id = sensor.get_id()
-        if board.config["type"] == "cube" or (
-            board.config["type"] in ["yahboom", "wonder_mv"]
```
odudex, line 91: Yahboom and WonderMV were excluded here.

tadeubas: yes see the lines below :stuck_out_tongue_closed_eyes:  only wonder_mv and yahboom CAN_FLIP_ORIENTATION

PR #533, opened 2025-03-08
Reset derivation (even if it was set custom for Miniscript)

jdlcdl:
### What is this PR for?
A minor issue found prior to release:
* If Key.policy is Miniscript, then custom_derivation was true, so it became sticky when customizing:
* became problematic if customizing for network between main/test w/o altering Policy... would not update derivation.

This pr changes key:
* no notion of "custom_derivation", derivation is like other customizable attributes, will be defaulted if initialized empty,

And wallet_settings.customize_wallet:
* any change of network, policy, script_type, account will unset current derivation, so that it gets re-defaulted
* if changing derivation (currently only for Miniscript) then it won't get re-defaulted.

- [x] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/key.py
+++ b/src/krux/key.py
@@ -113,7 +113,7 @@ class Key:
         passphrase="",
         account_index=0,
         script_type=P2WPKH,
-        custom_derivation="",
+        derivation="",
     ):
         self.mnemonic = mnemonic
         self.policy_type = policy_type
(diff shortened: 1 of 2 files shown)


odudex:
Perfect! Thank you for catching and fixing!

PR #535, opened 2025-03-10
binary grid scan 24w validation fix

odudex:
### What is this PR for?
Fix and eventual error being raised while scanning 24 words binary grids.
Thanks @tadeubas for catching!

- [X] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other


tadeubas:
Just for record, the bug was: 
- Trying to scan a Tiny Seed / OneKey / Binary grid with 24w. It was failing on the second part (13-24w) showing on the screen `ValueError('Invalid Binary Grid number',)`. 

Tested here on Amigo and confirmed that the bug was solved! Nice and fast fix @odudex  :+1: :+1:

PR #538, opened 2025-03-11
Hide menu New Mnemonic when hide_mnemonic enabled

tadeubas:
### What is this PR for?
The Hide Mnemonic settings will hide any BIP39 mnemonics, even newly created ones. So there is no point in enabling the New Mnemonic menu because the user will not see the mnemonic if hide is enabled.

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


odudex:
Nice, Thank you!

PR #541, opened 2025-03-16
Addresses displayed in groups of 4

tadeubas:
### What is this PR for?
- Addresses are displayed in groups of 4, separated by spaces, for better readability.
- Change affect scan/list address and sign PSBT

### Changes made to:
- [x] Code
- [x] Tests
- [x] Docs
- [x] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


src/krux/psbt.py
```diff
@@ -275,9 +275,38 @@ def _classify_output(self, out_policy, out):
             return SELF_TRANSFER
         return SPEND
 
+    def _get_resume_fee(self, inp_amount, out_amount, output_policy_count):
```
odudex, line 278: One detail Jeff used to ask me: Please don't add "get" in function names, to be more "pythonic"

tadeubas: This is just an internal function, it even don't need comments because of the `_` prefix, so I don't think it needs to be more or less "pythonic", but... what do you suggest to be more "pythonic"?

PR #545, opened 2025-03-18
Coverage 100% of src/krux/pages/new_mnemonic/dice_rolls.py

qlrd:
### What is this PR for?

This commit aim to target 100% on dice_rolls.py page by adding two tests, with same situation, in two different devices:

* The user begin the creation of an entropy in m5stickv but, in middle time, stop the procedure (when krux ask "proceed" and the user choose "No");

* The user begin the creation of an entropy in Amigo, but in middle time, stop the procedure (when krux ask "proceed" and the user choose "No).

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/tests/pages/new_mnemonic/test_dice_rolls.py
+++ b/tests/pages/new_mnemonic/test_dice_rolls.py
@@ -75,6 +75,53 @@ GOOD_ROLLS_SEQUENCE = [
 ]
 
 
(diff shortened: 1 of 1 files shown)


odudex:
Thank you @qlrd !!

PR #548, opened 2025-03-20
Removed deadcode verification

### LINKED ISSUE #547: [Other] Considerations on checking remaining bits seem to always be padding (null) in bbqr.py
To complete the 100% coverage on current krux/src/bbqr.py it's needed to cover [krux/src/bbqr.py:196](https://github.com/selfcustody/krux/blob/0375282894830c3f2e7f4a325e3e745e2340c4b9/src/krux/bbqr.py#L196). Researching on how to call it on tests (both reading code and talking with AI and @jdlcdl), i started to consider two scenarios:

* an edge case where this line is called, but if is called this can lead to another situation,  where `decode(encoded(data)) != data` (data corruption);

* The code is so well implemented that this line will never be reached;

The first case is a laborious one to have a consistent proof; so i created a [parallel project](https://github.com/qlrd/krux-fuzz) where, if is possible, get the [edge-case](https://github.com/qlrd/krux-fuzz/blob/main/fuzz/fuzz_targets/diff_base32.py) comparing it's output to ruts's [rust's base32](ttps://crates.io/crates/base32). I'm running at least 2 hours and nothing was found yet. To the second case, this made me thought that it's worth to either:

* remove this line from code; or

* instead append a `deadcode`, raise an Exception, because if it's a deadcode, it should be never be appended.

qlrd:
### What is this PR for?

Fix #547 

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other: code remotion that is unreachable

### Details

This commit remove the checking of remaining bytes in Rfc4648 implementation of krux/src/bbqr.py::base32_decode_stream function.

* The implementation appear to be well coded and as the condition seems to be logically unreachable;

* the base_decode_stream passed by some diff fuzz tests and appear to not have any issues, attesting to the correctness of the implementation.

### DIFF
--- a/src/krux/bbqr.py
+++ b/src/krux/bbqr.py
@@ -188,13 +188,6 @@ def base32_decode_stream(encoded_str):
             decoded_bytes.append((buffer >> bits_left) & 0xFF)
             buffer &= (1 << bits_left) - 1  # Keep only the remaining bits
 
-    # Process any remaining bits if they form a valid byte
-    if 0 < bits_left < 8:
-        remaining_byte = (buffer << (8 - bits_left)) & 0xFF
-        if remaining_byte != 0:
-            # Dead code? remaining bits seem to always be padding (null)
-            decoded_bytes.append(remaining_byte)
-
     return bytes(decoded_bytes)
 
 


qlrd:
```diff
diff --git a/src/krux/bbqr.py b/src/krux/bbqr.py
index 17fa3c41..ed7429ba 100644
--- a/src/krux/bbqr.py
+++ b/src/krux/bbqr.py
@@ -191,9 +191,6 @@ def base32_decode_stream(encoded_str):
     # Process any remaining bits if they form a valid byte
     if 0 < bits_left < 8:
         remaining_byte = (buffer << (8 - bits_left)) & 0xFF
-        if remaining_byte != 0:
-            # Dead code? remaining bits seem to always be padding (null)
-            decoded_bytes.append(remaining_byte)
```

the 2 last lines in red part are never called, but those outside red are called.

Maybe's its worth to remove them too?

odudex:
Yes, you can remove 2 lines above too, as the created variable is not used.

qlrd:
> Yes, you can remove 2 lines above too, as the created variable is not used.

Removed it

odudex:
Before merging, it's important to test features that might be affected by the changes in the device. While I don't believe this is the case here, note that code running on MicroPython may behave differently than on a PC. In this instance, simply load some BBQr PSBTs and descriptors to verify functionality.
In case you already did it, just let me know.

qlrd:
> e affected by the changes in the device. While I don't believe this is the case here, note that code running

Did not ran yet, neither on simulator or device. Thinking if there's a way to make it on a standard way.

odudex:
Some simple tests with single-sig should be enough. Having it merged on develop we end up testing it more alongside the other features being developed.

qlrd:
> Some simple tests with single-sig should be enough. Having it merged on develop we end up testing it more alongside the other features being developed.

Signed a @jdlcdl's grampa WPKH testnet transaction. The unsigned hex is:

```bash
0200000001e41fbe3fe1e636eaf1f9adddb0205735b5525b4828155785cf832ccd7846f0550000000000fdffffff01c52d000000000000160014447fde1e37d97255b5821d2dee816e8f18f6bac9d0ad0300
```

And the signed hex is:

```bash
02000000000101e41fbe3fe1e636eaf1f9adddb0205735b5525b4828155785cf832ccd7846f0550000000000fdffffff01c52d000000000000160014447fde1e37d97255b5821d2dee816e8f18f6bac90247304402203b439d5a5e1a0bc0fa249742e360e278d62c395cb807385a8f05ac865d0aaead02205b9fa4c02dd485d7e7c2da96fd3d89895f7e8874ab17e71901c58dc6eb1481ec012102c5d0e68f0b4bf4538d505f783324163a872c65a1a0ea0181026729da3e8308bfd0ad0300
```

qlrd:
This one is grampa's WPKH wallet using BBQr:

```bash
❯ echo "020000000001038b603a1f0e332e8e4d7e0c805ce9c0b1dcb79e2780dc5aeeadd193ad428d54b10000000000fdffffff9830478ce718bd10a8929236ff2072c48df61cfa7dbc4ef4a4000247c1c4ffe00000000000fdffffffffa68a0eccb1482890de2cc7626a2f744ae3ad974bae1abd77d4c5f9cbe523ba0000000000fdffffff010787000000000000220020a243748408ddbbffec51655210a9bf07772a4a64b9be2894e5e467eb09c45b4e024730440220192d766c7c4aac397827183dc131739556e1777cbbb30545ebff5b8689dc2c4c0220212c70e61db14d35c4387cc7adf3ca5bf5bbf1de1846ecace2f1a370d007f780012102adc555c0336a7b42b3234618aba94d07d93d8e35c19ba021560becb5c87ae37f0247304402203d3d1f8e55647096732bf8d82bd4188389055f192637c7c8a82d970616cf6fe002200e7d5865636bad9cc60a45d847d0939ddc1fe42f7d64778e01998309d21bf624012103be216cf23498cd317d2f44d9ad56a31dc263f4ffcd08e707e6a907ba9be74e0f0247304402201307aec07706d9afe1110d1bf0a118d25b712ea8f9c45c4213fbcbc328d296440220691d147a810ac72a4ab87f7c19c4c2d134a77572a09b5bdf62462ac1bf07dea101210276fea43c1c6f205bc25a87c546b4007d2e51d5e455e544381d087ca110506dd376ae0300" > multisig.txt
```

This one is using the process saving to SDCard:

```bash
❯ echo "020000000001038b603a1f0e332e8e4d7e0c805ce9c0b1dcb79e2780dc5aeeadd193ad428d54b10000000000fdffffff9830478ce718bd10a8929236ff2072c48df61cfa7dbc4ef4a4000247c1c4ffe00000000000fdffffffffa68a0eccb1482890de2cc7626a2f744ae3ad974bae1abd77d4c5f9cbe523ba0000000000fdffffff010787000000000000220020a243748408ddbbffec51655210a9bf07772a4a64b9be2894e5e467eb09c45b4e024730440220192d766c7c4aac397827183dc131739556e1777cbbb30545ebff5b8689dc2c4c0220212c70e61db14d35c4387cc7adf3ca5bf5bbf1de1846ecace2f1a370d007f780012102adc555c0336a7b42b3234618aba94d07d93d8e35c19ba021560becb5c87ae37f0247304402203d3d1f8e55647096732bf8d82bd4188389055f192637c7c8a82d970616cf6fe002200e7d5865636bad9cc60a45d847d0939ddc1fe42f7d64778e01998309d21bf624012103be216cf23498cd317d2f44d9ad56a31dc263f4ffcd08e707e6a907ba9be74e0f0247304402201307aec07706d9afe1110d1bf0a118d25b712ea8f9c45c4213fbcbc328d296440220691d147a810ac72a4ab87f7c19c4c2d134a77572a09b5bdf62462ac1bf07dea101210276fea43c1c6f205bc25a87c546b4007d2e51d5e455e544381d087ca110506dd376ae0300" > multisig2.txt
```

Then i compared
```bash
❯ if diff multisig.txt multisig2.txt > /dev/null; then
    echo "Worked"
else
    echo "Failed"
fi
Worked
```

qlrd:
@tadeubas, @jdlcdl, wdyt?

qlrd:
Added [BBQr::utils::encode_data](https://github.com/coinkite/BBQr/blob/8e90b1d2efe699ee2c9804ca564bf3fdfd74f346/python/bbqr/utils.py#L32) to [krux-fuzz](https://github.com/qlrd/krux-fuzz/blob/66e38477b97387be87d1a565bb31da34df7028d9/fuzz/fuzz_targets/diff_base32.rs#L23) tests. I will run it for some days before considering a merge.

It's worth to mention that the `encode_data` function (similar in a way to krux's `base32_encode_stream`) do not do the `padding` (in truth, it removes the padding).

But it's. expected that the non-padding encoding is the same.

odudex:
Thank you!

PR #552, opened 2025-04-01
Cover 100% of test_bip39.py

qlrd:
### What is this PR for?

The test_bip39.py was ignoring the case where the k_mnemonic_bytes could be called with ignore_checksum = True.

This commit just add this situation.

It's worth to mention that i tested with many bip39 vectors with [krux-fuzz](https://github.com/qlrd/krux-fuzz/blob/main/fuzz/fuzz_targets/diff_bip39.rs)

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other: coverage

### DIFF
--- a/tests/test_bip39.py
+++ b/tests/test_bip39.py
@@ -1,8 +1,10 @@
-from krux import bip39 as kruxbip39
+import secrets
+
+import pytest
 from embit import bip39
 from embit.wordlists.bip39 import WORDLIST
-import secrets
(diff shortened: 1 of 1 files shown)


odudex:
Thank you! Great to know your are running fuzz on this type of code!

PR #554, opened 2025-04-02
Fix false positive call on mocking is_on method

### LINKED ISSUE #553: [Bug] The wonder_mv fixture passed to test_toggle_on_wonder_mv is not correctly setting the board type in tests/test_light.py
#### Describe what it is

During a review on [tests/test_light.py::test_toggle_on_wonder_mv function](https://github.com/selfcustody/krux/blob/d66f21e1ef6352a4a9bc639fd13b761add6b97db/tests/test_light.py#L93C5-L93C29), verified the [if clause kboard.is_wonder_mv in src/krux/light.py::IOCircuit::in_on](https://github.com/selfcustody/krux/blob/d66f21e1ef6352a4a9bc639fd13b761add6b97db/src/krux/light.py#L38) isn't beign called properly.

With some scrutinity, i found that this test is called as if was a `m5stickv` device, not a `wonder_mv`. I just researched on this test, but i suspect that can occurs in other parts of tests suites, i.e, the wonder_mv fixture (or other fixtures) passed to test_toggle_on_wonder_mv and other functions) isnt beign correctly setting the board type.

### Steps to reproduce

apply this diff

```diff
diff --git a/src/krux/light.py b/src/krux/light.py
index 070f9c1c..d1c82511 100644
--- a/src/krux/light.py
+++ b/src/krux/light.py
 
@@ -35,6 +36,18 @@ class IOCircuit:
 
     def is_on(self):
         """Returns a boolean indicating if the circuit is currently on"""
+        print(
+            " ".join(
+                [
+                    f"is_m5stickv: {kboard.is_m5stickv}",
+                    f"is_amigo: {kboard.is_amigo}",
+                    f"is_cube: {kboard.is_cube}",
+                    f"is_bit: {kboard.is_bit}",
+                    f"is_yahboom: {kboard.is_yahboom}",
+                    f"is_wonder_mv: {kboard.is_wonder_mv}",
+                ]
+            )
+        )
         if kboard.is_wonder_mv:
             return self.circuit.value() == 1
         return self.circuit.value() == 0
diff --git a/tests/test_light.py b/tests/test_light.py
index 3a4d34cb..cd6fb505 100644
--- a/tests/test_light.py
+++ b/tests/test_light.py
 
@@ -103,6 +104,10 @@ def test_toggle_on_wonder_mv(mocker, wonder_mv):
     light.toggle()
     light.turn_off.assert_called()
 
+    # This is a placeholder for the actual test logic
+    # just to demonstrate the false positive case
+    assert False
+
     # Toggle again, now from off
     mocker.patch.object(light, "is_on", new=lambda: False)
     light.toggle()

```

Then run the test under review:

```bash
poetry run pytest --show-capture all --capture tee-sys tests/test_light.py
```

The expected result is that the actual device being called on `test_toggle_on_wonder_mv` is the `m5stickv`:

```bash
tests/test_light.py .is_m5stickv: True is_amigo: False is_cube: False is_bit: False is_yahboom: False is_wonder_mv: False
.....F                                                                                                                                                                                                            [100%]

===========FAILURES 
___________ test_toggle_on_wonder_mv

mocker = <pytest_mock.plugin.MockerFixture object at 0x106c794c0>, wonder_mv = None

    def test_toggle_on_wonder_mv(mocker, wonder_mv):
        mock_modules(mocker)
        from krux.light import Light

        light = Light()
        mocker.spy(light, "turn_off")
        mocker.spy(light, "turn_on")

        # Toggle from on
        mocker.patch.object(light, "is_on", new=lambda: True)
        light.toggle()
        light.turn_off.assert_called()

        # This is a placeholder for the actual test logic
        # just to demonstrate the false positive case
>       assert False
E       assert False

tests/test_light.py:109: AssertionError
```

qlrd:
### What is this PR for?

Fix #553

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other: coverage

### Context

Previsouly, the `tests/test_light.py` was patching the is_on method and the real method is never executed, and thus the test pass but not hit the tested code.

This is classic situation where what we want to mock isnt the thing, but the dependencies of the thing (in this case, what is beign called inside what we want to test).

Instead mock the is_on method, is necessary to mock the `IOCircuit::circuit.value()` call (that is really what we want to test).

This commit changes the toggles in mocker from `ligth.is_on` to `light.circuit.value`.

### DIFF
--- a/tests/test_light.py
+++ b/tests/test_light.py
@@ -99,11 +99,11 @@ def test_toggle_on_wonder_mv(mocker, wonder_mv):
     mocker.spy(light, "turn_on")
 
     # Toggle from on
-    mocker.patch.object(light, "is_on", new=lambda: True)
+    mocker.patch.object(light.circuit, "value", return_value=1)
     light.toggle()
     light.turn_off.assert_called()
 
     # Toggle again, now from off
-    mocker.patch.object(light, "is_on", new=lambda: False)
+    mocker.patch.object(light.circuit, "value", return_value=0)
     light.toggle()
     light.turn_on.assert_called()


odudex:
Good catch! Thank you!

PR #569, opened 2025-04-16
Remove deadcode in src/pages/menomic_editor.py

### LINKED ISSUE #568: [Other] alt_wordlist parameter deadcode on src/pages/mnemonic_editor.py
### Summary

The `alt_wordlist` parameter in `MnemonicEditor` methods such as `compute_search_ranges()`, `autocomplete()`, and `possible_letters()` is only meaningfully used in one specific case: **when editing the final word of a new mnemonic**.

In all other interactive flows (`edit()`, `edit_word()`), it is ignored, making much of the associated code effectively dead — unless forcefully tested with patched inputs.

---

### Details

The only path that uses `alt_wordlist` is:

```python
if self.new_mnemonic and index == self.mnemonic_length - 1:
    final_words = Key.get_final_word_candidates(...)
    word = self.capture_from_keypad(
        ...,
        lambda x: self.autocomplete(x, final_words),
        lambda x: self.possible_letters(x, final_words),
    )
```

The following logic in `autocomplete()` is extremely unlikely to be triggered in real-world use:

```python
# line 90
if letter not in self.search_ranges:
    return None

# or in line 113
if letter not in self.search_ranges:
    return ""
```

Even when the final word uses a filtered list (via `get_final_word_candidates()`), that list is still a subset of the official BIP39 wordlist. The virtual keyboard only presents valid next letters (from `possible_letters()`), which ensures users cannot enter a letter outside the `search_ranges`. This makes the above line nearly impossible to hit during interactive usage — only patching tests with manual overrides can trigger it.

---

### Why It Matters

1. The `alt_wordlist` path, and the `search_ranges` guard, are not reachable through standard UI use and only serve a very narrow internal case.

2. Tests written to “cover” this code require unrealistic patches that will never occur in production.

3. If the project has no intent to support fully custom wordlists (e.g., SLIP-39, Diceware), `alt_wordlist` should not be treated as a general-purpose feature.

4. Consider removing or renaming `alt_wordlist` to something more purpose-specific (e.g., `final_wordlist`) to reflect its internal use.

5. Consider removing the `if letter not in self.search_ranges:` check entirely or marking it with a comment like `# defensive, unreachable in UI`.

### Example of Unrealistic Test

To cover the unreachable code path above, this test was added (with multiple patches to `MnemonicEditor`):

```diff
+CUSTOM_WORDLIST = [
+    "academic",
+    "banjo",
+    "blooming",
+    "cobweb",
+    "glimmer",
+    "harvest",
+    "heaven",
+    "jackal",
+    "myriad",
+    "nectar",
+    "quantum",
+    "salary",
+    "sharply",
+    "stinger",
+]
+
+def test_edit_mnemonic_autocomplete_prefix_not_in_search_ranges(mocker, amigo):
+    from krux.pages import ESC_KEY
+    from krux.pages.mnemonic_editor import MnemonicEditor
+
+    ctx = create_ctx(mocker, [])
+    mnemonic_editor = MnemonicEditor(ctx, TEST_12_WORD_MNEMONIC)
+
+    mnemonic_editor.autocomplete = lambda p, alt=None: MnemonicEditor.autocomplete(mnemonic_editor, p, CUSTOM_WORDLIST)
+    mnemonic_editor.possible_letters = lambda p, alt=None: MnemonicEditor.possible_letters(mnemonic_editor, p, CUSTOM_WORDLIST)
+    mnemonic_editor.capture_from_keypad = lambda *_: ESC_KEY
+
+    result = mnemonic_editor.edit_word(0)
+    assert result is None
```

While this test could technically happen, it requires simulated behavior that would never occur during actual usage.

qlrd:
### What is this PR for?

Fix #568

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other: remotion of possible deadcode.

### Details

The `alt_wordlist` parameter in `MnemonicEditor` methods such as `compute_search_ranges()`, `autocomplete()` , and `possible_letters()`  is only meaningfully used in one specific case: when editing the final word of a new mnemonic.

In all other interactive flows (`edit()` , `edit_word()`), it is ignored, making calls such `if letter not in self.search_ranges` in line 90 and 113 effectively dead — unless forcefully tested with patched inputs.

The virtual keyboard only presents valid next letters (from `possible_letters()`), which ensures users cannot enter a letter outside the search_ranges. This makes the above line nearly impossible to hit during interactive usage — only patching tests with manual overrides can trigger it.

This commit remove the calls on these checks.

### DIFF
--- a/src/krux/pages/mnemonic_editor.py
+++ b/src/krux/pages/mnemonic_editor.py
@@ -86,8 +86,6 @@ class MnemonicEditor(Page):
             wordlist = WORDLIST
         if len(prefix) > 0:
             letter = prefix[0]
-            if letter not in self.search_ranges:
-                return None
             start, stop = self.search_ranges[letter]
             matching_words = list(
                 filter(
@@ -109,8 +107,6 @@ class MnemonicEditor(Page):
         if len(prefix) == 0:
             return self.search_ranges.keys()
         letter = prefix[0]
-        if letter not in self.search_ranges:
-            return ""
         start, stop = self.search_ranges[letter]
         return {
             word[len(prefix)]


odudex:
Thank you!

PR #573, opened 2025-04-18
Cover 100% src/krux/key.py

qlrd:
### What is this PR for?

* Add a verbose message when an invalid policy type is used to create a bip39 key in get_default_descriptor static method;

* add a fallback import case from urandom to random on tests/test_key.py/

* add a raise exception case of an invalid policy type on tests/test_key.py.

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other: cover 100% `src/krux/key.py`.


src/krux/key.py
```diff
@@ -234,7 +234,14 @@ def get_default_derivation(policy_type, network, account=0, script_type=P2WPKH):
             return DER_MULTI % (MULTISIG_SCRIPT_PURPOSE, network["bip32"], account)
         if policy_type == TYPE_MINISCRIPT:
             return DER_MINISCRIPT % (MINISCRIPT_PURPOSE, network["bip32"], account)
-        raise ValueError("Invalid policy type")
+
+        msg = "Invalid policy type: %s.\nValid values are: %s, %s, %s." % (
```
odudex, line 238: Since the person who will debug this error will have to look at the code to check what types 0,1,2 are, the second sentence is not necessary. Just let "Invalid policy type: %s"

tests/test_key.py
```diff
@@ -85,6 +85,57 @@ def mock_modules(mocker):
     mocker.patch("krux.key.hexlify", new=mocker.MagicMock(wraps=binascii.hexlify))
 
 
+def test_missing_urandom():
```
odudex, line 88: Any specific reason to test a missing module, and only this one?

tests/test_key.py
```diff
@@ -85,6 +85,57 @@ def mock_modules(mocker):
...
+
+        assert krux.key.random is random
+
+    # Restore the original import function
+    # so that other tests are not affected
+    finally:
+        builtins.__import__ = original_import
+
+
+def test_init_fail_wrong_default_derivation(mocker, m5stickv, tdata):
```
odudex, line 128: Test name and variable names are no coherent to the test subject: Policy type

tests/test_key.py
qlrd: > Any specific reason to test a missing module, and only this one?

Yes, the lines 25-28 at the begining of `src/krux/key.py`:

```python
try:
    import urandom as random
except:
    import random
```

wasn't being covered during the tests. 

This `test_missing_urandom` was only coded to cover these lines. But from what i readed in the source code (and tested), generally the `urandom` is always being imported leaving the `random` forgot (making it a potential dead code). Maybe it's worth to remove this exception clause (and the test itself)?

tests/test_key.py
qlrd: Fair enough.

I put as it because what really throw the exception is the `SDHandler.get_default_derivation` when called by `__init__`. I will change that too.

src/krux/key.py
qlrd: fixed

tests/test_key.py
qlrd: fixed

tests/test_key.py
```diff
@@ -85,6 +85,57 @@ def mock_modules(mocker):
...
+    # so that other tests are not affected
+    finally:
+        builtins.__import__ = original_import
+
+
+def test_init_fail_wrong_policy_type(mocker, m5stickv, tdata):
+    mock_modules(mocker)
+    from krux.key import Key
+
+    TYPE_WRONGSIG = 3
```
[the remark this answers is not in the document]
qlrd: fixed, also changed the function's name to `test_init_fail_unknown_policy_type` and added some comments.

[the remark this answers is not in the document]
odudex: It seems now there's a double definition of the test function.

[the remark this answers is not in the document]
qlrd: My bad, patched the commit wrong here

odudex:
I've been thinking, for the sake of simplicity and standardization, we could remove the try/except with `random`, just use `urandom`, and remove the specific test too. This align with other cases like `uhashlib`, were we use the standard micropython lib, and mock it with respective complete implementations on tests, simulator and Android.

qlrd:
> I've been thinking, for the sake of simplicity and standardization, we could remove the try/except with `random`, just use `urandom`, and remove the specific test too. This align with other cases like `uhashlib`, were we use the standard micropython lib, and mock it with respective complete implementations on tests, simulator and Android.

Agreed, this resolves an unecessary test, some lint and less code to maintain

qlrd:
> I've been thinking, for the sake of simplicity and standardization, we could remove the try/except with `random`, just use `urandom`, and remove the specific test too. This align with other cases like `uhashlib`, were we use the standard micropython lib, and mock it with respective complete implementations on tests, simulator and Android.

fixed

odudex:
> fixed

Did you test the simulator? I believe a mock for `urandom` which loads `random` will be needed

odudex:
Just noticed @tadeubas did the needed changes in #588.

odudex:
Thank you!

PR #581, opened 2025-04-22
SD Firmware Update Prompt Fix

odudex:
### What is this PR for?
1. **Signature‑check ordering**  
   - **Before**: UI prompted “Install now?” _then_ verified the firmware signature.  
   - **After**: Signature is verified _before_ the install prompt—now matching the docs.

2. **Hardware‑accelerated SHA‑256**  
   - Updated the hasher to use the on‑chip SHA‑256 accelerator.  
   - Verification is ~1 s faster as a result.

#### Why
@crwatkins pointed out that our UI flow contradicted the documentation: Users were being asked to install before signature verification. While the verification was being later performed, this PR better reflect the documented workflow, with signatures being performed before user is prompted to update.

#### Impact

- Users can no longer click “Install” on firmware until it’s been verified.  
- Validation is faster, shaving approximately one second off most updates.

### Changes made to:
- [X] Code
- [X] Tests
- [ ] Docs - No need
- [X] CHANGELOG

- [X] Yes - Amigo

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other: UI behavior fix


src/krux/firmware.py
```diff
@@ -204,21 +204,6 @@ def status_text(text, highlight_prefix=""):
...
-        + "\n\n\n"
-        + t("Install?"),
-        ":",
-    )
-    inp.buttons_active = True
-    if inp.wait_for_button() in (BUTTON_PAGE, BUTTON_PAGE_PREV):
-        display.clear()
-        inp.wait_for_release()  # Wait for button release loading inputs on context
-        return False
-
```
tadeubas, line 221: The only issue I see removing this, is the device "frozen" with a black screen for a longer time. To fix this we could just display "Processing.." here.

odudex: Where exactly does a frozen black screen occur?

src/krux/firmware.py
```diff
@@ -250,6 +235,21 @@ def status_text(text, highlight_prefix=""):
...
+        + "\n\n\n"
+        + t("Install?"),
+        ":",
+    )
+    inp.buttons_active = True
+    if inp.wait_for_button() in (BUTTON_PAGE, BUTTON_PAGE_PREV):
+        display.clear()
+        inp.wait_for_release()  # Wait for button release loading inputs on context
+        return False
+
```
tadeubas, line 252: The general coverage also went down 3 lines. This may be due the test when user negates installing the firmware is missing mocks for a valid file / signature

odudex: Yes, I'll fix coverage before setting it as ready to review

src/krux/firmware.py
tadeubas: My bad! There is already a msg on line 201 for that!

> status_text(t("New firmware detected.") + "\n\n" + t("Verifying.."))

tadeubas:
Tested on Yahboom with bad signature, ignore install and install, everything fine. Just opened a PR with translations update https://github.com/odudex/krux/pull/49

tadeubas:
Plz merge this PR to develop as it is now needed for #583

PR #589, opened 2025-05-01
Tools > Print Test QR asks before printing..

tadeubas:
### What is this PR for?
- User prompted before printing test QR code.
- Small code optimizations/fixes regarding `qr_format` initialization (Ex: `Utils.print_standard_qr("Some data")` would raise an exception because qr_format was initialized with `None`).

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [x] CHANGELOG

- [x] Yes, Yahboom

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Addresses are displayed in groups of 4, separated by spaces, for better readabil
 
 ### Other Bug Fixes and Optimizations
 - Fix printing words instead of numbers when using "Backup Mnemonic > Other formats > Numbers".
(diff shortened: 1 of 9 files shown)


odudex:
I don't know how the behavior is now, but before changes it was raising an error when a printer driver was not set ( specifically when testing the printer).
"Printer driver not set!"
We need this behavior to continue and this message to be shown, so users know what they have to do in order to get printer working.

tadeubas:
Wow, nice catch! :clap:

fixed with  `da3c487`

PR #557, opened 2025-04-05
Qr cnc svg

spatcho:
### What is this PR for?

Print qr to cnc, write to file : adapt code to latest qr_code array structure

Print qr to cnc, write directly to cnc  : uncomment, test, and adapt code to fit an openbuilds grbl 1.1 cnc, see comments on cnc.py

Save qr as svg image

### Changes made to:
- [X] Code
- [ ] Tests
- [ ] Docs
- [X] CHANGELOG

### What is the purpose of this pull request?
- [X] Bug fix
- [X] New feature
- [ ] Docs update
- [ ] Other

PR #593, opened 2025-05-07
Specific gcodes for Laser

qlrd:
### What is this PR for?

Partially fix of #557 

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other

### Details

This PR is a followup of #557 towards laser options on cnc, trying to preserve at maximum the great work made by @spatcho, while give some organization to commit to better review. 

* Add "machine type" choice: cnc or laser;
* add "tool speed" choice: pwn level of power, from 0 (0%) to 1000
(100%), only used in laser mode but could be used for variable speed
routers;
* no Z moves when laser selected;
* specific gcodes for laser;
* tested on opt laser xt8 as add on to openbuilds cnc and with
openbuilds blackbox x4 controller;


qlrd:
There's a issue with CNC menu organization. It have so many options that all items was not shown:

```python
def label(self, attr):
        """Returns a label for UI when given a setting name or namespace"""
        return {
            "invert": t("Invert"),
            "cut_method": t("Cut Method"),
            "unit": t("Unit"),
            "flute_diameter": t("Flute Diameter"),
            "plunge_rate": t("Plunge Rate"),
            "feed_rate": t("Feed Rate"),
            "cut_depth": t("Cut Depth"),
            "depth_per_pass": t("Depth Per Pass"),
            "part_size": t("Part Size"),
            "border_padding": t("Border Padding"),
            "head_type": t("Tool head type"),
            "head_power": t("Tool head speed / power"),
            "grbl": "GRBL",
        }[attr]

```

Maybe is better to elaborate them on subcategories. I thought in this:

* create a "Head" and `head_type` and `head_power` goes there;
* create a submenu "Rates" (or any other name) and `flute_diameter`, `feed_rate`, `plunge_rate, `cut_depth`, `depth_per_pass`, `part_size` and `border_padding` goes there

tadeubas:
I believe the effort required may outweigh the potential benefits in this case. Very few users make use of printers, and even fewer use CNC machines, so this menu is likely to see minimal usage. Additionally we will end with a few more translations.

qlrd:
> I believe the effort required may outweigh the potential benefits in this case. Very few users make use of printers, and even fewer use CNC machines, so this menu is likely to see minimal usage. Additionally we will end with a few more translations.

You say for entire [requested patch](https://github.com/selfcustody/krux/pull/557#issuecomment-2851093535) or for menu?

tadeubas:
Menu only

qlrd:
> Menu only

Ah ok, didn't pushed the changes about menu, just made as proof of concept.

spatcho:
When does it goes to develop branch ? I cannot test real life milling right now, I am upgrading the hardware and it take longer than expected, but I can test laser.

qlrd:
> When does it goes to develop branch ? I cannot test real life milling right now, I am upgrading the hardware and it take longer than expected, but I can test laser.

Hi spatcho, i'm waiting for peer review to be aproved. 

Tried to keep the maxium as possible your original code, with minor changes to fit linting and coverage.

I put one about in `transmit` function. IMO, `transmit` should raise a TimeoutError exception if after 10 seconds the transmition fails. But  IDK if this is a desired behaviour.

spatcho:
> > When does it goes to develop branch ? I cannot test real life milling right now, I am upgrading the hardware and it take longer than expected, but I can test laser.
> 
> Hi spatcho, i'm waiting for peer review to be aproved.
> 
> Tried to keep the maxium as possible your original code, with minor changes to fit linting and coverage.
> 
> I put one about in `transmit` function. IMO, `transmit` should raise a TimeoutError exception if after 10 seconds the transmition fails. But IDK if this is a desired behaviour.

If the transmit fail the 10 times retry loop, the processing would just hang, and an exception would be more informative. But I didn't encounter hang issues after adding that timeout loop. Best would be to better know why the transmit fail, I didn't go further on that.

qlrd:
> transmit fail the 10 times retry loop, the processing would just hang, and an exception would be more informative. But I didn't encounter hang issues after adding that timeout loop. Best would be to better know why the transmit fail, I didn't go further on that.

True, a "what" could be a little improvement, not the perfect, but is something. Maybe we can improve the "why" in another PR. WDYT @odudex, @tadeubas, @spatcho?

src/krux/krux_settings.py
```diff
@@ -215,6 +217,8 @@ def label(self, attr):
             "depth_per_pass": t("Depth Per Pass"),
             "part_size": t("Part Size"),
             "border_padding": t("Border Padding"),
+            "head_type": t("Tool head type"),
+            "head_power": t("Tool head speed / power"),
```
odudex, line 221: Speed, in which head moves, is a different property, covered by "feed_rate". I would replace this whole sentence with "Power".

src/krux/krux_settings.py
```diff
@@ -198,6 +198,8 @@ class CNCSettings(SettingsNamespace):
     depth_per_pass = NumberSetting(float, "depth_per_pass", 0.03125, [0.0001, 10000])
     part_size = NumberSetting(float, "part_size", 3.5, [0.0001, 10000])
     border_padding = NumberSetting(float, "border_padding", 0.0625, [0.0001, 10000])
+    head_type = CategorySetting("head_type", "router", ["router", "laser"])
```
tadeubas, line 201: I would love to see the string "router" and "laser" as CONSTS, as we did with `CNC_FILE_DRIVER`. My suggestion would be `CNC_HEAD_ROUTER` and `CNC_HEAD_LASER`

src/krux/printers/cnc.py
```diff
@@ -70,12 +70,26 @@ def print_string(self, text):
             )
         )
 
+    def on_xy_gcode(self, gcode):
+        """Handle xy gcode preprocessing"""
+        if Settings().hardware.printer.cnc.head_type == "laser":
```
tadeubas, line 75: `CNC_HEAD_LASER`

src/krux/printers/cnc.py
```diff
@@ -70,12 +70,26 @@ def print_string(self, text):
...
+
     def clear(self):
         """Clears the printer's memory, resetting it"""
         raise NotImplementedError(
             "Must implement 'clear' method for {}".format(self.__class__.__name__)
         )
 
+    def on_z_gcode(self, gcode):
+        """Handle z gcode preprocessing"""
+        if Settings().hardware.printer.cnc.head_type == "router":
```
tadeubas, line 90: `CNC_HEAD_ROUTER`

src/krux/printers/cnc.py
```diff
@@ -106,13 +120,15 @@ def print_qr_code(self, qr_code):
         self.on_gcode("G54")  # coord system 1
         self.on_gcode("G90")  # non-incremental motion
         self.on_gcode("G94")  # feed/minute mode
+        if Settings().hardware.printer.cnc.head_type == "laser":
```
tadeubas, line 123: `CNC_HEAD_LASER`

src/krux/krux_settings.py
```diff
@@ -198,6 +200,10 @@ class CNCSettings(SettingsNamespace):
     depth_per_pass = NumberSetting(float, "depth_per_pass", 0.03125, [0.0001, 10000])
     part_size = NumberSetting(float, "part_size", 3.5, [0.0001, 10000])
     border_padding = NumberSetting(float, "border_padding", 0.0625, [0.0001, 10000])
+    head_type = CategorySetting(
+        "head_type", "router", [CNC_HEAD_ROUTER, CNC_HEAD_LASER]
```
tadeubas, line 204: here "router" can be replaced by CNC_HEAD_ROUTER

src/krux/krux_settings.py
```diff
@@ -215,6 +221,8 @@ def label(self, attr):
             "depth_per_pass": t("Depth Per Pass"),
             "part_size": t("Part Size"),
             "border_padding": t("Border Padding"),
+            "head_type": t("Tool head type"),
```
tadeubas, line 224: Seeing other terms used, I would simplify this one to "Head Type" or "Tool Type" what do you think @odudex and @qlrd ?

tadeubas:
Just this: https://github.com/selfcustody/krux/pull/593#discussion_r2105350293

qlrd:
> Just this: [#593 (comment)](https://github.com/selfcustody/krux/pull/593#discussion_r2105350293)

fixed to `Head type`.

tadeubas:
I think it just need to be tested on device before merging develop... doesn't need to have a cnc laser, but to see if it is "working" as expected

odudex:
Please update Changelog. After this I believe it can be merged. Thanks!

PR #600, opened 2025-05-20
Coverage 100% src/krux/buttons.py

qlrd:
### What is this PR for?

* add test for __handler__ method;
* add test for generic button initialization;
* add test for release button events.

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other: coverage `src/krux/buttons.py`.

### DIFF
--- a/tests/test_buttons.py
+++ b/tests/test_buttons.py
@@ -1,3 +1,47 @@
+import pytest
+
+
+@pytest.fixture
+def mock_tactile_buttons(mocker):
+    from krux.buttons import fm, buttons_control
+
+    mocker.patch.object(fm, "register")
+    mock_gpio = mocker.MagicMock()
+    mocker.patch("Maix.GPIO", return_value=mock_gpio)
+
+    def fake_init_enter(pin):
+        buttons_control.enter = mock_gpio
+
+    def fake_init_page(pin):
+        buttons_control.page = mock_gpio
+
+    def fake_init_page_prev(pin):
+        buttons_control.page_prev = mock_gpio
+
+    mocker.patch.object(buttons_control, "init_enter", fake_init_enter)
+    mocker.patch.object(buttons_control, "init_page", fake_init_page)
+    mocker.patch.object(buttons_control, "init_page_prev", fake_init_page_prev)
+
+    buttons_control.enter_event_flag = False
+    buttons_control.page_event_flag = False
+    buttons_control.page_prev_event_flag = False
+
+    mocker.spy(buttons_control, "event_handler")
+
+    return buttons_control
+
+
+def test_handler(mocker, m5stickv, mock_tactile_buttons):
+    from krux.buttons import __handler__
+
+    calls = []
+    for pin in [10, 11, 12]:
+        __handler__(pin)
+        calls.append(mocker.call(pin))
+
+    mock_tactile_buttons.event_handler.assert_has_calls(calls)
+
+
 def test_tactile_buttons_initialization(mocker, m5stickv):
     from krux.buttons import buttons_control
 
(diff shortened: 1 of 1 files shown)


tests/test_buttons.py
```diff
@@ -1,3 +1,31 @@
+def test_handle_tactile_buttons(mocker, m5stickv):
+    from krux.buttons import __handler__, buttons_control
+
+    # Test a simple handler function
+    # Not related to an simbolic button press
+    # just to test the handler. If the value in the pin
+    # is same from `buttons_control.enter` then the event flag
+    # is set to True
+    __handler__(None)
```
odudex, line 9: Although this test is running the code, it's not testing it...
To properly test it, real numbers must be assigned to `enter`, `page` and `page_prev`, through `init_<button>` function, then `__handler__` must be called, with the respective button IO number, so the respective `<button>_event_flag` can be evaluated.

tests/test_buttons.py
```diff
@@ -1,3 +1,31 @@
...
+    from krux.buttons import Button, buttons_control
+
+    # Simulate button pin
+    mocked_pin = 8
+    button_generic = Button()
+
+    # Overwrite the pin's MagicMock with the mocked pin number
+    buttons_control.enter = mocked_pin
+
+    # Assert if button method are called correctly
```
odudex, line 23: "Assert button methods return default values"
```
assert button_generic.value() == 1
assert button_generic.event() == False
```

tests/test_buttons.py
```diff
@@ -25,6 +53,10 @@ def test_button_enter_event(mocker, m5stickv):
     # Assert that the event flag is reset
     assert not button_enter.event()
 
+    # Simulate button release
```
odudex, line 56: This is not simulating a release, it's simulating the device has no button Enter, thus `value()` returns default (released) value: 1

tests/test_buttons.py
```diff
@@ -39,9 +71,14 @@ def test_button_page_event(mocker, m5stickv):
     # Simulate button press
     buttons_control.event_handler(pin)
     assert button_page.event()
+
     # Assert that the event flag is reset
     assert not button_page.event()
 
+    # Simulate button release
+    buttons_control.page = None
```
odudex, line 79: again, simulating the device has no PAGE button.

tests/test_buttons.py
```diff
@@ -56,5 +93,10 @@ def test_button_page_prev_event(mocker, m5stickv):
     # Simulate button press
     buttons_control.event_handler(pin)
     assert button_page_prev.event()
+
     # Assert that the event flag is reset
     assert not button_page_prev.event()
+
+    # Simulate button release
```
odudex, line 100: again, simulating no PAGE_PREV pin instance on device

tests/test_buttons.py
tadeubas: This is worth pointing out. I talked to @qlrd about this the other day. We need tests that actually test the code, not just cover it! We already have good coverage, but bad tests.

odudex:
In sum, to properly mock buttons release, the `GPIO` class should be mocked, when a button is assigned with `None`, this means the device does not have that button and `buttons_control` will always return value() as `RELEASED` (1) and event() as `False`. If you want to test the case were buttons are not assigned to any IO (no button X in this device), then keep the tests and only change the comments.

tests/test_buttons.py
qlrd: While I agree it could use more real-world scenarios, I just want to clarify that it was intended as a unit test. IMHO, the tests that cover actual functionality (as mentioned) can be separated from the unit tests — which was the focus of this PR. Does that make sense?

tadeubas: It's just that your test covered the line, but didn't actually test the code, so this PR leaves us with a false sense of a quality change that, in my opinion, didn't change the quality of our tests.

odudex:
Unit tests are welcome, but there are some issues:

1. On `test_handle_tactile_buttons`, `__handler__(None)` will never happen in device. `__handler__()` is a function called by MCU's interrupts manager, and it will always have a pin assigned instead of None.
Ideally, calling `__handler__()` with `None` should NOT trigger any event.
```
assert buttons_control.enter_event_flag==False
assert buttons_control.page_event_flag==False
...
```
 But as this won't happen on device, we don't need to add code for this case.

2. Comments which can be misleading
```
# Simulate button release
    buttons_control.page = None
```
These comments are incoherent with the code. `buttons_control.page = None` means there's no `page` button object instantiated, not that the button is released. A new comment will fix it, like "# Simulate device does not have button X". The condition being tested is ok and can remain.
Maybe you used "release" term with a programing meaning, but as we are also talking a buttons, due to its possible states (pressed and released), this can be confusing.

tests/test_buttons.py
qlrd: Sure, i will try to fix that

qlrd: I think now i could apply what you guys said

tadeubas:
ACK

tests/test_buttons.py
```diff
@@ -9,52 +53,93 @@ def test_tactile_buttons_initialization(mocker, m5stickv):
...
-    # Simulate button page pin
-    mocked_pin = 11
-    pin = mocked_pin
-    button_page = ButtonPage(pin)
-    # Overwrite the pin's MagicMock with the mocked pin number
-    buttons_control.page = mocked_pin
+    # Simulate device has no button enter
+    mocker.patch.object(mock_tactile_buttons, "init_enter", lambda pin: None)
+    mock_tactile_buttons.enter = None
+    button_enter = ButtonEnter(None)
```
odudex, line 73: This button is being created alone, not as part of `mock_tactile_buttons`(which the goal is to mock `buttons_control`, right?). In addition, we don't create a ButtonX instance with `None` as argument.
To test the case where there's no button ENTER (to be coherent with test title), you can use `mock_tactile_buttons.enter = None`, as you did, but `ButtonEnter` is never instantiated. It seems what you are doing here is to test button's default values again.

tests/test_buttons.py
```diff
@@ -9,52 +53,93 @@ def test_tactile_buttons_initialization(mocker, m5stickv):
...
+
+    button = ButtonEnter(mock_tactile_buttons.enter)
+
+    # Simulate button pressed
+    mock_tactile_buttons.event_handler(mock_tactile_buttons.enter)
+
+    # if really pressed, should be True
+    assert button.event()
+
+    # if released, should be False (reseted the event flag)
```
odudex, line 114: Event flag will be reset to `False` independently of the button state (pressed/released)
The flag is set to `True` by an event (interruption), that occurs when respective button's IO change from logic 1 to logic 0, and is se to `False` (cleared) whenever the `event()` function is called

odudex: Test is good, just comments can be improved for better understanding of interrupt based buttons

tests/test_buttons.py
```diff
@@ -9,52 +53,110 @@ def test_tactile_buttons_initialization(mocker, m5stickv):
...
+def test_button_enter(mocker, m5stickv, mock_tactile_buttons):
+    from krux.buttons import ButtonEnter
+
+    mock_tactile_buttons.enter = 10
+    button = ButtonEnter(mock_tactile_buttons.enter)
+
+    # Simulate button pressed
+    mock_tactile_buttons.event_handler(mock_tactile_buttons.enter)
+
+    # when button is pressed (event method is called),
```
odudex, line 114: This comment is wrong, button release has nothing to do with events. Here's a fix:
When button is pressed, an interrupt is generated and `button_event_flag` is set.
When `button.event()` method is called, it will return True if there has been an event,
but before returning, it will clear `button_event_flag`.

Test code is fine

qlrd: I think it's fixed (copied your text to comments).

odudex (approved):
Thank you!

PR #609, opened 2025-06-04
On-Device Tests (accessed from within "Tools" menu)

jdlcdl:
### What is this PR for?
As a prototype/starting-point for on-device testing of functionality that works differently in simulator than on k210.

DeviceTests page stores "results" list and test_suite will fill them if they don't exist, else will just display the menu
of unit-test results so that user may run individual tests to see more complete results.   User backs out of the test-suite results menu as well as the top level menu which appears when leaving About page.

The last 8 tests are "pseudo tests" (to be removed), useless except for showing that a test_function is meant to return something boolean True to signify success and something boolean False (or raise an exception) to signify failure.  These will all be removed and replaced with minimal/useful tests.

### Changes made to:
- [x] Code
- [X] Tests
- [X] Docs
- [ ] CHANGELOG

- [x] Yes  amigo and yahboom_devkit

- [ ] Bug fix
- [x] New feature
- [X] Docs update
- [ ] Other


tadeubas:
I am against doing things without properly informing the user about it. I would create an entry in **tools** or in **settings** (as SS does) to run device tests. Later we could add test of buttons and touch area too

odudex:
Thank you!

PR #611, opened 2025-06-06
Expand keypad touch area

bitkit0:
### What is this PR for?
Fix the issue where touches on the rightmost column of the keypad did not register on the WonderMV device.

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/pages/keypads.py
+++ b/src/krux/pages/keypads.py
@@ -63,8 +63,9 @@ class KeypadLayout:
             for y in range(self.height + 1)
         ]
         self.x_keypad_map = [
-            x * key_h_spacing + MINIMAL_PADDING for x in range(self.width + 1)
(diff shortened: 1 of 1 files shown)


jdlcdl:
I've tried both before and after applying this pr on both amigo and the unsupported yahboom_devkit.

On both devices, I would grade this as an improvement because the touch screen reacts to touches accurately right up to the right-side margin.  I'd add that we might be able to do the same to the left side as well... as I note that I must get the touch much closer to the letter for a keypad reaction.

odudex:
Thank you @bitkit0! I'll merged it, and as @jdlcdl suggested, will also expand the touch area to the left edge too.

PR #618, opened 2025-06-23
Cover tests for `src/pages/device_tests.py`

qlrd:
### What is this PR for?
Follow-up of #609

* removed some unused imports;
* created some fixtures for better reading;
* added assertions of not called functions when skipping on device tests;

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

### Did you build the code and tested on device?
- [ ] Yes

### What is the purpose of this pull request?
- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other: coverage.

PR #621, opened 2025-06-28
Coverage followup for device_tests

qlrd:
### What is this PR for?

Followup of #618

* added warning message about possible tampering when a test fail;
* adjusted to allow pass interactive argument.
* cover situations where `deflate_compression` and
`deflate_decompression` could fail (injection of malicious code);
* added tests for possible injecion of expected codes;
* added cover case for interactive=True.

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other: tests

### DIFF
--- a/src/krux/pages/device_tests.py
+++ b/src/krux/pages/device_tests.py
@@ -53,7 +53,7 @@ class DeviceTests(Page):
         utils.print_standard_qr(title, title=title, check_printer=False)
         return MENU_CONTINUE
 
-    def test_suite(self):
+    def test_suite(self, interactive=False):
         """run each on-device tests in all_tests, report summary and details"""
         all_tests = [
             self.hw_acc_hashing,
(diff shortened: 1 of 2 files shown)


tadeubas:
I think what Odudex is saying is that device tests are not like the Python unit tests in folder `tests`. Device tests are only to ensure that HW is working as expected, and as they will ship with firmware, any increase in size (like longer strings) will increase the firmware file size, so it is better to summarize things. The Python unit tests can be verbose, can have multiple lines and various checks without restrictions.

qlrd:
So i remove the possibility to configure the interactive test as well the messages (basically the code in `src/krux/pages/device_tests.py` or just the `AssertionError` messages?

odudex:
Thank you!

PR #625, opened 2025-06-30
Add test for ESC key handling in `tests/pages/test_tc_code_verification.py`

0xcindyv:
### What is this PR for?  

Improve test coverage for `tc_code_verification.py` to 100% by adding missing ESC key handling test case.

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other : Add test for ESC key handling in `/tests/pages/test_tc_code_verification.py`

### DIFF
--- a/tests/pages/test_tc_code_verification.py
+++ b/tests/pages/test_tc_code_verification.py
@@ -73,3 +73,26 @@ def test_tc_code_verification(amigo, mocker):
             keypad_label,
             [NUM_SPECIAL_1, LETTERS, UPPERCASE_LETTERS, NUM_SPECIAL_2],
         )
+
+
+def test_tc_code_verification_esc_key(amigo, mocker):
+    from krux.pages import ESC_KEY
+    from krux.pages.tc_code_verification import TCCodeVerification
+    from krux.input import BUTTON_ENTER, BUTTON_PAGE_PREV
+
+    # Simulate user pressing ESC key
+    BTN_SEQUENCE = [
+        BUTTON_PAGE_PREV,  # Navigate to ESC key
+        BUTTON_PAGE_PREV,  # Navigate to ESC key
+        BUTTON_ENTER,      # Press ESC key
+        BUTTON_ENTER,      # Confirm "Yes" in prompt
+    ]
+
+    ctx = create_ctx(mocker, BTN_SEQUENCE)
+    tc_verifier = TCCodeVerification(ctx)
+    
+    # Test full user interaction flow
+    result = tc_verifier.capture()
+    
+    assert result == False
+    assert ctx.input.wait_for_button.call_count == len(BTN_SEQUENCE)


tests/pages/test_tc_code_verification.py
```diff
@@ -73,3 +73,26 @@ def test_tc_code_verification(amigo, mocker):
             keypad_label,
             [NUM_SPECIAL_1, LETTERS, UPPERCASE_LETTERS, NUM_SPECIAL_2],
         )
+
+
+def test_tc_code_verification_esc_key(amigo, mocker):
+    from krux.pages import ESC_KEY
```
qlrd: the constant `ESC_KEY` is not in use.

tests/pages/test_tc_code_verification.py
```diff
@@ -73,3 +73,26 @@ def test_tc_code_verification(amigo, mocker):
...
+        BUTTON_ENTER,      # Confirm "Yes" in prompt
+    ]
+
+    ctx = create_ctx(mocker, BTN_SEQUENCE)
+    tc_verifier = TCCodeVerification(ctx)
+    
+    # Test full user interaction flow
+    result = tc_verifier.capture()
+    
+    assert result == False
```
qlrd: No need to `assert result == False`, `assert not result` is better.

tests/pages/test_tc_code_verification.py
```diff
@@ -73,3 +73,26 @@ def test_tc_code_verification(amigo, mocker):
...
+        BUTTON_ENTER,      # Press ESC key
+        BUTTON_ENTER,      # Confirm "Yes" in prompt
+    ]
+
+    ctx = create_ctx(mocker, BTN_SEQUENCE)
+    tc_verifier = TCCodeVerification(ctx)
+    
+    # Test full user interaction flow
+    result = tc_verifier.capture()
+
```
qlrd: There's a blank space in the new line.

tests/pages/test_tc_code_verification.py
```diff
@@ -73,3 +73,26 @@ def test_tc_code_verification(amigo, mocker):
...
+    BTN_SEQUENCE = [
+        BUTTON_PAGE_PREV,  # Navigate to ESC key
+        BUTTON_PAGE_PREV,  # Navigate to ESC key
+        BUTTON_ENTER,      # Press ESC key
+        BUTTON_ENTER,      # Confirm "Yes" in prompt
+    ]
+
+    ctx = create_ctx(mocker, BTN_SEQUENCE)
+    tc_verifier = TCCodeVerification(ctx)
+
```
qlrd: There's a blank space in the new line.

qlrd:
Additionally i think that would be neat to test a situation where the user press esc and when asked to "Are you sure", moves to "No" and then esc again and then really exit.

qlrd:
We need to clean the commits to better review

qlrd:
ACK f4b044b

odudex:
Thank you, @bc1cindy!

One small detail I think could be improved is the specificity of the assertions.

For example, using `assert not variable_x `will pass if `variable_x` is `False`, `None`, `0`, or an empty list, so the test may unintentionally pass when it shouldn't. Similarly, assert `variable_x` is truthy for values like `True`, any non-empty string, number, or list, which might also lead to misleading results.

While direct assert statements look clean (and I used to prefer them), I've been reconsidering this lately. In the context of tests, I think it’s better to be more explicit about the expected values to avoid unintended behavior and make the tests more robust.

PR #626, opened 2025-07-01
Improve coverage `/tests/pages/test_fill_flash.py`

bc1cindy:
### What is this PR for?

Test amigo device flash with camera entropy data simulating user interaction flows

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other - Improve coverage `/tests/pages/test_fill_flash.py`

### DIFF
--- a/tests/pages/test_fill_flash.py
+++ b/tests/pages/test_fill_flash.py
@@ -1,3 +1,4 @@
+import pytest
 from . import create_ctx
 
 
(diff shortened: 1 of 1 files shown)


tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
 from . import create_ctx
 
 
-def test_fill_flash(amigo, mocker):
```
qlrd: What is the rationale to remove this test as it was?

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
-        FillFlash,
-        BLOCK_SIZE,
-        IMAGE_BYTES_SIZE,
-        TOTAL_BLOCKS,
-    )
-    from krux.input import BUTTON_ENTER
-    from krux.firmware import FLASH_SIZE
-    import flash
+def fill_flash_with_camera_entropy_functional(ctx):
+    """Functional wrapper for FillFlash.fill_flash_with_camera_entropy following the pattern"""
```
qlrd: I like docstrings in tests, but the standard until now is add verbose comments explaining the rationale

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+    from krux.pages import MENU_CONTINUE
+    from krux.input import BUTTON_ENTER, BUTTON_PAGE, BUTTON_PAGE_PREV
+    from krux.pages.fill_flash import IMAGE_BYTES_SIZE, BLOCK_SIZE, MAX_CAPTURE_PERIOD
+    import flash
+    import time
 
-    assert ctx.input.wait_for_button.call_count == len(BTN_SEQUENCE)
-    # Check that the flash was written to on half of the blocks
-    assert flash.write.call_count == TOTAL_BLOCKS // 2 - 1
+    def create_entropy_mock(entropy_type="good", gradual=False):
```
qlrd: This isn't a idiomatic way to write tests (it appears to be a js way to write). IMO over complicates and deviates from the principle of having focused tests. These scenarios could be better achieved with fixtures.

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+        """Factory for creating different entropy mock scenarios"""
+        def mock_snapshot():
+            counter = [0]
 
+            def next_image():
+                counter[0] += 1
+                img = mocker.MagicMock()
+                
+                if entropy_type == "poor":
```
qlrd: IMO, these cases could be addressed to different tests given some already defined constants for camera entropy. And inside, test different cases.

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+                    img.to_bytes.return_value = bytes([5] * IMAGE_BYTES_SIZE)
+                    stats = mocker.MagicMock()
+                    stats.l_stdev.return_value = 2
+                    stats.a_stdev.return_value = 1
+                    stats.b_stdev.return_value = 3
+                elif entropy_type == "gradual" and gradual:
```
qlrd: We don't have a `"gradual"` entropy type.

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+                
+                img.get_statistics.return_value = stats
+                return img
+            return next_image
+        return mock_snapshot
 
-def test_capture_image_with_sufficient_entropy(amigo, mocker):
-    from krux.pages.fill_flash import FillFlash
-    from krux.pages.capture_entropy import CameraEntropy, POOR_VARIANCE_TH
+    def setup_flash_mock(scenario="normal"):
```
qlrd: this could be better achieved throug mocker fixtures.

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
-    entropy_measurement = CameraEntropy(ctx)
-    mocker.spy(entropy_measurement, "entropy_measurement_update")
-    entropy_measurement.rms_value = mocker.PropertyMock(side_effect=STDEV_INDEX)
-    fill_flash.capture_image_with_sufficient_entropy(entropy_measurement)
-
-    # Check that the entropy_measurement_update was called 4 times
-    # 3 times for poor entropy and 1 time for good entropy
-    assert entropy_measurement.entropy_measurement_update.call_count == 4
+    # Test cases with behavior descriptions
+    test_cases = [
```
qlrd: Our standard, until now is an array of tuples with the expected values. The `name` and `description` could be replaced by comments. Also is better to used already defined constants for entropy, not hard-coded ones.

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+        {
+            "name": "complete_successful_flow",
+            "sequence": [BUTTON_PAGE, BUTTON_PAGE_PREV, BUTTON_ENTER],
+            "entropy": "mixed",
```
qlrd: We don't have a `"mixed"` entropy type

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+            "flash": "mixed",
+            "expected": MENU_CONTINUE,
+            "description": "Quick check of 'No' option, back to 'Yes' (confident user), immediate confirmation"
+        },
+        {
+            "name": "gradual_entropy_improvement",
+            "sequence": [BUTTON_PAGE, BUTTON_PAGE, BUTTON_PAGE_PREV, BUTTON_PAGE_PREV, BUTTON_ENTER],
+            "entropy": "gradual",
```
qlrd: We don't have `"gradual"` entropy

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
-    fill_flash.capture_image_with_sufficient_entropy(entropy_measurement)
-
-    # Check that the entropy_measurement_update was called 4 times
-    # 3 times for poor entropy and 1 time for good entropy
-    assert entropy_measurement.entropy_measurement_update.call_count == 4
+    # Test cases with behavior descriptions
+    test_cases = [
+        {
+            "name": "complete_successful_flow",
+            "sequence": [BUTTON_PAGE, BUTTON_PAGE_PREV, BUTTON_ENTER],
```
qlrd: is good in the sequence to have comments about the UI navigation

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+            "entropy": "mixed",
+            "flash": "mixed",
```
qlrd: IMO field is more a case that could be addressed by fixtures

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+            "entropy": "progressive",
+            "flash": "large",
+            "expected": MENU_CONTINUE,
+            "description": "Goes to 'No' (anxious), analysis paralysis, multiple back-and-forth, finally commits"
+        }
+    ]
+
+    # Execute test cases
+    for case in test_cases:
+        print(f"Testing: {case['name']} - {case['description']}")
```
qlrd: This is better when debbugging, but not when all works

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+        
+        # Setup entropy mock
+        if case["entropy"]:
+            gradual = case["entropy"] == "gradual"
+            mocker.patch("sensor.snapshot", side_effect=create_entropy_mock(case["entropy"], gradual)())
```
qlrd: When mocking is better to use fixtures

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+        # Setup time mock for timeout tests
+        if case.get("timeout"):
+            setup_time_mock()
+        
+        # Create context and execute test
+        ctx = create_ctx(mocker, case["sequence"])
+        
+        # Execute and verify
+        if case["expected"] == ValueError:
+            try:
```
qlrd: in assertions, do not use `try/except`, because this not assert if something was raised as expected. The correct way is use the `assertRaises`.

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+                fill_flash_with_camera_entropy_functional(ctx)
+                assert False, f"Case {case['name']}: Should have raised ValueError for timeout"
```
qlrd: No need to add messages in the assertions

tests/pages/test_fill_flash.py
```diff
@@ -1,57 +1,221 @@
...
+        else:
+            result = fill_flash_with_camera_entropy_functional(ctx)
+            assert ctx.input.wait_for_button.call_count == len(case["sequence"])
+            assert result == case["expected"]
+            
+            # Special assertion for flash error case
+            if case["flash"] == "error" and flash_counter:
+                assert flash_counter[0] > 0, "Flash should have been accessed"
+        
+        # Reset mocks for next iteration
```
qlrd: If you use `mocker.reset` maybe is better to separate tests and do some fixtures.

qlrd (changes requested):
The original tests were already covering basic functionality, no need to remove or change. 

I think, in general, the used wrappers aren't idiomatic and over complicated the test. 

Also, it was missing checking whether some internal methods of the `CameraEntropy` class were called appropriately.

Use `mocker.resetall` as last resource, is better and reasonable to isolate them with fixtures.

tests/pages/test_fill_flash.py
```diff
@@ -55,3 +56,304 @@ def test_capture_image_with_sufficient_entropy(amigo, mocker):
     # Check that the entropy_measurement_update was called 4 times
     # 3 times for poor entropy and 1 time for good entropy
     assert entropy_measurement.entropy_measurement_update.call_count == 4
+
+
+@pytest.fixture
+def mock_coverage_scenarios(mocker):
```
qlrd: These could be three different fixtures

tests/pages/test_fill_flash.py
```diff
@@ -55,3 +56,304 @@ def test_capture_image_with_sufficient_entropy(amigo, mocker):
...
+    # Comprehensive fixture providing mocks for various fill flash test scenarios
+    # Covers entropy validation, flash operations, timeout handling, and user interactions
+    from krux.pages.capture_entropy import INSUFFICIENT_VARIANCE_TH
+    from krux.firmware import FLASH_SIZE
+
+    def _mock_flash_read(address, size):
+        # Simulates flash memory with second half empty (0xff) and first half occupied (0x00)
```
qlrd: I think we need @jdlcdl to judge that

tests/pages/test_fill_flash.py
```diff
@@ -55,3 +56,304 @@ def test_capture_image_with_sufficient_entropy(amigo, mocker):
...
+        # This creates realistic conditions where only empty blocks need to be filled
+        if address > FLASH_SIZE // 2:
+            return b"\xff" * size
+        else:
+            return b"\x00" * size
+
+    return {
+        # Mock insufficient entropy scenario: RMS value below minimum threshold
+        # Used to test entropy validation failure paths in camera capture
+        "insufficient_entropy": mocker.PropertyMock(
```
qlrd: i advocate to patch instead use Property mock. Mocking properties instead patch or spy them could lead to false positive assertions

tests/pages/test_fill_flash.py
```diff
@@ -55,3 +56,304 @@ def test_capture_image_with_sufficient_entropy(amigo, mocker):
...
+            ctx.input.touch, "current_index", return_value=0
+        ),
+    }
+
+
+def test_fill_flash_coverage_scenarios(amigo, mocker, mock_coverage_scenarios):
+    from krux.pages.fill_flash import FillFlash, MENU_CONTINUE, IMAGE_BYTES_SIZE
+    from krux.input import BUTTON_ENTER, BUTTON_TOUCH, BUTTON_PAGE, BUTTON_PAGE_PREV
+
+    cases = [
```
qlrd, line 225: This isn't in the standard used by our tests (array of tuples instead dictionaries)

tests/pages/test_fill_flash.py
```diff
@@ -55,3 +56,304 @@ def test_capture_image_with_sufficient_entropy(amigo, mocker):
...
+            "expected_result": MENU_CONTINUE,
+            "verify_calls": ["wait_for_button"],
+        },
+    ]
+
+    for case in cases:
+        # Apply capture method mocks first to ensure they take precedence
+        if case.get("mock_capture_method", False):
+            if "capture_side_effect" in case:
+                mocker.patch(
```
qlrd: These patches inside loops could be used in custom fixtures

tests/pages/test_fill_flash.py
```diff
@@ -55,3 +56,304 @@ def test_capture_image_with_sufficient_entropy(amigo, mocker):
...
+                    "krux.pages.fill_flash.FillFlash.capture_image_with_sufficient_entropy",
+                    return_value=case["capture_return_value"],
+                )
+
+        ctx = create_ctx(mocker, case["btn_sequence"])
+        fill_flash = FillFlash(ctx)
+
+        # Configure touch interface if test case uses touch
+        if "touch_return" in case:
```
qlrd: cases are used with tuple elements

qlrd (changes requested):
the tests have a large number of scenarios into single function, inside loops with patching inside each test; 
this pattern can make it more challenging to maintain.

prefer to use not-so-complex fixtures to set up different test conditions and keep individual test cases focused. Try to find out which is the functionality of what you want to cover, so you be able to refactor these aligned with our usual testing practices.

qlrd:
I think is better to keep it Draft while we review that

tests/pages/test_fill_flash.py
```diff
@@ -49,9 +50,390 @@ def test_capture_image_with_sufficient_entropy(amigo, mocker):
...
+    from krux.pages.fill_flash import FillFlash, MENU_CONTINUE
+    from krux.input import BUTTON_TOUCH, BUTTON_PAGE, BUTTON_PAGE_PREV
+
+    btn_sequence = [
+        BUTTON_PAGE,  # Navigate options to understand choices
+        BUTTON_PAGE_PREV,  # Navigate back to explore both options
+        BUTTON_TOUCH,  # Finally use touch interface to decline
+    ]
+
+    ctx = create_ctx(mocker, btn_sequence)
```
odudex: `create_ctx`can receive `touch_seq` argument, there's no need to create a dedicated fixture.

PR #629, opened 2025-07-04
Highlight addresses for easy verification

tadeubas:
### What is this PR for?
* Addresses are highlighted in alternating groups of 4 — the first is highlighted, the next is not, and so on. Enhances this [already merged PR](https://github.com/selfcustody/krux/pull/541)
* Change affect scan/list address and sign PSBT

### Changes made to:
- [x] Code
- [x] Tests
- [x] Docs
- [x] CHANGELOG

- [x] Yes, yahboom

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other


src/krux/pages/home_pages/home.py
```diff
@@ -284,34 +333,12 @@ def sign_psbt(self):
...
-        from ...psbt import PSBTSigner
-
-        signer = PSBTSigner(self.ctx.wallet, data, qr_format, psbt_filename)
-
-        del data
-        gc.collect()
+        return True
 
+    def _post_load_psbt_warn(self, signer):
+        """Warns cases of incorrect / missing PSBT info"""
```
odudex, line 341: There are 3 checks in this code, now isolated in a  function, but now if one is triggered, subsequent checks are bypassed, right?
Please describe these and other changes which do not seem to correlate with the title of the PR.

tadeubas: These changes were necessary because pylint was complaining about the number of lines / variables / branches in this function, so I had to isolate some parts in order to pass pylint.

The code is to work as it was before, we show warnings and then wait for user input, if user doesn't agree to proceed, it leaves the signing PSBT process (again, as it was before). If user agrees to proceed, it will check the next warning and so on. You can see tests already covered these situations and there wasn't a change in those.

tadeubas: Note that the `POST_load_psbt_warn` the check remained the same, it was:
```
                return MENU_CONTINUE
```                
And changed to:
```
                return False
```

The `PRE_load_psbt_warn` is different, it has only one check, so there is no problem to use return on the `self.prompt` instead of another if:
```
return self.prompt(t("Proceed?"), BOTTOM_PROMPT_LINE)
```

odudex: You're right! I glitched here and misinterpreted the code!

odudex:
Thank you!

PR #642, opened 2025-07-16
Docs: update mobile android app links

tadeubas:
### What is this PR for?
- Update links for Krux Mobile Android app in docs.
- Fixed Tinyseed name (according to their site https://tinyseed.io/product/tinyseed/).
- Tools added Device tests options img. 
- Also added "The Bitcoin Hardware Store" video tutorial.
- Small text changes, added links, sequences and img updates

### Changes made to:
- [ ] Code
- [ ] Tests
- [x] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [x] Docs update
- [ ] Other


docs/faq.en.md
```diff
@@ -8,17 +8,17 @@ Krux does not include hardware secure elements. The security of stored data reli
 **Note**: Due to the inherent fragility of electronic components, never use your Krux device or SD card encrypted storage as your sole backup method. Always maintain a physical backup for added security.
 
 ## What is Beta version?
-The Beta version includes the latest and most experimental features, which we occasionally share on our social media. These can be found exclusively in the [test (beta) repository](https://github.com/odudex/krux_binaries/). Use and flash the beta firmware if you are curious about new features or want to participate in the development process by hunting bugs, providing feedback, and sharing ideas in our Telegram groups or other social media platforms.
+The Beta version includes the latest and most experimental features, which we occasionally share on our social media. These can be found exclusively in the [test (beta) repository](https://github.com/selfcustody/KruxMobileApp). Use and flash the beta firmware if you are curious about new features or want to participate in the development process by hunting bugs, providing feedback, and sharing ideas in our Telegram groups or other social media platforms.
```
odudex: Betas continue to be at my repository

tadeubas: sorry, my bad... fixed!

docs/img/maixpy_amigo/device-tests-test-suite-300.en.png
odudex, line 1: Will we let these tests for production? If we do, we should remove the ones that fail "on purpose" to avoid confusing users.

tadeubas: Oh now I understood, so I have commented them on code

src/krux/pages/device_tests.py
```diff
@@ -53,21 +53,22 @@ def print_test(self):
         utils.print_standard_qr(title, title=title, check_printer=False)
         return MENU_CONTINUE
 
+    # pylint: disable=unused-argument
```
odudex: @jdlcdl @qlrd What do you think about removing tests(not only commenting), arguments, pylint disables, etc. of tests, functions and arguments which won't be used. Can we need them later? I believe these code snippets have already fulfilled their purpose.

jdlcdl: I would agree that removing them is best, and also the pylint disabler comments that won't be needed once deleted.

Their original purpose was to prove that they were working and to relay intention to other devs.   As tests below were adjusted, 100% success should be the expectation.

added: For features that we become more comfortable-with over time (or as we learn to better mimick them via simulator), I could see them being removed eventually too.  A test that might be useful (but arguably might be security-theater) could be a "deterministic_sigs" test that where the on-device input/expected-output gets altered just before release so that we're verifying that a "fresh" message always results in the same bit-for-bit signature on both simulator and devices.

tadeubas: Removed the commented code

tadeubas: As I've changed nothing, just commented the code (that has now been removed) in order to generate the right screen to DOCS, can we create another PR to address `device_tests` as this one is focused only in DOCS?

tadeubas: removed changes from files not in docs

PR #645, opened 2025-07-21
Improved UX when reviewing and signing PSBTs

tadeubas:
### What is this PR for?
This PR improves the user experience when reviewing and signing PSBTs (especially when the user prefers to use QR codes).

Previously, users who loaded a PSBT via camera and reviewed its details had no way to revisit that information without reloading the PSBT. Additionally, when displaying the signed QR code, an accidental tap (especially on small touch devices) could dismiss the screen prematurely, leading to a frustrating user experience.

To address these issues, this PR introduces:
- A **PSBT Review button** in the sign menu, allowing users to re-check PSBT details without needing to reload it.
- A **confirmation prompt** before exiting the signed QR code screen, reducing the chance of accidental dismissals.

### Changes made to:
- [x] Code
- [x] Tests
- [x] Docs
- [x] CHANGELOG

- [x] Yes, yahboom

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other


src/krux/pages/home_pages/home.py
```diff
@@ -227,50 +227,59 @@ def load_psbt(self):
...
-                    ),
-                ],
-                back_status=lambda: None,
-            )
-            index, _ = sign_menu.run_loop()
-            return index
+        submenu = Menu(
+            self.ctx,
+            [
+                (t("Review"), lambda: None),
```
odudex: IMO "Review" alone, at this stage, is confusing, as user just did that. It could be "Review Again"

tadeubas: Nice suggestion, thx! Changed for "Review Again" :wink:

PR #647, opened 2025-07-24
Cover `kef.py` test

bc1cindy:
### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other


tests/test_kef.py
```diff
@@ -1602,3 +1602,66 @@ def test_brute_force_compression_checks(m5stickv):
...
+        with patch.object(cipher, "_authenticate", side_effect=Exception("Test error")):
+            result = cipher.decrypt(test_payload, version)
+            assert result is None
+
+    # Test skip disabled versions
+    original_versions = {}
+    for k, v in kef.VERSIONS.items():
+        original_versions[k] = v.copy() if isinstance(v, dict) else v
+
+    try:
```
qlrd: if a test uses `try/except` clauses, it means that something isn't being tested well. It isn't about just cover lines. In these cases we should expect some exceptions to be raised (or not be raised, depending on the context)

tests/test_kef.py
```diff
@@ -1602,3 +1602,66 @@ def test_brute_force_compression_checks(m5stickv):
         )
     )
     # assert 0
+
+
+def test_kef_coverage_edge_cases(m5stickv):
+    """Test edge cases for decrypt exception and disabled versions"""
```
qlrd: tests with exceptions must test if the mentioned exceptions are raised. In a test we don't have a choice to ignore it if we know that something will be raised.

qlrd (commented):
IMHO those things can be covered in #630.

qlrd:
> > IMHO those things can be covered in #630.
> 
> #630 isn't my PR, so I can't add changes to it. How should I proceed? @qlrd

I think we should wait for a @jdlcdl review

tests/test_kef.py
```diff
@@ -1602,3 +1602,66 @@ def test_brute_force_compression_checks(m5stickv):
         )
     )
     # assert 0
+
+
+@pytest.fixture
```
qlrd: The project have a standard to use fixtures as a pattern to setup potentialy repetitive mocks or mocks that are complex and could make the test code a mess to be read. There's not any mock here. Also, any fixture should be placed before the tests as mean for organization.

tests/test_kef.py
```diff
@@ -1602,3 +1602,66 @@ def test_brute_force_compression_checks(m5stickv):
...
+
+    # Restore original VERSIONS
+    for k, v in original_versions.items():
+        kef.VERSIONS[k] = v
+
+
+def test_kef_coverage_edge_cases(m5stickv, restore_kef_versions):
+    """Test edge cases for decrypt exception and disabled versions"""
+    from krux import kef
+    from unittest.mock import patch
```
qlrd: this kind of import should be avoided in functions and be placed at the top.

qlrd (changes requested):
>Regarding the exception test: we're not testing if an exception is raised, but rather testing that decrypt() returns None when
_authenticate() fails (the exception is caught internally by the code). That's why assert result is None instead of using
pytest.raises.


Check in the `src/krux/kef.py` that a `exception` will be raised when an instance of `AES` -- a sugar syntax for `ucryptolib.aes` -- call for a `verify` in a `MODE_GCM` context.

If this call fails, then a exception will be called (and then the result will be a `None`). 

So, we must ask to ourselves: what conditions lead to a failure when we try to `verify` something that was encrypted (instead just mock the result of `_authenticate`?

If we see in the line 247, a possibility could be when we got nothing from `auth = payload[-v_auth:]`. i.e, **when a payload is an empty list**. 

In this case, i would even propose an "empty list check" to the begining of `_authenticate` instead mock the result.

Another possibility is when the extracted value from payload is invalid and, in truth, this is where we should focus the test. 

What is a proper context that lead to an invalid payload (a valid data with invalid authentication) so we could call `decrypt` and the `_authenticate` method could raise an exception and then return `None`?

bc1cindy:
Thanks for the thorough review @qlrd and @jdlcdl!

Fixed all the points: removed unnecessary fixture, split into two focused tests, used bytes directly with structured payloads (iv + ciphertext + auth), and added all 4 modes with 10K iterations.

On the mock vs real data - I agree real payloads would be ideal, but as @jdlcdl noted this is a hard-to-reach edge case. The mock ensures we test the specific exception handling behavior needed for coverage.

@jdlcdl looking forward to your TODO cleanup PR and happy to merge when ready. 

Thanks for the guidance.
Ready for re-review!

qlrd:
> On the mock vs real data - I agree real payloads would be ideal, but as @jdlcdl noted this is a hard-to-reach edge case. The mock ensures we test the specific exception handling behavior needed for coverage.

When we mock too much, maybe we facing with a possible dead code.

bc1cindy:
> > @jdlcdl looking forward to your TODO cleanup PR and happy to merge when ready.
> 
> It was previously pr 5 in your fork of krux, but I have since closed it because it would be outdated after a `git push --force`... it would have undone the fixes you last made. It is now pr 6 in your fork, consisting of a bit more this time:
> 
> * It has the same completion of TODO tasks that validate aes_object in the _authenticate() method,
> * It also has alterations to your lastest changes in the test_suggest_versions_with_disabled_versions() test because while the fixture was removed, the logic to backup kef.VERSIONS and to restore them in a try/finally block were still there.  I removed the backup of kef.VERSIONS, ran the test outside of a try/finally block, removed the restoration of kef.VERSIONS all because they were necessary.  Additionally, I altered the assert statement which used syntax that includes a failure message -- not because it's un-useful (in fact, it's a nice feature of assert) but more because it's not a pattern currently used in the module or codebase... so it was more about a style choice.
> 
> Please review and if you approve merging, then the changes would appear in this pr.




Thanks @jdlcdl!

Merged and all changes have been integrated into a single commit:

TODO removed and aes_object validation implemented in _authenticate() ; Test cleanup from PR #6 applied ; All tests passing.

Ready for final review.

jdlcdl:
I hereby ACK all of these latest changes.

I will add a few notes for you Cindy:

* I've proposed a change to kef.py and test_kef.py that would become a "git conflict" with some of your changes, so depending on the order that prs are merged, one of us will need to resolve and we'll be here to help, but my pr will still require yours to fill the missing test coverage that this pr achieves.
* While reviewing different versions of this pr, the fact of force-pushing considerable changes makes it difficult for those who have read previously to keep track of what changed, without start from top down again.  Just keep that in mind please, whenever you know that others are pulling your pr and working with similar changes... a force-push will change history whereas just adding more commits on top are easy to keep track of.  A squash at merge time or after all ACKs are in is also possible to keep commits clean.
* I made a mistake yesterday by not including kef.py in pr6 of your branch.  I had done so in pr5, with slightly different validation, but I did notice that you resolved the validation, and either form (yours or mine) would work fine.  Good job on that.

Thanks again for your pr and the timely adjustments.

tadeubas:
> > IMHO those things can be covered in #630.
> 
> #630 isn't my PR, so I can't add changes to it. How should I proceed? @qlrd

Hi @bc1cindy, welcome aboard!
You can contribute to someone else's PR by opening a PR to their repository, targeting the same branch they used for the PR here. Once they accept your changes into their branch, the PR on this repository will automatically update.

bc1cindy:
> > > IMHO those things can be covered in #630.
> > 
> > 
> > #630 isn't my PR, so I can't add changes to it. How should I proceed? @qlrd
> 
> Hi @bc1cindy, welcome aboard! You can contribute to someone else's PR by opening a PR to their repository, targeting the same branch they used for the PR here. Once they accept your changes into their branch, the PR on this repository will automatically update.

Thanks for the guidance! @tadeubas 

As suggested, I've contributed these changes to PR #630 via jdlcdl/krux#7

bc1cindy:
Since is merged with PR #630 do I close this PR? @jdlcdl

odudex:
Thank you!

PR #648, opened 2025-07-25
Add coverage for `power.py` test

bc1cindy:
Cover missing battery calculation for generic devices and USB detection. Using unit tests instead of functional pattern as PowerManager is a utility class without UI - no button sequences or user interaction to test.

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other


tests/test_power.py
```diff
@@ -146,3 +146,47 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+
+            result = manager.battery_charge_remaining()
+            assert result == expected
+
+        elif test_type == "usb":
+            manager.pmu.usb_connected.return_value = value
+
+            result = manager.usb_connected()
+            assert result is expected
+            manager.pmu.usb_connected.assert_called_once()
```
qlrd: you should test different things in different test cases

tests/test_power.py
```diff
@@ -146,3 +146,47 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+        from krux.power import PowerManager
+
+        # Setup device-specific mocking
+        if device_type == "dock":
+            # Mock generic device (not amigo/m5stickv/cube)
+            mock_kboard = mocker.MagicMock()
+            mock_kboard.is_amigo = False
+            mock_kboard.is_m5stickv = False
+            mock_kboard.is_cube = False
+            mocker.patch("krux.power.kboard", mock_kboard)
```
qlrd: you should use the already configured device fixtures and not mock them in the middle of the function.

tests/test_power.py
```diff
@@ -146,3 +146,47 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+
+        elif test_type == "usb":
+            manager.pmu.usb_connected.return_value = value
+
+            result = manager.usb_connected()
+            assert result is expected
+            manager.pmu.usb_connected.assert_called_once()
+
+        # Reset mocks for next iteration
+        mocker.resetall()
```
qlrd: no need to reset mocks if the test is properly configured with the correct fixtures

tests/test_power.py
```diff
@@ -146,3 +146,47 @@ def test_reboot_without_pmu(mocker, m5stickv):
 
     krux.power.machine.reset.assert_called()
     krux.power.sys.exit.assert_called()
+
+
+def test_power_manager_coverage(mocker, dock, m5stickv):
+    # Test cases for battery and USB coverage (lines 68 and 80)
```
qlrd: avoid LLM comments or method suffixes like these

qlrd (changes requested):
I think the assertions are sufficient, but mixed too much the tests and the mocks are improper. Also, there's only a device being tested

tests/test_power.py
```diff
@@ -146,3 +146,60 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+
+    manager = PowerManager()
+
+    if manager.pmu is not None:
+        manager.pmu.get_battery_voltage.return_value = 4000
+        manager.pmu.charging.return_value = True
+
+        charge = manager.battery_charge_remaining()
+
+        assert charge <= 0.9
```
qlrd: instead use `multiple_devices` fixture, is better to test each device that has `pmu` object. Also, is good to `spy` some internal calls and assert that they are called. Made a "PR in your PR"

tests/test_power.py
```diff
@@ -146,3 +146,60 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+    if manager.pmu is not None:
+        manager.pmu.get_battery_voltage.return_value = voltage
+        manager.pmu.charging.return_value = False
+
+        charge = manager.battery_charge_remaining()
+
+        if board.config["type"] in ("amigo", "m5stickv", "cube"):
+            assert 0.0 <= charge <= 1.0
+        else:
+            assert charge == expected
```
qlrd: same, but calling a "newer device"

tests/test_power.py
```diff
@@ -146,3 +146,60 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+
+    if manager.pmu is not None:
+        manager.pmu.usb_connected.return_value = True
+        assert manager.usb_connected() is True
+        manager.pmu.usb_connected.assert_called_once()
+
+        manager.pmu.usb_connected.reset_mock()
+        manager.pmu.usb_connected.return_value = False
+        assert manager.usb_connected() is False
+        manager.pmu.usb_connected.assert_called_once()
```
qlrd: separated it in "pairing tests"

qlrd (changes requested):

src/krux/power.py
```diff
@@ -59,11 +59,25 @@ def has_battery(self):
...
     def battery_charge_remaining(self):
         """Returns the state of charge of the device's battery"""
+        # Initial charge value so we will not have a possible
+        # unbound value when calculating charge
+        charge = 1.0
+
+        # Start to calculate charge
         mv = int(self.pmu.get_battery_voltage())
+
+        # For amigo and m5stickv he had done testing
```
qlrd: some typo, sooowy

```suggestion
        # For amigo and m5stickv we had done testing
```

qlrd (changes requested):

tests/test_power.py
```diff
@@ -146,3 +146,188 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+        assert charge <= case[1]
+
+    manager.pmu.get_battery_voltage.assert_has_calls(get_battery_voltage_calls)
+    manager.pmu.charging.assert_has_calls(charging_calls)
+
+
+# This test is for potentially new devices
+# that are not implemented yet in the firmware
+# and do not have the necessary linear approximation
+# # to calculate the battery charge remaining.
```
qlrd: extra `#` here

```suggestion
# to calculate the battery charge remaining.
```

tests/test_power.py
```diff
@@ -146,3 +146,188 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+# It will use the MAX_BATTERY_MV and MIN_BATTERY_MV
+# # to calculate the charge remaining. and work
```
qlrd: here too

qlrd (changes requested):

src/krux/power.py
```diff
@@ -59,11 +59,25 @@ def has_battery(self):
 
     def battery_charge_remaining(self):
         """Returns the state of charge of the device's battery"""
+        # Initial charge value so we will not have a possible
+        # unbound value when calculating charge
+        charge = 1.0
+
+        # Start to calculate charge
```
odudex: IMO these comments could be removed

bc1cindy: removed @odudex

bc1cindy:
removed @odudex

tests/test_power.py
```diff
@@ -146,3 +146,188 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+    manager.pmu.get_battery_voltage.assert_has_calls(get_battery_voltage_calls)
+    manager.pmu.charging.assert_has_calls(charging_calls)
+
+
+# This test is for potentially new devices
+# that are not implemented yet in the firmware
+# and do not have the necessary linear approximation
+# to calculate the battery charge remaining.
+# It will use the MAX_BATTERY_MV and MIN_BATTERY_MV
+# to calculate the charge remaining. and work
```
qlrd:
```suggestion
# to calculate the charge remaining and work
```

qlrd (changes requested):

tests/test_power.py
```diff
@@ -146,3 +146,188 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+        (3000, 0.0),  # Empty
+    ]
+
+    get_battery_voltage_calls = []
+    charging_calls = []
+    for case in cases:
+        manager.pmu.get_battery_voltage.return_value = case[0]
+        manager.pmu.charging.return_value = False
+
+        manager.battery_charge_remaining()
```
odudex: Sorry I just looked at the main code before, now checking the test:

Why are you calling the function twice?

And below, why `assert charge <= case[1]`? Shouldn't it be an exact value, or a value between an accepted range?

bc1cindy:
Fixed both issues @odudex 

1. Removed duplicate function call - now calling battery_charge_remaining() only once
3. Changed assertions to use exact values with pytest.approx() tolerance instead of <=

Also switched from mocker.spy to mocker.patch.object to properly mock return values.

Thank you.

tests/test_power.py
```diff
@@ -146,3 +146,155 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+
+def test_battery_charge_amigo(mocker, amigo):
+    from krux.power import PowerManager
+
+    manager = PowerManager()
+
+    # (voltage, expected_charge)
+    cases = [
+        (4200, 1.0),
+        (4000, 1.0),
```
qlrd, line 186: IMO, the second argument should be less than 1.0

tests/test_power.py
```diff
@@ -146,3 +146,155 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+        (3600, 0.494),
+        (3000, 0.0),
+    ]
+
+    for voltage, expected in cases:
+        mocker.patch.object(manager.pmu, "get_battery_voltage", return_value=voltage)
+        mocker.patch.object(manager.pmu, "charging", return_value=False)
```
qlrd: What's is the reason to swap the `mock.spy` with `mocker.patch`?

IMO mocking the return calls, in this context, do not really test if the internal functions (`get_battery_voltage` and `charging`) was called correctly. 

This force a result to be a "would like" thing, not what "should be". Maybe isn't better to mock the "battery" itself like before? 

Also, this not checking if these internal methods are being called correctly in their argument calls (`assert_called_once_with`).

tests/test_power.py
```diff
@@ -146,3 +146,155 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+    cases = [
+        (4200, 1.0),
+        (4000, 1.0),
+        (3600, 0.593),
+        (3000, 0.0),
+    ]
+
+    for voltage, expected in cases:
+        mocker.patch.object(manager.pmu, "get_battery_voltage", return_value=voltage)
+        mocker.patch.object(manager.pmu, "charging", return_value=False)
```
qlrd: Same here

tests/test_power.py
```diff
@@ -146,3 +146,155 @@ def test_reboot_without_pmu(mocker, m5stickv):
...
+    cases = [
+        (4200, 1.0),  # Fully charged
+        (4000, 0.833),  # Partially charged
+        (3600, 0.5),  # Half charged
+        (3000, 0.0),  # Empty
+    ]
+
+    for voltage, expected in cases:
+        mocker.patch.object(manager.pmu, "get_battery_voltage", return_value=voltage)
+        mocker.patch.object(manager.pmu, "charging", return_value=False)
```
qlrd: Same here

qlrd (changes requested):

tests/test_power.py
bc1cindy: The formulas were calibrated so that ~3900mV already represents a full battery in practice. Any value above this gets clamped by min(1, charge) to never exceed 100%. The test validates this actual implementation behavior. 

Am I right? @qlrd

bc1cindy:
You're right @qlrd! I've updated the tests to:

1. Use MagicMock directly
2. Add spy tracking, to verify method calls
3. Include assert_called_once(), for proper call verification

New approach:
manager.pmu.get_battery_voltage = mocker.MagicMock(return_value=voltage)
voltage_spy = mocker.spy(manager.pmu, "get_battery_voltage")
# ... test ...
voltage_spy.assert_called_once()

This maintains consistency while properly testing the internal method calls. Thanks for the feedback!

PR #657, opened 2025-08-03
Fixed sign message from SD for binary files

tadeubas:
### What is this PR for?
Basically it enables signing of Krux releases via SD card while maintaining other functionalities like signing messages at addresses on Sparrow.

- Resolves display issue on M5StickV where part of the HEX public key prompt was being cut off during SD card save
- Improves UX by moving "Save to SD card?" and "Overwrite?" prompts to the bottom of the screen for better readability
- Adds a line break before displaying the filename in the "Saved to SD card:" message to enhance clarity
- Fix `tests/pages/home_pages/test_sign_message_ui.py` not checking conditions as expected. Also added a new test case to ensure that he sign of binary files from the SD card is working properly
- Refactors a few other tests to use with `patch("builtins.open"` instead of `mocker.patch("builtins.open"` for better compatibility with VSCode’s test runner

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [x] CHANGELOG

- [x] Yes, on Yahboom

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -211,7 +211,12 @@ def _sign_at_address_from_qr(self, data):
 
     def _sign_at_address_from_sd(self, data):
         """Message signed at a derived Bitcoin address - SD card"""
-        data = data.decode() if isinstance(data, bytes) else data
```
odudex, line 214: Was it here that you were having issues with binary data?

tadeubas: Yes

PR #665, opened 2025-08-11
bugfix: to differentiate between "failed to decrypt",

jdlcdl:
### What is this PR for?
Correctly showing "Failed to decrypt" on wrong decryption key, instead of allowing execution to continue with wrong assumption that original data is worth parsing.

Callers of `encryption_ui.decrypt_kef` were catching all errors, instead of ValueError, and were not deciphering for the important case of `KeyError("Failed to decrypt)` which shouldn't be caught and should disrupt flow.

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Yes

- [x] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/pages/encryption_ui.py
+++ b/src/krux/pages/encryption_ui.py
@@ -54,6 +54,11 @@ def decrypt_kef(ctx, data):
     from binascii import unhexlify
     from krux.baseconv import base_decode, hint_encodings
 
+    # nothing to decrypt or declined to encrypt will raise ValueError here,
+    # so that callers can `except ValueError: pass` to deal with original data.
+    # If user decides to decrypt and fails with wrong key, then
+    # `KeyError("Failed to decrypt")` is raised calling `KEFEnvelope.unseal_ui()`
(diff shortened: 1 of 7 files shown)


src/krux/pages/encryption_ui.py
```diff
@@ -364,7 +369,7 @@ def unseal_ui(self, kef_envelope=None, prompt_decrypt=True, display_plain=False)
         plaintext = cipher.decrypt(self.ciphertext, self.version)
         self.__key = None
         if plaintext is None:
-            raise KeyError("Failed to decrypt")
+            raise KeyError(t("Failed to decrypt"))
```
tadeubas: We don’t translate messages when raising errors; translation is applied only when displaying them (e.g., via `flash_error` after catching the error).

jdlcdl: Thank you!

PR #670, opened 2025-08-20
Reuse texts and small changes on UX for Datum

tadeubas:
### What is this PR for?
Changes in DATUM TOOL:
- Simplifies some texts to improve translations
- Removes the "save to SD?" prompt (users can still cancel using ESC)
- Removes QR transcribe option (most QRs are too big)
- Fixes `identify_datum` misclassifying short text as `XPUB`, `DESCRIPTOR` or `ADDR`

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, yahboom

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


src/krux/pages/datum_tool.py
```diff
@@ -422,7 +422,7 @@ def view_qr(self):
             # when not sensitive, allow export to sd
             kvargs = {}
             if not self.sensitive:
-                kvargs = {"allow_export": True}
+                kvargs = {"allow_export": True, "transcript_tools": False}
```
jdlcdl: Since the condition here is about whether or not contents are sensitive, to disallow saving what might be a mnemonic or mnemonic entropy to sdcard, we might have a len(contents) check separately to decide whether or not transcript_tools are offered... for passphrases and encryption-keys that don't look "sensitive".

tadeubas: Yes, it makes sense. Only QRs that have are bigger than 37x37 should be disabled :+1:

PR #672, opened 2025-08-21
Unicode Ellipsis (1char) replaces ".." (2chars)

jdlcdl:
### What is this PR for?
Alters ".." ellipsis to unicode '\u2026' horizontal ellipsis "…"

  TODO:
* fix i18n.py to post-correct auto-translations where unicode ellipsis were translated to "" or "..." or "……"
* for some reason, firmware's "Shutting down…" kills test suite
* add ellipsis glyph to wide-asian fonts and make accompanying pr to MaixPy so that all projects have new font.c
* on-device tests in english and other locales
* docs images where traditional ".." becomes "…"

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


i18n/i18n.py
```diff
@@ -148,7 +148,23 @@ def print_missing(save_to_file=False, merge_after=False):
             for slug in slugs:
                 if slug not in translations or translations[slug] == "":
                     try:
+                        # fix poorly translated newlines
                         translated = translator.translate(slug).replace(" \ n", "\\n")
+
+                        # fix poorly translated ellipsis
```
tadeubas: Wow! I would not do that, but instead I would replace the `\u2026` to three dots (...) and later after automatic translation replace again the three dots (...) with `\u2026`

firmware/font/FusionPixel-14.bdf
```diff
@@ -1376,6 +1376,14 @@ BITMAP
 A8
 10
 ENDCHAR
+STARTCHAR uni2026
```
tadeubas, line 1379: We have `ter-u14n 16 and 24`, if they already have the `\u2026` why we need to add to `FusionPixel-14`? And I am assuming that `NotoSansCJK-24` and `unifont-16` so do already have `\u2026` right?

src/krux/firmware.py
```diff
@@ -433,5 +433,5 @@ def status_text(text, highlight_prefix=""):
         os.remove(firmware_path)
         os.remove(firmware_path + ".sig")
 
-    display.flash_text(t("Shutting down.."))
+    display.flash_text("Shutting down…")
```
tadeubas: Why remove the `t()` for the "Shutting down..." text?

src/krux/display.py
```diff
@@ -317,11 +317,11 @@ def to_lines_endpos(self, text, max_lines=None):
         # Append/replace ellipsis on last line if we didn't finish the text
         if line_count == max_lines and start < len(text):
             len_last = len(lines[-1])
-            if len_last + 3 > columns:
-                lines[-1] = lines[-1][: columns - 2] + ".."
-                end += columns - len_last - 3
+            if len_last + 1 > columns:
```
tadeubas: Here we don't have the same logic as before I think. Had you found a bug? I think to maintain the original it would be `len_last + 2 > columns`

src/krux/settings.py
```diff
@@ -51,7 +51,8 @@
     NAME_MINISCRIPT,
 ]
 
-THIN_SPACE = " "
+THIN_SPACE = " "  # "\u2009"
+ELLIPSIS = "…"  # "\u2026"
```
tadeubas, line 55: Why add this CONST if is not used anywhere?

tadeubas (commented):
I loved seeing the new images in docs, these new ellipses are beautiful! :heart_eyes: 

Why not add it here on dice rolls too?

image


And on this WARN message from PSBT?

image


And on wallet passphrase display?

image


And on import wallet miniscript descriptor?

image

firmware/font/FusionPixel-14.bdf
jdlcdl: Yes, I was assuming that FusionPixel, unifont, and NotoSansCJK bdf files were the source fonts for wide-asian locale glyphs.  All of the other source files had an ellipsis for `\u2026` except for FusionPixel-14.

src/krux/firmware.py
jdlcdl: Nice catch.   Removing it was not the intention; it was a result of floundering to figure out why replacing ".." with  unicode ellipsis (and failing only its test) was making the entire test suite fail to complete.

src/krux/settings.py
jdlcdl: I thought I'd find places to use it, then never did.  Left it in as "documentation".  After your review, it will be used in at least the wallet_descriptor.

src/krux/display.py
jdlcdl: I need to give this one some thought.

I suspect that the 3 is there because previously the ellipsis was 3 "." chars, and that I forgot to shrink the condition from 3 to 2 when going to "..".

More importantly, it must be missing a test on one of the edges of "can barely fit appending ellipsis" and "too big to append ellipsis", so I'll get test(s) done to prove which one is correct, then I'll adjust or leave it as `len_last + 1` (which I suspect is correct).

i18n/i18n.py
jdlcdl: I'm lost on this one.

The multi-condition if statement is there to document the pattern of poorly translated ellipsis where ellipsis is the very last char of the string.  I'll ask you to explain what you mean because it's not, yet, sinking in.

On an unrelated note 2 lines up, I've been wanting to change that first arg of replace from `" \ n"` to `" \\ n"` for quite some time, because the SyntaxWarning keeps catching my attention... but not sure what (I suspect nothing) will break.???

tadeubas: What I meant is: instead of handling that special case with conditions, I’d normalize it in two steps:

1. **Pre-translation**: Replace every \u2026 (ellipsis) with three plain dots `...`.
2. **Post-translation**: After the automatic translation is done, replace every occurrence of three dots `...` back into `\u2026`.

That way, the translation system only ever sees the simple `...`, and we don’t need a multi-condition check for “ellipsis at the end of a string.”

About the replacement from `" \ n"` to `" \\ n"` — I don’t think it has any effect today. It was probably a workaround for a newline (`\n`) issue in texts and automatic translations. Since we’ve already removed all of them from translated texts, it should be safe to drop this.

jdlcdl: Ok, I understand what you mean now.

While I agree that this would look cleaner, I'll respectfully disagree with it as a solution because if we change the slug in advance of the translation, we lose the info that the translation was "poor" and the chance to learn that it might be getting better over time.

I will try a solution that appears cleaner, makes room for post-translation-processing, and that achieves informing us of how often this is happening.

I'm imagining same logic as the latest "hack", but all in a post_translation_processing() function that documents what we're having to do.  It could receive the slug and translation in order to affect post-translation fixes, output to stderr so that the dev who runs it sees what happens -- and how often, and returns the corrected translation.  I'll fit the existing new-line fix into the same -- without the SyntaxWarning that keeps catching my eyes (as if it's a real problem).  And then we'll have a "clean" place to do the same sort of thing in the future.  In order to test this, I'll delete all of the possible slugs/translations from all locale files and re-run a "fill_merge" to see how well it does, how loud it is, and to be sure that it actually makes no real changes (since the latest commit has translations corrected already).

odudex:
Thank you!

PR #686, opened 2025-08-30
Fix poetry issues and update deps

tadeubas:
### What is this PR for?
- Update dependencies to align with GitHub Actions; otherwise, the workflow may fail there even if it passes locally.
- **embit**: Fix `poetry install` or `poetry lock` fails with `Unable to create package with no name`: Needs this other PR https://github.com/diybitcoinhardware/embit/pull/103
- **ur**: Fix `poetry install` or `poetry lock` fails with `Unable to create package with no name`: Needs this other PR https://github.com/selfcustody/foundation-ur-py/pull/1

### Changes made to:
- [ ] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


pyproject.toml
```diff
@@ -53,14 +55,13 @@ qrcode = "^7.4.2"
 
 [tool.poetry.group.dev.dependencies]
 black = "^25.1.0"
-pylint = "^3.2.5"
-pytest = "^8.2.2"
+pylint = "^3.3.8"
+pytest = "^8.3.5"
```
qlrd: [pytest pip](https://pypi.org/project/pytest/#history) show a more recent version `8.4.1`

pyproject.toml
```diff
@@ -53,14 +55,13 @@ qrcode = "^7.4.2"
...
 pytest-cov = "^5.0.0"
 pytest-mock = "^3.14.0"
 PyQRCode = "^1.2.1"
-pycryptodome = "^3.17.0"
+pycryptodome = "^3.22.0"
```
qlrd: [pycryptodome pip](https://pypi.org/project/pycryptodome/#history) shows a more recent version `3.23.0`

pyproject.toml
tadeubas: thx qlrd, I think we should update

pyproject.toml
tadeubas: thx! Will update too

PR #694, opened 2025-09-04
test: add tests for non-extended pubkey and upper case address

qlrd:
### What is this PR for?

Cover situations where we could (at least theoritcally) use a raw hexadecimal pubkey into `parse_address` method (and then it will skip the `if pubkey.is_extended` clause on `wallet.py::427`) as well a test that check if we use a improper address in uppercase. 

It also removes a repetitive check made after a proper check with `address_to_scriptpubkey` that try fails, then call the `raise` procedure. If the `sc` variable is a `Script` it will pass and will return directly the addr.

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


tests/test_wallet.py
```diff
@@ -739,6 +751,14 @@ def test_wallet_qr(mocker, m5stickv, tdata):
         n += 1
 
 
+def test_fail_invalid_uppercase_address(mocker, m5stickv, tdata):
+    from krux.wallet import parse_address
+
+    data = "BC1QRHJQRZ2D9TDYM3P2R9M2VWZN2SN2YL6K5M357YABCDEFG".upper()
```
odudex: doesn't need the `.upper()` here , right?

qlrd: lol right

qlrd: done

PR #699, opened 2025-09-06
Decrypt a matryoshka as first on-device test

jdlcdl:
### What is this PR for?
Yesterday, @tadeubas found a bug when booting a k210 device via ktool directly from a firmware file, as opposed to booting from flash.   Symptoms were failed KEF decryption with correct key.

Theory is that an uninitialized hardware-accelerated hasher may be the case in uhashlib_hw.pbkdf2_hmac_sha256() because symptoms can be repeated by booting from ktool and trying to decrypt GCM KEFs, until at least a single hw-accelerated sha256() call is made.  So, decrypting any non-GCM KEF (which uses sha256 for auth validation) will fail once then everything works after that.

This pr adds attempting to decrypt a multi-level KEF (which starts with GCM) as the first test in the on-device-test-suite.
If the hasher is NOT initialized, it will fail on the first run after booting from ktool, but will thereafter succeed since hw_acc_hashing() also runs later in that test suite.

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


tadeubas:
ACK – **Tested**. When flashing the device, I can decrypt `AES-GCM` normally with the key. However, when booting the firmware to RAM, decryption of `AES-GCM` always fails. Interestingly, if I run `Device Tests > Test Suite` I will see the `multi_layer_decrypt` test fail, then go back and attempt `AES-GCM` decryption again, it succeeds. Running `Device Tests > Test Suite` for the second time it will succed when booting the firmware to RAM.

Nice analysis, @jdlcdl!

odudex:
Great catch @tadeubas , great analysis @jdlcdl .
Thanks to the clear clues from analysis and developed test tools, it was straight forward to identify the source of the issue: uhashlib_hw.pbkdf2_hmac_sha256() was not ensuring proper initialization of hardware accelerated hasher.
https://github.com/selfcustody/MaixPy/pull/40 should fix it.

odudex:
Thank you! Great work guys!

PR #702, opened 2025-09-08
Loading Data: fall-back to KEF

jdlcdl:
DRAFT: For a future release AFTER v25.09.0

### What is this PR for?

Focus of this PR is to re-prioritize loading of data (scanned or from sd) to be treated as plaintext, within context, instead of first attempting to parse and decrypt a KEF envelope. PSBTs in particular, which are often loaded in binary, form can appear to be KEF, but if they can be parsed as PSBTs first, then user will not have to answer "No".  If parsing fails, in this case -- as a PSBT, we could then fall-back to see if it was KEF, and try again after decryption.

Loading of datum-types TO DO:
- [x] addresses
- [x] descriptors
- [ ] messages to be signed
- [ ] PSBTs

In some cases, this is not possible, a KEF must be checked first, in cases where data is treated as random bytes
and no other obvious parsing could be done... ie) for mnemonic entropy, passphrases, and for decryption keys we
should check if it's a KEF first (rather than treating 32 bytes as a mnemonic, or base43 text as a passphrase, or a string of bytes as the decryption key when in fact these would normally parse as kef envelopes to prompt decryption... and in the case of a false-KEF the user will know NOT to decrypt.

In most other cases, we could try to parse in plaintext form first, and then fall back to trying KEF if it fails.

### Changes made to:
- [x] Code
- [x] Tests
- [x] Docs
- [ ] CHANGELOG

- [ ] Yes

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/docs/getting-started/features/encryption/kef-specifications.en.md
+++ b/docs/getting-started/features/encryption/kef-specifications.en.md
@@ -86,7 +86,59 @@ It is expected that any implementation can decrypt a KEF envelope that was creat
 
 * **On truncated Authentication** At first glance it may be concerning that `auth` bytes for many versions have been truncated and are trivially "weak". Note that KEF's use-case for authentication is to validate that the user has correctly entered their decryption `key`. In the worse case, "false-authenticated" success will occur at a rate of 1:16M (or 1:4B for others) if using an incorrect decryption `key`; similar if an attacker has modified the KEF envelope. In these "false-authenticated" success cases, data will result from decryption, but that data will NOT be the original secret or plaintext; it will be of no value.
 
-## Common Structure of a KEF Envelope
+
+### On Identifying a KEF Envelope
+
+  As noted previously, only the owner of a KEF envelope, having the correct decryption key, can be certain that any string of bytes is indeed a KEF envelope.  Because "auth" bytes are truncated, even a KEF implementation that successfully parses and decrypts an envelope without errors may not know if the plaintext was the original secret or "useless" bytes which resulted in validated decryption under another key for the same truncated `auth` bytes.
+
+  This trait, whether deemed a desirable "feature" or an undesirable "bug" should be carefully considered by implementations that deal with KEF envelopes.  In particular, deciding "when" to offer a "Decrypt?" user-interface may be troublesome for larger byte-strings that can be confused as KEF envelopes. There are three obvious strategies to approach any datum in the context that it might be a KEF envelope:
+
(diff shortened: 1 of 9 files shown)


docs/getting-started/features/encryption/kef-specifications.en.md
```diff
@@ -3,7 +3,7 @@
 ...`The K stands for "KEF"` --anon
 
 
-## Motivation
+## 1. Motivation
```
qlrd, line 6: I was thinking why put numbers is worth. Maybe a pedagogical enforcement for the reader?

docs/getting-started/features/encryption/kef-specifications.en.md
```diff
@@ -75,18 +75,15 @@ It is expected that any implementation can decrypt a KEF envelope that was creat
...
 
 * **On truncated Authentication** At first glance it may be concerning that `auth` bytes for many versions have been truncated and are trivially "weak". Note that KEF's use-case for authentication is to validate that the user has correctly entered their decryption `key`. In the worse case, "false-authenticated" success will occur at a rate of 1:16M (or 1:4B for others) if using an incorrect decryption `key`; similar if an attacker has modified the KEF envelope. In these "false-authenticated" success cases, data will result from decryption, but that data will NOT be the original secret or plaintext; it will be of no value.
 
-## Common Structure of a KEF Envelope
+* **On common `bytes` encodings** KEF does not explicitely define how a KEF envelope is encoded for storage or transport, however each implementation will need to deal with encodings. This topic is further discussed below in the section **On Identifying a KEF envelope**.
```
qlrd, line 82:
```suggestion
* **On common `bytes` encodings** KEF does not explicitely define how a KEF envelope is encoded for storage or transport, however each implementation will need to deal with encodings. This topic is further discussed below in the section [**On Identifying a KEF envelope**](#7-on-identifying-a-kef-envelope).
```

tests/pages/home_pages/test_addresses.py
```diff
@@ -713,6 +713,79 @@ def test_scan_address_fails_on_decrypt_kef_key_error(mocker, m5stickv):
...
+        BUTTON_PAGE_PREV,  # back to Esc
+        BUTTON_ENTER,  # go Esc
+        BUTTON_PAGE_PREV,  # deny key ''
+        BUTTON_ENTER,
+    ]
+    ctx = create_ctx(mocker, btn_seq)
+    addresses_ui = Addresses(ctx)
+    assert addresses_ui.scan_address() == MENU_CONTINUE
+    # assert ctx.input.wait_for_button.call_count == len(btn_seq)
+    print(ctx.display.method_calls)
```
qlrd, line 729: I think we can leave commented when done

PR #705, opened 2025-09-11
fix: strictly checks derivation path

qlrd:
### What is this PR for?

Does a stricter check on `_is_valid_derivation_path` while removes its `try/exception` in a way that excludes the possibility of an unreachable code.

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes: amigo, tested with sparrow

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/tests/pages/home_pages/test_sign_message_ui.py
+++ b/tests/pages/home_pages/test_sign_message_ui.py
@@ -235,6 +235,110 @@ def test_sign_message(mocker, m5stickv, tdata):
         assert ctx.input.wait_for_button.call_count == len(case[3])
 
 
+def test_sign_message_invalid_derivations(mocker, m5stickv, tdata):
+    from krux.pages.home_pages.sign_message_ui import SignMessage
+    from krux.wallet import Wallet
+    from krux.input import BUTTON_ENTER, BUTTON_PAGE, BUTTON_PAGE_PREV
+    from krux.qr import FORMAT_NONE
+    from krux.pages.qr_capture import QRCodeCapture
+
+    # These edge cases will be treated as pure text and the
+    # stuff `signmessage <derivation> ascii:<message>` will
+    # not be parsed and be linked to the address on display
+    # (they will be treated as a whole message)
+    cases = [
+        (
+            [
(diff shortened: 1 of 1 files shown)


src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -69,16 +69,17 @@ def _load_message(self):
...
 
-    def _is_valid_derivation_path(self, derivation_path):
-        """Checks if the derivation path is valid"""
-        try:
-            parts = derivation_path.split("/")
-            return parts[0] == "m" and all(
-                p[-1] in ("'hH") or p.isdigit() for p in parts[1:]
-            )
-        except:
+    def _is_valid_derivation_path(self, derivation_path: str) -> bool:
```
tadeubas: This does not enforce that `derivation_path` must be a `str` at runtime, if someone calls with a binary for example you will get `AttributeError: 'int' object has no attribute 'split'` Exception. Is this fine to the code that actually uses this function?

src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -69,16 +69,17 @@ def _load_message(self):
...
+        """Strictly checks if the derivation path is valid according to BIP32"""
+        parts = derivation_path.split("/")
+        if not parts or parts[0] != "m":
```
tadeubas: I think `.split()` always returns at least one element (e.g. [""]), so `if not parts` will never trigger.

src/krux/pages/home_pages/sign_message_ui.py
qlrd: oh, this annotation was more a pratiice of mine than of strict enforce. The enforce, I think, is more in the `all` method. No problem in remove that.

src/krux/pages/home_pages/sign_message_ui.py
qlrd: true, removed

PR #721, opened 2025-09-18
Tools > Device Tests > Touchscreen

tadeubas:
### What is this PR for?
New device test to let users verify touch detection across the entire screen.

### Changes made to:
- [x] Code
- [x] Tests
- [x] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on TZT and Dock 

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/simulator/kruxsim/mocks/lcd.py
+++ b/simulator/kruxsim/mocks/lcd.py
@@ -47,6 +47,8 @@ screen = None
 portrait = True
 landscape = False
 
+count_call_fill_rectangle = 0
+
 
 def rgb565torgb888(color):
     """convert from gggbbbbbrrrrrggg to tuple"""
@@ -412,7 +414,12 @@ def fill_rectangle(x, y, w, h, color, radius=0):
         x = width() - w - x
     radius = min(radius, min(w, h) // 2)
     pg.event.post(pg.event.Event(events.LCD_FILL_RECTANGLE_EVENT, {"f": run}))
-    time.sleep(0.001)
+
+    global count_call_fill_rectangle
+    count_call_fill_rectangle += 1
+    if count_call_fill_rectangle > 9:
+        time.sleep(0.01)
+        count_call_fill_rectangle = 0
 
 
 def draw_circle(x, y, radious, quadrant, color):


src/krux/pages/device_tests.py
```diff
@@ -153,6 +157,32 @@ def test_suite(self, interactive=False):
...
+            wdt.feed()
+            if self.ctx.input.touch_value() == PRESSED:
+                x, y = self.ctx.input.touch.release_point
+                self.ctx.display.fill_rectangle(x, y, 10, 10, theme.fg_color)
+            elif (
+                self.ctx.input.enter_value() == PRESSED
+                or self.ctx.input.page_value() == PRESSED
+                or self.ctx.input.page_prev_value() == PRESSED
+            ):
+                break
```
qlrd, line 177:
```suggestion
            elif PRESSED in (
                self.ctx.input.enter_value(),
                self.ctx.input.page_value(),
                self.ctx.input.page_prev_value()
            ):
                break
```

simulator/kruxsim/mocks/lcd.py
```diff
@@ -411,6 +412,7 @@ def run():
         x = width() - w - x
     radius = min(radius, min(w, h) // 2)
     pg.event.post(pg.event.Event(events.LCD_FILL_RECTANGLE_EVENT, {"f": run}))
+    time.sleep(0.001)
```
qlrd: `0.001` froze my simulator on first click and logged the follow message on terminal:

```bash
Exception in thread Thread-1 (run_krux):
```

With a bigger value was possible:

```suggestion
    time.sleep(0.01)
```

qlrd (changes requested):
cACK with some nits

src/krux/pages/device_tests.py
tadeubas: Plz make this test:
```
import time

def a():
    time.sleep(1)
    return 1

def b():
    time.sleep(3)
    return 0

def c():
    time.sleep(10)
    return 0


in_time = 0
pre_exec = time.perf_counter_ns()
if 0 in (a(),b(),c()):
    in_time = time.perf_counter_ns() - pre_exec
    print("0 in (s)", in_time/ 1e9)

equal_time = 0
pre_exec = time.perf_counter_ns()
if (a() == 0 or b() == 0 or c() == 0):
    equal_time = time.perf_counter_ns() - pre_exec
    print("== 0 (s)", equal_time/ 1e9)

print("How much faster was == compared to in (s)?", (in_time - equal_time)/ 1e9)
```

What are the results?

simulator/kruxsim/mocks/lcd.py
tadeubas: Can you test again plz? I did use another approach because yours could lag Tinyseed

src/krux/pages/device_tests.py
qlrd: yeah `==` is faster, better to keep the unchanged code.

odudex: The `<button>_value()` functions do not have any delay in them, it's an IO read. Please elaborate on this test, also,  why different delay values?
Removing the delays I get:
0 in (s) 1.279e-06
== 0 (s) 9.6536e-05

odudex: Tested in device (computer mocks IO). And `==` is faster indeed!

odudex:
Thank you!

src/krux/pages/device_tests.py
tadeubas: The test shows that multiple `or` checks can be faster because evaluation short-circuits as soon as a true condition is met, meaning not all function calls are executed. In contrast, `in` (and likewise `any()`) must first evaluate all function results to build the iterable before checking, which adds overhead.

That said, the test was simplified to illustrate the point and doesn’t use the actual firmware code. In reality, functions like `self.ctx.input.enter_value()`, `self.ctx.input.page_value()`, and `self.ctx.input.page_prev_value()` include logic such as conditionals, `wdt.feed()`, and `return` handling, so they cannot be treated as simple I/O reads.

Also, keep in mind that benchmarks can vary depending on many factors. A single run isn’t reliable - I repeated the test without artificial delays and still observed different results:
```
0 in (s) 1.423e-06
== 0 (s) 4.95e-07
```

PR #727, opened 2025-09-21
feat: add Mnemonic XOR (SeedXOR)

### LINKED ISSUE #384: [Enhancement] support SeedXOR in krux
Currently the ColdCard supports SeedXOR https://seedxor.com/. 
I would like to see it in krux, so we could generate and restore seedxor backups, and cross-check with the coldcard's implementation.

qlrd:
### What is this PR for?

This PR add an entry to `Backup Mnemonics > Other formats` menu as well add a new page class to `src/krux/pages/seedxor_ui` called `SeedXOR`. This class allow to choose between 2 up to 4 shares, add entropies from camera to the new shares as well show the mnemonic shares.

Close #384.

### Changes made to:
- [x] Code
- [x] Tests
- [x] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on 

- [ ] Bug fix
- [x] New feature
- [x] Docs update
- [ ] Other

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Krux can now recognize and work with a wider range of SD cards that were previou
 ### Discontinued Support for Maix Bit Device
 The Maix Bit device has long been discouraged due to its poor-quality camera. Starting with this release, we are discontinuing support and it will no longer be included in future builds. The support and parameters for building its firmware from source, however, will be kept.
 
+### Mnemonic XOR
+Krux can now apply XOR operations on entropy bytes between a loaded mnemonic and another chosen one, similar to the Coinkite's `SeedXOR` protocol.
+
 ### Other Bug Fixes and Improvements
 - Touchscreen test added in Tools for detection check
 - wbits for deflate-decompress window set to 10 bits to match KEF spec.


jdlcdl:
Played with this today, I like the flow.  Used 3 XORed mnemonics to get to a my expected 4th, and was able to remove one to get back to a prior mix.

jdlcdl:
I'm not sure I understand why MaixPy has changes.  Please verify that this is intended, I suspect that this branch is pointing to an older MaixPy.

I like that there is a check for the result of another XOR not altering the mnemonic (as if someone used all 0x00s "abandon " * 11 + "about").  Shows error ValueError('XORed mnemonic didn't changed')

I'm wondering if it could also be useful (to avoid confusion and silly errors) to also check for:
* does the result of another XOR produce all 0xFFs like "zoo " * 11 + "wrong" which would happen if someone inverted their mnemonic via that key?
* does the result of another XOR produce all 0x00s like "abandon " * 11 + "about", which would happen if someone used the existing mnemonic.

Similarly, these 3 (or 4) checks could be done earlier, instead of waiting to see what the result of the xor will be, simply by using the arguments passed into xor.
* Is either of the mnemonic entropies all 0x00 bytes (then this will result in unchanged mnemonic)
* is A == B (then this will result in 0x00 bytes)
* Is xor(A, 0xff bytes) == B (then this will result in all 0xff bytes)
* is either of the mnenonics all 0xff bytes (then this will result in inverting the other mnemonic).

qlrd:
> I'm not sure I understand why MaixPy has changes. Please verify that this is intended, I suspect that this branch is pointing to an older MaixPy.

I think was a mistake when pushing. Maybe forgot to make submodules stuff. I will check that

qlrd:
> * Is either of the mnemonic entropies all 0x00 bytes (then this will result: "mnemonic unchanged")
> * is A == B (then this will result in 0x00 bytes: "low entropy mnemonic")
> * Is xor(A, 0xff bytes) == B (then this will result in all 0xff bytes: "low entropy mnemonic")
> * is either of the mnenonics all 0xff bytes (then inverted entropy is result: "mnemonic inverted").

Amazing observations. I will try that and update the PR.

qlrd:
Was thinkin about: `All 0x00 bytes wouldn't result in a new mnemonic` is the same to check "A == B" no? so we can have only 3 checks?

src/krux/pages/home_pages/home.py
```diff
@@ -163,6 +163,14 @@ def bip85(self):
         bip85.export()
         return MENU_CONTINUE
 
+    def mnemonic_xor(self):
+        """Handler for the 'Mnemonic XOR' menu item"""
+        from .mnemonic_xor import MnemonicXOR
+
+        mnemonic_xor = MnemonicXOR(self.ctx)
+        mnemonic_xor.export()
```
odudex: Result from XOR will be loaded, not exported, right? I think `export()` is not appropriated name for the function.

src/krux/pages/home_pages/home.py
```diff
@@ -173,6 +181,7 @@ def wallet(self):
                 (t("Passphrase"), self.passphrase),
                 (t("Customize"), self.customize),
                 ("BIP85", self.bip85),
+                ("Mnemonic XOR", self.mnemonic_xor),
```
odudex: This term should be translated

src/krux/pages/home_pages/mnemonic_xor.py
```diff
@@ -0,0 +1,164 @@
...
+        return status
+
+    def _load_key_from_words(self, words, charset=LETTERS, new=False):
+        """
+        Similar method from krux.pages.login.Login without loading a key,
+        instead, it add the bytes from a mnemonic's entropy to the list of entropies
+        """
+        self.ctx.display.clear()
+        self.ctx.display.draw_centered_text(
+            t("The following mnemonic is the choosen share")
```
odudex: Suggestion:
"Verify the mnemonic before combination. Its entropy will be XORed with currently loaded mnemonic"

src/krux/pages/home_pages/mnemonic_xor.py
```diff
@@ -0,0 +1,164 @@
...
+
+        # XOR mnemonics
+        xored_mnemonic = MnemonicXOR.xor_mnemonics(
+            self.ctx.wallet.key.mnemonic, mnemonic
+        )
+
+        # Show derived mnemonic
+        self.ctx.display.clear()
+        self.ctx.display.draw_centered_text(
+            t("The following mnemonic is the XORed mnemonic")
```
odudex: Suggestion:
"Following mnemonic contains the XOR result. Loading it will replace original mnemonic and discard any passphrase or loaded descriptor"

src/krux/pages/home_pages/mnemonic_xor.py
odudex: If the original stays, "choosen" becomes "chosen".

src/krux/pages/home_pages/mnemonic_xor.py
```diff
@@ -0,0 +1,164 @@
...
+            + ". "
+            + t("It will delete current Key, Passphrase and Descriptor")
+            + "."
+        )
+        if not self.prompt(t("Proceed?"), BOTTOM_PROMPT_LINE):
+            return MENU_CONTINUE
+
+        new_mnemonic = MnemonicEditor(self.ctx, xored_mnemonic).edit()
```
odudex: What does happen if user give-up loading the XOR result?

src/krux/pages/home_pages/mnemonic_xor.py
```diff
@@ -0,0 +1,164 @@
...
+
+        if not self.prompt(t("Proceed?"), BOTTOM_PROMPT_LINE):
+            return MENU_CONTINUE
+
+        from ..mnemonic_editor import MnemonicEditor
+        from ...key import Key
+        from ...wallet import Wallet
+
+        # Show share before XOR
+        mnemonic = MnemonicEditor(self.ctx, " ".join(words)).edit()
```
odudex: What does happen if user give up at this point? Error

odudex:
When XOR is completed, user fall to the loading mnemonic page again, without any prompt asking if they would like to XOR more mnemonics or not.

src/krux/pages/home_pages/mnemonic_xor.py
qlrd: AFAIK, it will not load the XORed mnemonic, but maybe an error could be raised. Good catch

src/krux/pages/home_pages/mnemonic_xor.py
qlrd: An error occurs when give-up, will fix that

src/krux/pages/home_pages/mnemonic_xor.py
qlrd: no error occurs when choose "not proceed"

qlrd:
Applied the points made by @tadeubas in comment above (as well in Private group) and from @odudex in https://github.com/qlrd/krux/pull/16

docs/getting-started/features/mnemonic-xor/index.en.md
```diff
@@ -0,0 +1,34 @@
+# What's a Mnemonic XOR?
+
+It's a implementation of [XOR logical gate](https://www.geeksforgeeks.org/digital-logic/xor-gate/) to operate, indefinitely, an [exclusive OR](https://en.wikipedia.org/wiki/Exclusive_or) upon a loaded mnemonic, based on [Coinkite's SeedXOR](https://github.com/Coldcard/firmware/blob/master/docs/seed-xor.md),
+
+# Why we named `Mnemonic XOR` and not `SeedXOR`?
```
odudex: We won't XOR BIP39 seeds, nor copy the SeedXOR mental model or UI (with splitting, backup, recover, etc), so this term must be removed from translations and docs (Except one a mention with credits for the idea of using XOR).
We are just doing mnemonic entropy XOR, or "Mnemonic XOR". We can explain how it works and what can be done, but we don't need to sound guilty for having a different UI model.
It's false that "we do only the recover part". The idea is we do Mnemonic XOR, and it can be used for both splitting and joining mnemonic entropy, and at the same time user learns about the properties of XOR.

docs/getting-started/features/mnemonic-xor/three-shares.en.md
```diff
@@ -0,0 +1,91 @@
+
```
odudex: No need for 3 shares, 4 shares dedicated file... an addendum can be added to a "xor-splitting-mnemonics.en.md" file mentioning the example with splitting a mnemonic into two case can be modified to 3, 4, 5...

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ Krux can now recognize and work with a wider range of SD cards that were previou
 ### Discontinued Support for Maix Bit Device
 The Maix Bit device has long been discouraged due to its poor-quality camera. Starting with this release, we are discontinuing support and it will no longer be included in future builds. The support and parameters for building its firmware from source, however, will be kept.
 
+### Mnemonic XOR
+Krux can now apply XOR operations on entropy bytes between a loaded mnemonic and another chosen one, similar to the Coinkite's `SeedXOR` protocol.
```
jdlcdl, line 19: suggest:

Apply binary-XOR operator between the entropy bytes of your loaded mnemonic and another mnemonic, compatible with Coinkite's `SeedXOR` protocol.

docs/getting-started/features/mnemonic-xor.en.md
```diff
@@ -0,0 +1,82 @@
+# What's the Mnemonic XOR?
+
+It's a implementation of [XOR logical gate](https://www.geeksforgeeks.org/digital-logic/xor-gate/) to operate, indefinitely, an [exclusive OR](https://en.wikipedia.org/wiki/Exclusive_or) upon a loaded mnemonic, based on [Coinkite's SeedXOR](https://github.com/Coldcard/firmware/blob/master/docs/seed-xor.md).
```
jdlcdl, line 3: suggest:

Compatible with [Coinkite's SeedXOR](https://github.com/Coldcard/firmware/blob/master/docs/seed-xor.md), this feature applies the binary-XOR operator between entropy bytes of your loaded mnemonic and another, resulting in -- and optionally loading, a new mnemonic which acquires an interesting `lossless hiding` attribute from the XOR operation.

docs/getting-started/features/mnemonic-xor.en.md
```diff
@@ -0,0 +1,82 @@
...
+
+# How it works
+
+To derive a new mnemonic (and thus, a new seed) from other mnemonics, an operation occurs with the **mnemonic's entropy bytes**.
```
jdlcdl, line 7: suggest:

XOR is an operation that takes two independent `0` or `1` inputs, and results `1` if both are the same `0` if they are different.

Mnemonic XOR effectively `masks` your loaded mnemonic with another, resulting in a new mnemonic that can restore either of its input mnemonics by providing the other input mnemonic.

Using BIP39 entropy-bytes between two mnemonics, XOR is applied to each bit so that the result is the set of both entropy-bytes.  This can be repeated with more unique mnemonics, indefinitely, and each of its input mnemonics can be restored from the result by providing all of the `other` input mnemonics.  We can XOR mnemonic A with B, creating C, and then we can XOR mnemonic C with A to restore B -- or with B to restore A.

docs/getting-started/features/mnemonic-xor.en.md
```diff
@@ -0,0 +1,82 @@
...
+Operate a XOR between them will derive a new **entropy bytes** that will be converted to a new mnemonic and then, to a new seed:
+
+<img src="../../../img/mnemonic_xor.png" align="center">
+
+- We get two different mnemonics (A and B), extract their *entropy bytes*;
```
jdlcdl, line 12: - Using *entropy bytes* of two independent mnemonics A and B,

docs/getting-started/features/mnemonic-xor.en.md
```diff
@@ -0,0 +1,82 @@
...
+- validate the input entropies to avoid useless or dangerous operations:
```
jdlcdl, line 13: - validation of each input mnemonic's *entropy-bytes*,

docs/getting-started/features/mnemonic-xor.en.md
```diff
@@ -0,0 +1,82 @@
...
+  - `A XOR B = A`: it will not change the XORed mnemonic;
+  - `A XOR B = A'` where `A'` is a "inverse" version of `A`;
+- once the inputs are checked, krux will apply a XOR between the *entropy bytes*;
```
jdlcdl, line 16: - apply binary XOR between the input mnemonics' *entropy-bytes*,

docs/getting-started/features/mnemonic-xor.en.md
```diff
@@ -0,0 +1,82 @@
...
+- validate the output entropy (same as above);
```
jdlcdl, line 17: - validate resulting output *entropy bytes*,

docs/getting-started/features/mnemonic-xor.en.md
```diff
@@ -0,0 +1,82 @@
...
+
+You can split a mnemonic into two separate mnemonics (or "shares") using the XOR operation. Neither share reveals any information about the original secret on its own. The original mnemonic can only be recovered when **both shares are combined**.
+
+### Core Principle
+
+Now we show a basic step
+If `A XOR B = C`, then `B XOR C = A`.
+
+- **A**: Your original mnemonic (the secret to protect);
+- **B**: A newly generated, random mnemonic (Share 1);
```
jdlcdl, line 31: - **B**:  Another backed-up, random mnemonic;

docs/getting-started/features/mnemonic-xor.en.md
```diff
@@ -0,0 +1,82 @@
...
+- **C**: The resulting mnemonic from the XOR operation (Share 2).
```
jdlcdl, line 32: - **C**: The resulting mnemonic from the XOR operation

PR #730, opened 2025-09-22
Clean up touch checks and fix touch settings menu bug

tadeubas:
### What is this PR for?
- Make use of kboard to check for touch in all screens
- Fix Settings > Hardware > Touchscreen menu appearing when touch device had touch disabled
- Change Pylint to check for unused imports on `_init_.py` files
- Removed few unused imports on `src/krux/pages/__init__.py`

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, build and tested on TZT

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -257,7 +257,7 @@ dummy-variables-rgx=_+$|(_[a-zA-Z0-9_]*[a-zA-Z0-9]+?$)|dummy|^ignored_|^unused_
 ignored-argument-names=_.*|^ignored_|^unused_
 
 # Tells whether we should check for unused import in __init__ files.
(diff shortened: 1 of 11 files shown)


odudex:
I didn't see in the code what device without touch has touch settings.

Or if you refer to "disabled touch" as it was an option, it's not. It's a hardware fault protection. In this case, there's no reason to hide the setting, or, even worse, a wrong setting could trigger the fault. Hiding the setting would lock the user out of being able to change it again.

PR #731, opened 2025-09-23
Work to optimize DatumTool: don't die on contents >30K

jdlcdl:
### What is this PR for?
@Tadeubas has found that large files (30k+) which should be workable on device do not succeed in loading in datum tool.

Optimizations achieved via:
* SLOW_ENCODING_MAX_SIZE at 16K to avoid some checks/operatioins on large contents (base43 especially)
* SUFFICIENT_SAMPLE_SIZE at 512 to avoid sampling entire contents in detect_encodings and info_box preview
* identify_datum() receives `encodings` as call param (optional) to avoid calling slow detect_encodings() again,
* show_contents will use bytes data instead of converting to hex for entire contents, then will page those bytes
   hexlified per page, adding an extra ellipsis when a half-byte/nyble would be shown as last char on page.
* delete temp copies of contents and garbage collection where appropriate
* avoid resets via feeding watchdog after operations that can be slow
* avoid user concern by displaying "Processing" when slow and/or "Failed to convert" (ie: mem alloc err)

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on amigo, yahboom_devkit, and various simulators

- [x] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/pages/datum_tool.py
+++ b/src/krux/pages/datum_tool.py
@@ -376,7 +376,8 @@ class DatumToolMenu(Page):
         if not contents:
             return MENU_CONTINUE
 
-        self.ctx.display.draw_centered_text("\n\n" + t("Processing…") + "\n\n")
+        self.ctx.display.clear()
+        self.ctx.display.draw_centered_text(t("Processing…"))
 
         # utils.load_file() always returns binary
         try:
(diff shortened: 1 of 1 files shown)


odudex:
Processing screens are a nice addition. I suggest you to use `self.ctx.display.clear()` before them to avoid text superposition and avoid `"\n\n"` use.

PR #740, opened 2025-10-01
test: add `wonder_k` board to tests

qlrd:
### What is this PR for?

This PR add `board_wonder_k` function to `conftest.py` as well adapt some tests to cover cases when `multiple_devices` fixture is used.

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other: test


tests/shared_mocks.py
```diff
@@ -828,3 +887,29 @@ def mock_context(mocker):
                 draw_hcentered_text=mocker.MagicMock(return_value=1),
             ),
         )
+
+    # TODO: adjust to real dimensions
```
odudex: Is there something to adjusted yet? The display is 240x320, so I believe it's right already.

To get a cleaner and more maintainable code, maybe it would be a good idea to create some definitions like:
```
USABLE_WIDTH_240x320 = 240 - 2 * 10
USABLE_PIXELS_IN_LINE240x320 =  240 - 2 * 10
USABLE_CHARS_P_LINE_240x320 = (240 - 2 * 10) // 8
...
```
And re-use them with similar displays, as most of them are 240x320 and use the same fonts

qlrd: > Is there something to adjusted yet? The display is 240x320, so I believe it's right already.

I believed that this could be wrong.

PR #745, opened 2025-10-20
fix: low entropy calc when using Mnemonic XOR

qlrd:
### What is this PR for?

This commit add a fix for shannon entropy calculation of a choosen entroypy bytes before operate the XOR between input and current mnemonic.

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/pages/home_pages/mnemonic_xor.py
+++ b/src/krux/pages/home_pages/mnemonic_xor.py
@@ -20,6 +20,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
+import math
 from .. import Menu, LETTERS, MENU_CONTINUE, MENU_EXIT
 from ..login import MnemonicLoader
 from ...krux_settings import Settings, t
@@ -27,6 +28,15 @@ from ...display import BOTTOM_PROMPT_LINE, FONT_HEIGHT
 from ...themes import theme
 from ...key import Key
 from ...wallet import Wallet
+
+ENTROPY_THRESHOLD = 3.418296
+"""Calibrated in :py:func:`test_export_xor_fail_low_entropy` extremes
+(0x00 / 0xFF), a flat distribution (~2.29 bits/word), and a
+borderline-acceptable mnemonic (~3.42).
+
+This measures histogram spread of derived 11-bit word indices (ENT||checksum),
+not general cryptographic quality.
+"""
 
 
 class MnemonicXOR(MnemonicLoader):
@@ -38,8 +48,13 @@ class MnemonicXOR(MnemonicLoader):
     # See implementation reference at
     # https://github.com/Coldcard/firmware/blob/2445b4d4350d0aad0f4ef8f966697a67ab9ecbdc/shared/utils.py#L752
     @staticmethod
-    def _xor_bytes(a: bytes, b: bytes) -> bytearray:
+    def _xor_bytes(a, b):
         """XOR two byte sequences of equal length"""
+
+        if not isinstance(a, (bytes, bytearray)) or not isinstance(
+            b, (bytes, bytearray)
+        ):
+            raise TypeError("Sequences should be bytes or bytearray")
 
         # All sequences should have same length because it would
         # reveal the last bytes from the longer bytestring as they are.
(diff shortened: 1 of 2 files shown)


odudex:
Shannon of derivatives (variation), instead of raw data, was used in dice rolls to detect arithmetic progression patterns, but to have Shannon measurement you shouldn't derive the data.
If you're going to measure it byte-wise, the "histogram" or bucket will range from 0 to 255. But in the context of mnemonics, to have a entropy measurement maybe a 11bit words should be used, with 2048 possible values for the bucket.

qlrd:
Tested with device and the mnemonics shared by @jdlcdl on https://github.com/selfcustody/krux/pull/727.

qlrd:
re-ready, with 
<div>
Captura de Tela 2026-04-29 às 13 15 26
<p> <b>1. Stats for Nerds</b> </p>
<div>

<div>
Captura de Tela 2026-04-29 às 13 15 40
<p> <b>1st round xor</b> </p>
</div>

<div>
Captura de Tela 2026-04-29 às 13 16 29
<p> <b>2nd round xor</b> </p>

EDIT: removed. Just showing the fingerprint and the message Low Entropy

odudex:
IMO:
- Mnemonic XOR is not a tool to create entropy, it only manipulates it, for this reason stats for nerds do no fit here.
- The concept alone is already hard for most people to digest, so it must be as simple and straight forward as possible to not overwhelm user.
- Flow must be interrupted only in case low entropy inputs or result is detected under the hood.
- No setting should be created for entropy threshold. A value which covers cases which are already being blocked is enough.

In sum: Simplify, simplify!

qlrd:
> IMO:
> 
>     * Mnemonic XOR is not a tool to create entropy, it only manipulates it, for this reason stats for nerds do no fit here.
> 
>     * The concept alone is already hard for most people to digest, so it must be as simple and straight forward as possible to not overwhelm user.
> 
>     * Flow must be interrupted only in case low entropy inputs or result is detected under the hood.
> 
>     * No setting should be created for entropy threshold. A value which covers cases which are already being blocked is enough.
> 
> 
> In sum: Simplify, simplify!

Done

src/krux/pages/home_pages/mnemonic_xor.py
```diff
@@ -151,30 +210,68 @@ def _load_key_from_words(self, words, charset=LETTERS, new=False):
...
+            return MENU_CONTINUE
+
+        self.ctx.display.clear()
+
+        # XOR the mnemonic entropy and if we receive any error, just warn the user.
+        try:
+            xored_mnemonic = self._xor_with_current_mnemonic(mnemonic_to_xor)
+            return self._load_xored_mnemonic(fingerprint_to_xor, xored_mnemonic)
+        except Exception as e:
+            self.ctx.display.clear()
```
odudex, line 268: use flash_error()

src/krux/pages/home_pages/mnemonic_xor.py
```diff
@@ -151,30 +210,68 @@ def _load_key_from_words(self, words, charset=LETTERS, new=False):
         if not self.prompt(t("Proceed?"), BOTTOM_PROMPT_LINE):
             return MENU_CONTINUE
 
-        # If last word do not match, raise an error
-        MnemonicXOR._check_last_word(xored_mnemonic)
-
         # Show the XOR result
+        if xored_mnemonic is not None:
```
odudex, line 214: this check is not needed (dead code)

src/krux/pages/home_pages/mnemonic_xor.py
```diff
@@ -151,30 +210,68 @@ def _load_key_from_words(self, words, charset=LETTERS, new=False):
...
+        if not self.prompt(t("Proceed?"), BOTTOM_PROMPT_LINE):
+            return MENU_CONTINUE
+
+        self.ctx.display.clear()
+
+        # XOR the mnemonic entropy and if we receive any error, just warn the user.
+        try:
+            xored_mnemonic = self._xor_with_current_mnemonic(mnemonic_to_xor)
+            return self._load_xored_mnemonic(fingerprint_to_xor, xored_mnemonic)
+        except Exception as e:
```
odudex, line 267: Catching bare Exception and showing str(e) will surface any error (TypeError, KeyError, etc.) as if it were a user-facing entropy issue. Consider catching ValueError only.

src/krux/pages/home_pages/mnemonic_xor.py
```diff
@@ -54,27 +69,81 @@ def _xor_bytes(a: bytes, b: bytes) -> bytearray:
...
+
+    @staticmethod
+    def _bits_of_bytes(data):
+        """MSB-first bitstring; used to form 11-bit tokens."""
+        if not isinstance(data, (bytes, bytearray)):
+            raise TypeError("Data should be bytes or bytearray")
+        return "".join("{:08b}".format(b) for b in data)
+
+    @staticmethod
+    def _word11_histogram(entropy):
```
odudex, line 88: _word11_histogram builds a Python string then re-parses with int(..., 2).
int.from_bytes + bit shifting would avoid 264 chars of allocation per call

odudex (commented):
Shannon entropy of 12 samples over 2048 bins is a high-variance estimator, the gate is really a "uniqueness ratio" check dressed in entropy notation.
In practice, it's doing a very fancy calculation to detect duplicated words. The threshold will determine how many duplicates are allowed. And here there's a big problem, with the threshold set to 3.418296, a single duplicate in a 12 words mnemonic will fail.
Measuring Shannon with a sample (12 or 24)  that is **much** smaller than "histogram" size (2048) is weird. And a shared threshold between 12 and 24 words does not make sense statistically too.
| Word count | Max possible H | Threshold | Headroom above threshold |
| -------- | -------- | -------- | -------- |
|12 | 3.585 bits/word | 3.418 | 0.167 (~5%) |
|24 | 4.585 bits/word | 3.418 | 1.167 (~25%) |

odudex:
Sorry I didn't get this deep on previous review, but now with less code I could see better

Issue #522, opened 2025-03-05
[Enhancement] Passphrase QR Code scan process can detect and decode Encrypted QR Code


comassky:
Hi Krux !

I love the mechanics of scanning an encrypted qr code that asks for the decryption key.

This process is requested when the mnemonic loading via QRCode feature is used.

I realized that passphrase scanning doesn't use this “if qr code encrypted, then I propose to decrypt it via keyboard or qr code” mechanic.

Is it possible to apply this to passphrase scanning?

This would allow you to store your passphrase in encrypted form.

In my opinion, this is a gain in terms of usage (especially since the process is already in place for loading wallets).

tadeubas:
For me it is strange this kind of request because:
- Encrypted mnemonics gives you the ability to use a password
- The mnemonic gives you the ability to use a passphrase

So you will use a password, then another password to get the passphrase... we could do this multiple times too, but what is the use case? If you want the passphrase in encrypted form, the password will not be plain text? Using the password to decode this passphrase for me is the same as using this password as the passphrase itself.

To be crystal celar, what we gain from `password to decrypt a -> password to decrypt a -> password to decrypt a -> password to decrypt a -> passphrase` ?  Would it not be the same of a `password to decrypt a -> password to decrypt a -> password to decrypt a -> password to decrypt a -> mnemonic` ? Why you need Krux to do that? why is that better than a strong password?

jdlcdl:
Indeed, if you were to use a normal krux encrypted-mnemonic QR as your passphrase, then the resulting plaintext will NOT be a string of 12 words separated by spaces, it will rather be 16 bytes of binary data (exactly like a compact SeedQR which represents bip39 entropy bytes).

In order to achieve the result you're looking for, you would need to use Datum Tool (in tools before loading mnemonic).
* Datum Tool
* Via Manual Input
* enter your 12 word mnemonic with lowercase letters and one space between each word, then hit Go
* Convert Datum
* from UTF-8 (you're changing the long string of text into bytes)
* Encrypt (and follow instructions:  AES-GCM is a sane default but any version would work, use a strong key, edit the ID to something other than "Custom Text")
* to base43 (this will setup for the most compact QR, but it will be big... 12 words are much bigger than 16 bytes)
* Done Converting
* QR Code

I'm honestly not sure how big the resulting QR will be.

[...]

PR #756, opened 2025-10-29
BugFix: Passphrase (via encrypted mnemonic entropy) fails gracefully

jdlcdl:
related to #522
replaces #755

### What is this PR for?
To resolve a bug found by @tadeubas around encrypted passphrases when a krux encrypted-mnemonic is scanned.
User doesn't really want encypted bip39 entropy bytes, they want encrypted lowercase-space-separated mnemonic string as a passphrase.

In Cpython a UnicodeDecodeError was being caught/ignored by the except ValueError leg,
while in MaixPy, a TypeError was being raised, not caught, bubbling up:
...whenever an encrypted passphrase returned bytes that could not be decoded to a string.

Symptoms:
on simulator: appeared as if NOT KEF or user declined to decrypt,
on k210: TypeError("Can't convert 'int' object to str implicitely",)

Fix:
to decode decrypted plaintext in its own try/except block, separate from try_decrypt()

Similar pattern has been found, and solved for:

wallet descriptors
addresses

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, build and tested on simulator and amigo

- [x] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/pages/wallet_settings.py
+++ b/src/krux/pages/wallet_settings.py
@@ -91,7 +91,7 @@ class PassphraseEditor(Page):
                     self.flash_error(t("Failed to load"))
                     continue
             # Check if passphrase string is within ascii range
-            if any(byte > 127 for byte in passphrase.encode()):
+            if any(byte > 126 for byte in passphrase.encode()):
                 self.flash_error(t("Failed to load"))
                 continue
 


src/krux/pages/wallet_settings.py
```diff
@@ -83,6 +83,18 @@ def load_passphrase_menu(self, mnemonic):
...
 
+            # Decode passphrase in case it's a "bytes" object
+            if isinstance(passphrase, bytes):
+                try:
+                    passphrase = passphrase.decode()
+                except:
+                    self.flash_error(t("Failed to load"))
+                    continue
+            # Check if passphrase string is within ascii range
+            if any(byte > 127 for byte in passphrase.encode()):
```
tadeubas: I think here it is byte > 126 as ord(~) == 126

odudex: Did a PR atop @jdlcdl 's with the change to 126

jdlcdl: right, its the unprintable DEL

Issue #399, opened 2024-05-28
[Enhancement] Inclusion of special characters as accents in the letters for pashphrases


brunohitman:

jdlcdl:
Just a thought on this:
* when creating a new bip39 passphrase... device should be strict and only allow keys that are most common to most keyboards, else it could be difficult to recover on different devices/wallets.
* when recovering an existing bip39 passphrase... device should be tolerant in allowing as many keyboards as might have been used to create the original wallet.
Not sure how best to achieve that on a stateless device, but would love to hear others' ideas on this.

tadeubas:
I've tested this with Sparrow, [Seed Tool](https://bitcoiner.guide/seed/) and Krux create a different fingerprint when scanning `á` as passphrase from QR code. This is a problem even if Krux can't type these special characters, both Sparrow and Seed Tool show `411905a1` for fingerprint, but Krux show `9325504a`

brunohitman:
This is a big problem!

odudex:
Incorporating a variety of language glyphs could compromise the user experience and pose risks; therefore, we currently have no plans to implement such a feature.

earthdiver:
> I've tested this with Sparrow, [Seed Tool](https://bitcoiner.guide/seed/) and Krux create a different fingerprint when scanning `á` as passphrase from QR code. This is a problem even if Krux can't type these special characters, both Sparrow and Seed Tool show `411905a1` for fingerprint, but Krux show `9325504a`

Isn’t this because Krux wasn’t fully compliant with the BIP-39 specification (UTF-8 NFKD)? It needs to perform normalization to NFKD
-- for example: passphrase = unicodedata.normalize("NFKD", passphrase).

In release 25.10.1, it seems the only hardware wallet supporting BIP-39–compliant passphrases that allow UTF-8 may now be CryptoGuide’s heretical fork. Perhaps it wasn’t necessary to insist on strict ASCII -- UTF-8 would likely have been fine (for example, the probability that a random 16-byte sequence is recognized as valid ASCII or UTF-8 is on the order of 10^-5 and 10^-4, respectively; even the latter is sufficiently small, and the probability drops rapidly toward zero as the data size increases).

odudex:
[...]

PR #769, opened 2025-11-15
Give warning for non-ASCII BIP39-passphrase

tadeubas:
### What is this PR for?
BIP39 requires passphrases to be pass NFKD normalization, unfortunately we cannot do that on our end because of firmware size constraints so we decide to deny non-ascii passphrases starting from [v25.10.1](https://github.com/selfcustody/krux/releases/tag/v25.10.1).

This PR changes the denial of non-ascii passphrases to warn instead, allowing users to use QR codes with non-ascii text NFKD normalized. Below are some examples:

**à without NFKD** normalization == à (if you copy this à and put on Seed Tool is gives e15d2228):
a

**à WITH NFKD** normalization == a` (if you copy this à and put on Seed Tool is also gives e15d2228):
b

**Seed Tool** fingerprint with any à or à (I know it appears the same char but they are not :smirk: )
Screenshot from 2025-11-14 11-16-49

Krux reading **without NFKD** (on simulator works, on device it gives error):
Screenshot from 2025-11-14 11-15-04

Krux reading **WITH** NFKD:
Screenshot from 2025-11-14 11-16-14

Krux warning message when loading the passphrase via QR-code with non-ASCII:
Screenshot from 2025-11-14 11-14-53

### Changes made to:
- [x] Code
- [ ] Tests
- [x] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on Embed Fire

- [ ] Bug fix
- [x] New feature
- [x] Docs update
- [ ] Other


jdlcdl:
This worked for me, in that it was able to load the NFKD normalized two-character three-byte accented-"a".  I saw the warning which is great... but then later I saw "aa" when confirming the passphrase, because firmware is missing fonts for what are likely common accents by themselves.  Still, I derived the correct bip32 wallet which is the whole point.

But when I loaded the one-byte version of the accented-"a" in the upper-ascii/latin-1 range, I got "failed to load" while scanning the QR.  I believe this is the correct thing to do (because it's neither ASCII nor normalized UTF-8), but I wonder if others already have wallets that would need this.

@earthdiver and @brunohitman chimed in on an earlier issue that has since been closed, #399 and if they can see this PR, I'd love to have their input regarding whether this fixes it for them.
NOTE: If a user knows the hex values for the bytes of the non-ascii passphrase which leads to their wallet, they can type these into Datum Tool as hex, then Convert "from hex" to bytes, before exporting a binary QR that can be used as a passphrase.

earthdiver:
I’ve reviewed the changes to the source code. I’m really glad that, with the updates in this PR, my NFKD-normalized passphrase should be usable again. Thank you very much.

earthdiver:
It might be helpful to include a short code snippet in the document showing how to perform NFKD normalization on a PC.

on windows(PowerShell):
```
Write-Output ('NFKD-normalized utf-8 string : ' + 'utf-8 string'.Normalize('FormKD'))
Write-Output ('NFKD-normalized utf-8 in hex : ' + ([Bitconverter]::ToString([Text.Encoding]::UTF8.GetBytes('utf-8 string'.Normalize('FormKD')))).Replace('-','').ToLower())
```
on linux(python):
```
python3 <<EOD
import unicodedata
print('NFKD-normalized utf-8 string : ',unicodedata.normalize('NFKD','utf-8 string'))
print('NFKD-normalized utf-8 in hex : ',unicodedata.normalize('NFKD','utf-8 string').encode('utf-8').hex())
EOD
```

src/krux/pages/wallet_settings.py
```diff
@@ -92,8 +92,19 @@ def load_passphrase_menu(self, mnemonic):
...
             # Check if passphrase string is within ascii range
             if any(byte > 126 for byte in passphrase.encode()):
-                self.flash_error(t("Failed to load"))
-                continue
+                self.ctx.display.clear()
+                self.ctx.display.draw_hcentered_text(
+                    t(
+                        "Non-ASCII characters were detected in your passphrase. "
+                        "Krux cannot guarantee that other wallets will derive the "
+                        "same BIP39 fingerprint."
```
odudex: The BIP39 fingerprint is just an ID, or easy to remember object that represent a seed/key. To be clear that a different key/wallet can be created, I think we could warn:
"Krux cannot guarantee that other tools will derive the same **key**."

tadeubas: Nice catch!changed.

odudex:
Thank you!

PR #771, opened 2025-11-15
UI: Context arrows, customizable colors, and a page index indicator for menu items + more

tadeubas:
#### Menu Items:
- Add **context arrows** and **custom color** to _items_ (we had custom color just for disable)
- New **pages index** on _menus_ with more than one page
- Add `< Back` on missing _menus_

#### Touch feedback highlight
- Added **touch-feedback highlighting** to _menu items_, _keypads_, _confirmation prompts_, _category settings_, _Mnemonic Editor_, _Stackbit 1248_ and _Tinyseed_

#### Keypads
- Improved **keyset index** visuals
- New style for **key indicator**

~~#### Touch / Swipe handling~~
~~- **Diagonal swipes**: Swipes are now accepted only if the delta on one axis is 2 times greater than the other~~
~~- **Long-hold**: Long hold swipes (>750ms) are now ignored~~
~~- **Edge touches**: Discards touches near edges of adjacent regions~~
PR #797  

~~#### Settings / Constants~~
~~- Increased range for **Touch Threshold** and **Buttons Debounce**~~
~~- Reduced **Buttons Debounce** default value a little (customized default for _M5StickV_)~~
~~- Slightly reduced the Swipe Threshold constant~~
PR #793 

image

image

image

image

image

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on Amigo, Embed Fire and M5StickV

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other


tadeubas:
Just wrapped up the test coverage for this PR, it came out to about 420 new lines (~150 of those are tests).

Here’s how I think it’s best to review it:

1. **First**, check that no existing tests were removed or altered in a way that hides a real issue. In other words, confirm that nothing was changed just because a code update broke a test

2. **Second**, flash a touch-enabled device and verify the behavior described in the PR’s “What is this PR for?” section. You can compare it against another device running `develop` or the latest release to spot differences (if you want / can do this with a non-touch device too)

3. **Third**, note that the code changes are all within input/touch, pages and other UI-related parts. Because of that, the hands-on device testing you did in step 2 is far more valuable than doing a deep code review. So no need to spend too much time trying to dissect every line of the diff

i18n/translations/es-MX.json
```diff
@@ -70,7 +70,6 @@
     "Device Tests": "Pruebas del dispositivo",
     "Display": "Pantalla",
     "Do not power off, it may take a while to complete.": "No apagues el dispositivo, puede tardar un tiempo en completarse.",
-    "Done Converting": "Listo para convertir",
```
kdmukai: My Spanish is just okay, but I would guess that this should be more like "Conversión terminado", similar to the French construction.

tadeubas: This line was removed, this translation doesn't exist anymore :smile:

kdmukai: Oh, ha. Even better then!

tadeubas:
> Note: the `< | Go | >` style pages do not have the touch highlight.

I didn’t update those category-settings screens because they’re very simple, and I figured the visual change would already be clear there. But if you think it feels inconsistent since the other screens have touch feedback, I can add it, no problem.

tadeubas:
> For screens with a `< Back` option, maybe worth considering supporting swipe (from left to right) as well?

I’m not sure… for now, swipe up/left goes to the next page, and swipe down/right goes to the previous one on menus that have multiple pages. Only a few menus actually do, like: `Tools > Device tests > Check SD Card`, `Settings > Hardware > Printer > CNC`, or when exporting QR codes as images or using Datum convert. On the M5StickV, a few more menus can span multiple pages.

kdmukai:
> But if you think it feels inconsistent since the other screens have touch feedback, I can add it, no problem.

Yeah, I think this sort of feedback becomes a subconscious expectation and then feels a tiny bit jarring when it's missing.

During SeedSigner's UX overhaul, Easy really held my feet to the fire to ensure we were being consistent everywhere. Tough pages required a tiny bit of extra flexibility in a handful of cases, but then even those exceptions were made to at least conform to each other (i.e. "you can't violate principle X, Y, or Z!" "Um, well, I need to violate Y here, here, and here." "Errg, fine, but only do it using approach "Y.b.").

kdmukai:
> since we may not have enough space to print the arrow when the text uses all available pixels, we could end up with something like this

Ya, probably more work than it's worth, but you'd need some sort of smarter `draw_hline` (or somewhere around there): If there's room inline, render the context arrow attached to the text. If it's forced to multiline, detach the context arrow and respect the space.

tadeubas:
> Yeah, I think this sort of feedback becomes a subconscious expectation and then feels a tiny bit jarring when it's missing.

Sure will do this!

> If there's room inline, render the context arrow attached to the text. If it's forced to multiline, detach the context arrow and respect the space.

This I would leave to another PR :wink:

odudex:
Would you mind splitting this PR so we get the chance to discuss each change?

tadeubas:
I’ll open another PR referencing this one. If everything looks good and we merge or close it, I’ll create additional ones for the remaining items.

tadeubas:
Another idea for multiple pages is to show dots instead of bars:

image

jdlcdl:
I see this old pr and will address it one commit at a time, starting from the top-down.  Please feel free to rebase on top of develop, but please don't squash commits so that I can work through the entire pr.

qlrd:
cACK.

I will do some review and let you guys now about.

qlrd (commented):
Needs rebase and resolve several conflicts. Was able to rebase and resolve them mannually (i think -- trying to not break) until b4bde201 at `src/krux/pages/settings_page.py`.

Also a reword on git main message (maybe `refactor(ui)`? to be accepted on CI or a follow-up to include `ui` on conventional-commits step)

Issue #760, opened 2025-11-10
[Enhancement] More Consistent Screensaver


odudex:
Screensaver is inconsistent now, being active only in some menu screens.
This behavior is interpreted as malfunction by most users.

PR #772, opened 2025-11-17
Screensaver now appears in screens with infobox

tadeubas:
### What is this PR for?
Screensaver now appear in all menu screens (even with info boxes) as requests from issue #760 

Tested all screens that are changed by this PR, waiting for screensaver and seeing the result after, more than one time

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on Embed Fire

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/src/krux/pages/device_tests.py
+++ b/src/krux/pages/device_tests.py
@@ -159,28 +159,31 @@ class DeviceTests(Page):
                 wdt.feed()
 
         # info_box summary
-        self.ctx.display.clear()
-        failures = len([x for x in self.results if not x[1]])
-        total = len(self.results)
-        num_lines = self.ctx.display.draw_hcentered_text(
-            "\n".join(
-                [
-                    x
-                    for x in [
-                        t("Test Suite Results"),
-                        t("success rate:")
-                        + " {}%".format(int((total - failures) / (total) * 100)),
-                        (
-                            t("failed:") + " {}/{}".format(failures, total)
-                            if failures
-                            else None
-                        ),
+        def _print_infobox():
+            self.ctx.display.clear()
+            failures = len([x for x in self.results if not x[1]])
+            total = len(self.results)
+            return self.ctx.display.draw_hcentered_text(
+                "\n".join(
+                    [
+                        x
+                        for x in [
+                            t("Test Suite Results"),
+                            t("success rate:")
+                            + " {}%".format(int((total - failures) / (total) * 100)),
+                            (
+                                t("failed:") + " {}/{}".format(failures, total)
+                                if failures
+                                else None
+                            ),
+                        ]
+                        if x is not None
                     ]
-                    if x is not None
-                ]
-            ),
-            info_box=True,
-        )
+                ),
+                info_box=True,
+            )
+
+        num_lines = _print_infobox()
 
         # results menu
         line_fmt = "{:" + str(chars_per_line - 3) + "s} {:>2s}"
@@ -201,6 +204,7 @@ class DeviceTests(Page):
                 for x in self.results
             ],
             offset=(num_lines + 1) * FONT_HEIGHT,
+            infobox_callback=_print_infobox,
         )
         idx, _ = results_menu.run_loop()
         if idx == len(self.results):
--- a/src/krux/pages/home_pages/bip85.py
+++ b/src/krux/pages/home_pages/bip85.py
@@ -159,16 +159,19 @@ class Bip85(Page):
                     ),
                 ),
             ]
-            self.ctx.display.clear()
-            info_len = self.ctx.display.draw_hcentered_text(
-                info, info_box=True, highlight_prefix=":"
-            )
-            info_len *= FONT_HEIGHT
-            info_len += DEFAULT_PADDING
+
+            def _print_infobox():
+                self.ctx.display.clear()
+                return self.ctx.display.draw_hcentered_text(
+                    info, info_box=True, highlight_prefix=":"
+                )
+
+            info_len = _print_infobox()
             submenu = Menu(
                 self.ctx,
                 menu_items,
-                offset=info_len,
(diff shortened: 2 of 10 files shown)


tadeubas:
Hey @kdmukai, thanks for testing this PR so quickly!
Yeah, asking you for a full code review would be a bit much. You already helped a ton by testing on three devices (two touch models and one with battery + button only). Really appreciate it! :heart: 

> One oddity: My initial rebuild and flash of this PR on the wonder_k just made it freeze.
> … I didn’t notice any errors (but wasn’t paying close attention) …

I think I know exactly what happened. If you build with a typo like:

```
./krux build maixpy_wonderk
```

it just outputs, **but nothing is built**:

```
Wrong or unsupported device name, use maixpy_devicename
```

If you then run (without a typo :stuck_out_tongue_winking_eye: ):

```
./krux flash maixpy_wonder_k
```

It will _flash_ whatever was previously built. So if your last successful _build_ was for **TZT**, for example, that firmware ends up being _flashed_ onto the **Wonder K**, and of course it won’t behave correctly 😅

On the next _build_ attempt (using the correct device name), the files get overwritten properly. Flashing after that installs the correct firmware, which I think is why everything worked afterward.

odudex:
I think this is not the way to go for a more consistent Screensaver https://github.com/selfcustody/krux/issues/760. 
IMO it should be called by system timer, just like "auto shutdown", and its code should be self contained in a single file, not spread across multiple files.
We should be able to disable it in a few situations, like QR scanning and animated QR codes, and re-enable when finished.

src/krux/pages/home_pages/bip85.py
```diff
@@ -159,16 +159,19 @@ def _derive_base64_password(self):
...
-                info, info_box=True, highlight_prefix=":"
-            )
-            info_len *= FONT_HEIGHT
-            info_len += DEFAULT_PADDING
+
+            def _print_infobox():
+                self.ctx.display.clear()
+                return self.ctx.display.draw_hcentered_text(
+                    info, info_box=True, highlight_prefix=":"
+                )
```
qlrd, line 167: Why not define it as a private method in the class instead a anon function?

src/krux/pages/home_pages/pub_key_view.py
```diff
@@ -78,19 +78,28 @@ def _pub_key_text(version):
...
+                    info_text,
+                    offset_y=FONT_HEIGHT,
+                    info_box=True,
+                )
+                self.ctx.display.draw_hcentered_text(
+                    self.ctx.wallet.key.fingerprint_hex_str(pretty=True),
+                    offset_y=FONT_HEIGHT,
+                    color=theme.highlight_color,
+                    bg_color=theme.info_bg_color,
+                )
```
qlrd, line 94: Same here

src/krux/pages/new_mnemonic/dice_rolls.py
```diff
@@ -308,25 +308,31 @@ def delete_roll(buffer):
...
+                self.ctx.display.clear()
+                n_lines = self.ctx.display.draw_hcentered_text(
+                    rolls_str, info_box=True, max_lines=max_lines
+                )
+                self.ctx.display.draw_hcentered_text(
+                    t("Rolls:"),
+                    color=theme.highlight_color,
+                    bg_color=theme.info_bg_color,
+                )
+                return n_lines
```
qlrd, line 324: Same here

src/krux/pages/device_tests.py
```diff
@@ -159,28 +159,31 @@ def test_suite(self, interactive=False):
...
+                        if x is not None
                     ]
-                    if x is not None
-                ]
-            ),
-            info_box=True,
-        )
+                ),
+                info_box=True,
+            )
```
qlrd, line 184: Same here

src/krux/pages/login.py
```diff
@@ -287,34 +288,40 @@ def _load_key_from_words(self, words, charset=LETTERS, new=False):
...
+
+                # draw network with highlight color
+                self.ctx.display.draw_hcentered_text(
+                    network_name,
+                    DEFAULT_PADDING + FONT_HEIGHT,
+                    color=Utils.get_network_color(network_name),
+                    bg_color=theme.info_bg_color,
+                )
+
+                return num_lines
```
qlrd, line 312: Same here.

src/krux/pages/qr_view.py
```diff
@@ -460,10 +460,15 @@ def save_qr_image_menu(self):
...
-        self.ctx.display.clear()
-        self.ctx.display.draw_hcentered_text(
-            t("Res. - Format"), FONT_HEIGHT, info_box=True
-        )
+
+        def _print_infobox():
+            self.ctx.display.clear()
+            self.ctx.display.draw_hcentered_text(
+                t("Res. - Format"), FONT_HEIGHT, info_box=True
+            )
```
qlrd, line 468: Same Here

src/krux/pages/wallet_settings.py
```diff
@@ -195,9 +195,24 @@ def customize_wallet(self, key):
...
+                )
+
+                # draw network with highlight color
+                self.ctx.display.draw_hcentered_text(
+                    network_name,
+                    color=Utils.get_network_color(network_name),
+                    bg_color=theme.info_bg_color,
+                )
+
+                return n_lines
```
qlrd, line 213: Same here

src/krux/pages/home_pages/bip85.py
tadeubas: I defined it inside the function because it’s an implementation detail, not a reusable behavior of the class. Its scope is explicit: it cannot be called from anywhere else.

PR #781, opened 2025-11-24
Refactor capture entropy test to be less fragile, more DRY

kdmukai:
### What is this PR for?

Refactor to make the `test_capture_entropy.py` tests less fragile. Original tests were reconstructing the exact `draw_centered_text` args to test against, creating an unnecessary maintenance burden. This revised approach simply looks for the presence of "Insufficient" in any of its calls.

Revision ends up being more DRY and simplifies the tests a bit.

Minor fixes to two comments that were incorrectly saying the test was for "Insufficient" when it was actually the success case.

Minor fixes to remove redundant `CameraEntropy` instantiation.

---

### Changes made to:
- [x] Tests

- [x] N/A

- [x] Other: test improvement

### DIFF
--- a/tests/pages/test_capture_entropy.py
+++ b/tests/pages/test_capture_entropy.py
@@ -5,13 +5,16 @@ import random
 
 def _called_with_insufficient_entropy_message(ctx):
     """Helper bool to determine if ctx.display.draw_centered_text was called with a message including "Insufficient" """
-    return len(
-        [
-            call
-            for call in ctx.display.draw_centered_text.call_args_list
-            if "Insufficient" in call.args[0]
-        ]
-    ) > 0
+    return (
+        len(
+            [
+                call
+                for call in ctx.display.draw_centered_text.call_args_list
+                if "Insufficient" in call.args[0]
+            ]
+        )
+        > 0
+    )
 
 
 def test_cancel_capture(amigo, mocker):


tadeubas:
Hey @kdmukai, I checked the changes, but I’m still not sure how these tests would catch a bug if the code behavior changes. It feels like the tests might adapt to any modification, which would make them less reliable at actually detecting regressions. Could you clarify how they ensure real test coverage?

kdmukai:
> As long as the message contains the word “Insufficient”, the test can confirm that the behavior is correct. We don’t actually need to match the full text shown on the screen. Is that right?

Yes, that's a more concise way to say it!

tadeubas:
Oh, I see now, we already have tests for `entropy_measurement_update`, so those would catch (or should catch) any incorrect changes to the logic.

In that case, the tests you modified were only verifying that the mocked messages were printed to the screen. They weren’t actually validating the numeric values, just making sure the UI displayed the info. So no real loss there (or maybe we should also check for the word _bits_ and or _bits/px_ ?).

kdmukai:
> the tests you modified were only verifying that the mocked messages were printed to the screen. They weren’t actually validating the numeric values, just making sure the UI displayed the info. So no real loss there 

Right.

> (or maybe we should also check for the word _bits_ and or _bits/px_ ?).

I view presentation issues / changes / choices as being mostly outside the domain of a test suite. Perhaps some day someone argues, for example, that the bits/px readout is nonsense and so it's dropped from the UI. Sure. I wouldn't expect that presentation change to have any bearing on the test suite. Or conversely if there's some additional metric that's useful to display, adding it shouldn't break the test suite (though certainly some new tests would be desirable to cover the new metric).

Whereas testing dynamic UI states (show/hide the "Insufficient" message) is obviously a necessary part of the test suite.

PR #787, opened 2025-11-27
ci: add convetional commits to krux

qlrd:
### What is this PR for?

Convetional commits are a specification for adding human and machine readable meaning to commit messages. This PR is a proposal for check if the follow-up commits agree with those specified in the https://www.conventionalcommits.org/en/v1.0.0/. While this changes some of the current workflow for commits, this brings more organization and a development standard (the semmantic git commit message) to a project that had a lot of attention in the last months.

### Changes made to:
- [ ] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG
- [x] CI

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/.github/workflows/conventional-commits.yml
+++ b/.github/workflows/conventional-commits.yml
@@ -1,0 +1,13 @@
+name: Conventional Commits
+
+on:
+  pull_request: # We only need to check commit messages in PRs
+
+jobs:
+  conventional-commits:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: webiny/action-conventional-commits@v1.3.0
+        with:
+          allowed-commit-types: "feat,fix,docs,style,refactor,test,i18n,ci,chore,git"


qlrd:
> I think this is more about a self awareness and discipline.

In aspect, yes. It help me to organize myself. In other aspect, Krux is a project that have a solid and stable idea in users minds. But i think it's not same for new developers that we started to receive. And in time we can have more and more developers. The Conventional Commits are a standard in many Bitcoin FOSS.

IMHO, the most direct benefits are scope knowledge management from outside developers, so they can choose which aspect of the code they could work, as well enforce some clarity to this or that commit, helping in review process (giving an idea of what will be changing before read the actual code -- i think the imagination given by semantic git commits is a powerful mental tool).

Also, I believe it's time to us start in thinking in a `CONTRIBUTING.md`.

> So, do we gain anything measurable from enforcing this convention? I don't know.

I know there are some ways to use this "measurability" with gh-actions with some tools. I'm not an expert, but can research on that.

odudex:
Currently I use squash commits to keep history clean, which has been working well for our scale.
Before adding enforcement tooling, I'd like to understand the concrete problems we're experiencing. Are commit messages currently causing confusion? Would automated changelog generation provide real value?
Let's monitor merged commits for the next 2-4 weeks and evaluate: 
- How often unclear commits cause issues
- Whether our current squash workflow addresses the same concerns.

If we identify specific pain points that conventional commits would solve, we can revisit.
My preference is toward minimalism, fewer dependencies, less tooling, less maintenance overhead. But I'm open to structure where it demonstrably helps the project.

One thing I will change is taking extra care with PRs. I won't merge those with excessively wide scope. Oversized PRs inevitably lead to less precise commit messages, weaker correlation between messages and actual code changes, and harder bug tracking. There's no automated tool for this, it requires deliberate review discipline.

qlrd:
> Currently I use squash commits to keep history clean, which has been working well for our scale. Before adding enforcement tooling, I'd like to understand the concrete problems we're experiencing. Are commit messages currently causing confusion? Would automated changelog generation provide real value? Let's monitor merged commits for the next 2-4 weeks and evaluate:
> 
>     * How often unclear commits cause issues
> 
>     * Whether our current squash workflow addresses the same concerns.
> 
> 
> If we identify specific pain points that conventional commits would solve, we can revisit. My preference is toward minimalism, fewer dependencies, less tooling, less maintenance overhead. But I'm open to structure where it demonstrably helps the project.
> 
> One thing I will change is taking extra care with PRs. I won't merge those with excessively wide scope. Oversized PRs inevitably lead to less precise commit messages, weaker correlation between messages and actual code changes, and harder bug tracking. There's no automated tool for this, it requires deliberate review discipline.

Yeah, I get where you’re coming from about keeping things minimal and avoiding big PRs — totally fair. But just to add my perspective:

Sometimes a change is naturally wide in scope, and trying to slice it into many tiny PRs actually makes things harder to review because the full context gets scattered. In a lot of Bitcoin projects, big PRs aren’t a “style choice,” it’s just that certain features or refactors touch multiple areas at once, and it’s easier to understand them when they’re shown together.

I’m not advocating for chaotic monster PRs 😅 — just that when a feature or refactor is really one coherent thing, it can be cleaner to review as a single larger PR than as 8 micro-PRs that only make sense when combined.

Conventional commits (or any structure on top) help mostly with keeping history readable and tracking why something changed, regardless of PR size. If after a few weeks it doesn’t bring real value, sure, we can drop it. No problem.

So yeah — not pushing for “always big PRs,” just suggesting we don’t rule out wide-scope ones when the change is inherently cross-cutting. It can actually make life easier (at least for me), not harder.

PR #793, opened 2025-12-01
Small change on Touch Threshold, Buttons Debounce and Swipe Threshold

tadeubas:
#### Settings / Constants
- Increased range for **Touch Threshold** and **Buttons Debounce**
- Reduced **Buttons Debounce** default value a little (customized default for _M5StickV_)
- Slightly reduced the Swipe Threshold constant

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on m5stickv

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


src/krux/krux_settings.py
```diff
@@ -314,7 +315,7 @@ class TouchSettings(SettingsNamespace):
 
     namespace = "settings.touchscreen"
     default_th = 40 if kboard.is_wonder_k else 22
-    threshold = NumberSetting(int, "threshold", default_th, [10, 200])
+    threshold = NumberSetting(int, "threshold", default_th, [2, 200])
```
odudex, line 318: Any real situation where sensitivity had to be set so aggressively (value=2) ?

tadeubas: I can't test against all possible device lots, so I'd rather leave this flexible. In some cases, like when the user is wearing gloves, higher sensitivity can be helpful

odudex: If it didn't brick any device (like being always pressed), it's fine.

tadeubas: If the user changes this value, the device immediately applies it before exiting the settings menu, so they will notice right away and can simply reset the device to revert to the previous value. I tested with a setting of 2 and didn't manage to brick any of my touch devices (couldn't test on the Amigo due to its existing touch issues): TZT, Yahboom, WonderMV, Embed Fire, WonderK Pro. 

On the WonderK Pro, a little below 10 already feels good, and the Embed Fire was performing so well that it was already set to 2. Other devices had poorer results, but that's exactly why the setting exists: anyone can adjust it according to the behavior of their specific device, rather than us enforcing or constraining the value.

odudex: Nice, thank you for testing.

PR #791, opened 2025-11-30
Scan QR codes inverted (black bg)

tadeubas:
### What is this PR for?
Added support for inverted QR codes, enabling scans of QRs with a black background

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [x] CHANGELOG

### Did you build the code and tested on device?
- [x] Yes, build and tested on TZT

### What is the purpose of this pull request?
- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other

PR #795, opened 2025-12-01
Inverted Color QR Code Decoding

odudex:
### What is this PR for?
Alternative to #791 

This PR adds inverted color QR code detection while improving overall decoder performance.
To avoid adding lag with the new feature, I first optimized the existing pipeline:
- Utilized dual-core processing for QR decoding where possible
- Streamlined the image processing path

These optimizations reduced decoding time from ~33ms to ~11ms.
Adding inverted color detection increased processing by ~7ms, resulting in a final average of ~18ms - still a net improvement of ~15ms over the original implementation.

### Changes made to:
- [X] Code
- [X] Tests
- [ ] Docs
- [X] CHANGELOG

- [ ] Yes, build and tested on Amigo

- [ ] Bug fix
- [X] New feature
- [ ] Docs update
- [ ] Other


odudex:
Maybe the small OLED model, but I don't see any user doing that. If someday a coordinator uses it, or any user demands it for any reason, we can add more complex logic that works with inverted color animated QRs

PR #802, opened 2025-12-05
Loading a Descriptor Sets Wallet Attributes

odudex:
### Problem
Krux currently requires users to set wallet attributes before loading a key. 
While this enables conveniences and sanity checks, it creates "error walls" 
that force users to configure attributes before proceeding.

### Solution
This PR improves the user experience by:
- Auto-detecting wallet attributes from the descriptor being loaded
- Prompting users to update wallet settings dynamically during the loading process
- Removing support for loading xpubs/ypubs/zpubs as if they were descriptors

[Screencast from 2025-12-05 14-19-32.webm](https://github.com/user-attachments/assets/65307642-837d-46d7-8fef-cd7a33b851e4)

### Feature Removal: xpubs as descriptors

This PR removes the ability to load xpubs, ypubs, and zpubs as if they were descriptors.

**Rationale:**
- Treating extended public keys as descriptors could require making assumptions about derivation paths and script types
- These assumptions increased code complexity and introduced potential risks during automatic wallet attribute detection
- Modern descriptor format provides explicit, unambiguous wallet configuration

### Changes made to:
- [X] Code
- [X] Tests
- [x] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on Amigo

- [ ] Bug fix
- [X] New feature
- [ ] Docs update
- [ ] Other


src/krux/key.py
```diff
@@ -345,3 +352,15 @@ def get_final_word_candidates(words):
...
+
+def get_policy_type_name(policy_type):
+    """Returns human-readable name for a policy type constant"""
+    return POLICY_TYPE_NAMES_MAP.get(policy_type, "Unknown")
+
+
+def get_network_name(network):
+    """Returns human-readable name for a network object (Mainnet or Testnet)"""
+    if not network:
+        return None
```
qlrd, line 365: It's possible that the `network` would be `""` or `None`?

odudex: I think it's a defensive safeguard is worth keeping, as `_determine_descriptor_network` can return `None`

PR #805, opened 2025-12-11
Add new battery percentage indicator

tadeubas:
### What is this PR for?
- Adds a new battery percentage indicator that updates approximately every 5 seconds

**Note**: In rare cases, such as when the user briefly connects and disconnects the charger with a low battery, the displayed value may be inaccurate for about a minute. Most of the time, the feature works as expected.

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on Amigo, Cube, M5StickV

- [ ] Bug fix
- [x] New feature
- [ ] Docs update
- [ ] Other


qlrd (commented):
LGTM, i will test here

odudex:
Show battery charge as percentage could be an option if a precise measurement method, such as Coulomb counter, was used.

PR #485, opened 2024-11-30
Krux Apps, small utility programs to enhance Krux

tadeubas:
### What is this PR for?
Create a new Tool to enable the execution of developer's signed Krux apps (Kapps) as `.mpy` files in the SD card.

To test on the simulator, copy the `.mpy` file and its corresponding `.mpy.sig` into the `simulator/sd` folder. When executed, both files are automatically copied to `simulator/flash`. The simulator then emulates the execution of the selected Kapp by running the corresponding `.py` file from the simulator’s root directory. To avoid duplication, it’s recommended to create symbolic links to these `.py` files instead of copying them.

In any case you will need to generate a `.mpy` file using the tool `firmware/MaixPy/components/micropython/core/mpy-cross`. Follow its [`README.md`](https://github.com/selfcustody/micropython/tree/master/mpy-cross) instructions in order to make and then compile a `.py` file into `.mpy`. It is implied that all Kapps must implement the `run()` function to be executed.

You will also need to create a `.sig` file based on the `.mpy`. Remember to change the contents of the `SIGNER_PUBKEY` inside `src\krux\metadata.py` following instructions here [To generate a keypair and Signing the firmware (firmware will be your .mpy file)](https://selfcustody.github.io/krux/getting-started/installing/from-source/) in order to sign and verify the signature within Krux code (Ex: `./krux sign kapps/nostr.mpy privkey.pem`).

Use this nice tool from @kdmukai to test Nostr Kapp: https://nostrtool.com/

It’s also implied (though we can change this later) that **Kapps won’t be translated** - existing firmware strings will be, but new ones won’t. I also think we need to distribute the `.mpy` files (and their signatures) with each release.

It is possible to keep more than one Kapp in flash memory. There is a security setting to enable the startup execution of a Kapp (before Krux firmware for OPSec), only Kapps that have the const `ALLOW_STARTUP = True` are listed there.

#### Other changes
- **New Kapp - k_QR**: QR scanner for inconspicuous startup (OPSec) [scan QR with text: `krux`]
- **New Kapp - Steganography**: Capture photos as bmp, view them and hide and revel data in the pixels of the img
- **Krux script**: signing now produces only compact 70-byte signatures, preventing the 71- or 72-byte variants that embit would flag as invalid
- **pyproject.toml**: New tasks to handle mpy generation (`poetry run poe mpy-all`)
- **Simulator**: flash folder now simulates device's user flash space
- **firmaware.py**: Refactored code to allow reuse of `get_pubkey` and `check_signature`
[...]

PR #811, opened 2026-01-04
Add QR scanning to text keypads

oroderico:
### What is this PR for?

This PR adds an optional QR scan action directly to text keypads.

The scan action is placed on the last special-character page, preserving the existing keypad grid and key sizes across device profiles. It provides one consistent input mechanism for:

- Tamper Check Codes
- BIP39 passphrases
- Encryption keys

This replaces the separate Type/Scan selection screens and removes the need for a dedicated TC input-mode setting.

https://github.com/user-attachments/assets/400966c8-d723-4d85-999b-31a747a140ff

### Input flow

- Pressing the QR glyph opens the scanner.
- If the buffer already contains text, Krux asks before replacing it.
- A successful scan returns the value to the editable keypad buffer for review.
- Cancelling or failing a scan preserves the previous buffer.
- The user can edit the scanned value before pressing Go.

Existing keypad callers remain unchanged unless they explicitly provide a scan callback.

### Tamper Check considerations

This does not change TC Code validation, startup behavior, or the Tamper Check threat model. QR input only reduces input friction for an already user-known value.

Users should avoid storing a TC Code QR next to the device, as doing so would undermine the purpose of the Tamper Check.

### Design collaboration

This implementation evolved through the discussion in this PR:

- @odudex proposed replacing the separate Type/Scan flow with a reusable QR action inside the keypad and helped define its placement and interaction behavior.
- @Naman015 designed and tested the keypad layout, interaction flow, and QR glyph across device profiles.
- @qlrd contributed review, Tamper Check context, and feedback on the final glyph direction.

Thank you all for shaping the implementation.

The four implementation commits credit Naman Gupta as coauthor for his direct design collaboration.

### Validation

- `uv run pytest --cache-clear tests/pages/test_keypads.py tests/pages/test_page.py tests/pages/test_encryption_ui.py tests/pages/test_tc_code_verification.py tests/pages/test_wallet_settings.py`
  - 109 passed
- `uv run poe lint`
  - 10.00/10
- `git diff --check`


tadeubas:
The Tamper Check (TC) Code is not intended to defend against attackers or protect confidential data. Its purpose is to make firmware tampering *obvious* and to prevent *casual bypass by normal users*, not to provide strong access control. The device is intentionally open to modification and scrutiny.

This PR **does not reduce attack friction**. It only reduces **user friction** when entering an already user-known value.

QR input also enables the use of high-entropy TC Codes that may be impractical to type or memorize. This allows users to choose longer, more complex strings without worrying about typing errors, while preserving the same security assumptions and behavior.

**I have no objections to this PR**; the comments above are only meant to ensure that the author and reviewers share a clear understanding of what is involved.

If the ability to choose between manual input and QR input is being considered as an OPSEC, I already have a related PR https://github.com/selfcustody/krux/pull/485 that ~~intentionally **does not** address inconspicuous startup, in order to keep the scope small and improve the chances of an earlier merge~~ has now a small QR code Kapp to address inconspicuous startup. ~~The broader idea there is to later introduce a game or similar mechanism to improve OPSEC.~~

oroderico:
> The Tamper Check (TC) Code is not intended to defend against attackers or protect confidential data. Its purpose is to make firmware tampering _obvious_ and to prevent _casual bypass by normal users_, not to provide strong access control. The device is intentionally open to modification and scrutiny.
> 
> This PR **does not reduce attack friction**. It only reduces **user friction** when entering an already user-known value.
> 
> QR input also enables the use of high-entropy TC Codes that may be impractical to type or memorize. This allows users to choose longer, more complex strings without worrying about typing errors, while preserving the same security assumptions and behavior.
> 
> **I have no objections to this PR**; the comments above are only meant to ensure that the author and reviewers share a clear understanding of what is involved.
> 
> If the ability to choose between manual input and QR input is being considered as an OPSEC, I already have a related PR #485 that intentionally **does not** address inconspicuous startup, in order to keep the scope small and improve the chances of an earlier merge. The broader idea there is to later introduce a game or similar mechanism to improve OPSEC.

@tadeubas 

Thanks for the clarification — fully aligned.

This PR intentionally avoids framing TC input as a security or OPSEC feature.
The goal is purely to reduce user friction and error rates when entering an already user-known value, especially on constrained input devices.

No changes were made to validation logic, threat model, or startup behavior, and QR input remains strictly optional.

docs/getting-started/settings.en.md
```diff
@@ -161,9 +161,12 @@ Set the time it takes for Krux to automatically shut down. This feature not only
...
 <img src="../../img/maixpy_amigo/settings-tc-boot-300.png" align="right" class="amigo">
 
-Chose if you would like to run [Tamper Check Flash Hash](features/tamper-detection.md) every time the device is powered on.
+Choose if you would like to run [Tamper Check Flash Hash](features/tamper-detection.md) every time the device is powered on.
 
-Activating *TC Flash Hash* at boot helps prevent unauthorized use by requiring the *TC Code*. But is important to note, unlike a PIN, the *TC Code* does not provide access control over USB. This means that the device's memory remains accessible via USB, allowing it to be flashed with firmware that does not require the *TC Code*.
+Activating *TC Flash Hash* at boot prompts for the *TC Code* before generating the
+boot flash hash. Unlike a PIN, the *TC Code* does not provide access control over
+USB. This means that the device's memory remains accessible via USB, allowing it
+to be flashed with firmware that does not require the *TC Code*.
```
qlrd: Cosmetic one just to keep less diff, non-blocker:

```suggestion
Activating *TC Flash Hash* at boot prompts for the *TC Code* before generating the boot flash hash. Unlike a PIN, the *TC Code* does not provide access control over USB. This means that the device's memory remains accessible via USB, allowing it to be flashed with firmware that does not require the *TC Code*.
```

docs/getting-started/settings.en.md
```diff
@@ -175,6 +178,21 @@ When `True`, Krux will disable the [New Mnemonic](./usage/generating-a-mnemonic.
 
 <div style="clear: both"></div>
 
+#### TC Input Mode
+
+Choose how Krux asks for the Tamper Check Code when running
+[Tamper Check Flash Hash](features/tamper-detection.md). `Manual` is the default
+and preserves the existing keypad flow. `Scan QR` reads the code from a QR code,
+reducing typing friction and allowing convenient use of longer user-known codes.
+`Ask Every Time` lets you choose between manual and QR input each time.
```
qlrd: Cosmetic one just to keep less diff, non-blocker:

```suggestion
Choose how Krux asks for the Tamper Check Code when running [Tamper Check Flash Hash](features/tamper-detection.md). `Manual` is the default and preserves the existing keypad flow. `Scan QR` reads the code from a QR code, reducing typing friction and allowing convenient use of longer user-known codes. `Ask Every Time` lets you choose between manual and QR input each time.
```

docs/getting-started/settings.en.md
```diff
@@ -175,6 +178,21 @@ When `True`, Krux will disable the [New Mnemonic](./usage/generating-a-mnemonic.
...
+
+QR input trims surrounding whitespace and trailing line endings. QR codes with
+empty, malformed, or multiline contents are not accepted. Creating, changing, or
+replacing the Tamper Check Code always requires manual input, regardless of this
+setting.
```
qlrd: Cosmetic one just to keep less diff, non-blocker:


```suggestion
QR input trims surrounding whitespace and trailing line endings. QR codes with empty, malformed, or multiline contents are not accepted. Creating, changing, or replacing the Tamper Check Code always requires manual input, regardless of this setting.
```

tests/pages/test_settings_page.py
```diff
@@ -621,6 +621,36 @@ def test_wrong_code_set_new_tc_code(amigo, mocker):
...
+    settings_page = SettingsPage(ctx)
+
+    assert settings_page.namespace(Settings().security)() == MENU_EXIT
+    assert captured_labels == [
+        "Shutdown Time",
+        "TC Flash Hash at Boot",
+        "Hide Mnemonics",
+        "TC Input Mode",
+        "Tamper Check Code",
+    ]
```
qlrd: This appear to be a unit test (mocking the Menu class -- not against to have them here). On `test/pages` i would recommend to do some functional ones with some clicking buttons procedures

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
+    b'<9y~\xba\xfdqv\xa0\xb5\x0f\xf9v6 lht\xf4\x17\xcfmw"J\xac8bkx\xc3\xfa'
+)
+INVALID_EXTENDED_HASH = (
+    b"\x98\x99kJ\x04\x98r\xec \x9d\xd6\xbaG\xc2P\xbb9\x00\xe53("
+    b"\x98\xb9\x1a,\x13-.\x1e\xe6Z\xc8"
+)
+
+
+def _patch_device_secret(mocker):
+    mocker.patch("machine.unique_id", return_value=b"\x01" * 32)
```
qlrd: Non-blockable, but it's nice to follow some idiomatic pytest things, or something like that (just thinking):

```suggestion
@pytest.fixture
def device_secret(mocker):
    mocker.patch("machine.unique_id", return_value=b"\x01" * 32)
```

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
     )
+
+
+def test_tc_code_input_mode_default_is_manual(amigo):
+    from krux.krux_settings import Settings, TAMPER_CHECK_INPUT_MANUAL
+
+    assert Settings().security.tamper_check_code_input_mode == TAMPER_CHECK_INPUT_MANUAL
+
+
+def test_tc_code_verification_manual_cases(amigo, mocker):
```
qlrd: If you choose to follow the commented above:


```suggestion
def test_tc_code_verification_manual_cases(amigo, mocker, device_secret):
```

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
+        tc_verifier._capture_from_qr.assert_not_called()
 
 
-def test_tc_code_verification_esc_key(amigo, mocker):
+def test_tc_code_verification_scan_qr_mode(amigo, mocker):
+    from krux.krux_settings import Settings, TAMPER_CHECK_INPUT_SCAN_QR
+    from krux.pages.qr_capture import QRCodeCapture
+    from krux.pages.tc_code_verification import TCCodeVerification
+
+    _patch_device_secret(mocker)
```
qlrd: same

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
-        tc_verifier.capture_from_keypad.assert_called_once_with(
-            keypad_label,
-            [NUM_SPECIAL_1, LETTERS, UPPERCASE_LETTERS, NUM_SPECIAL_2],
-        )
+        _assert_keypad_called_for_tc_code(tc_verifier, case[3])
+        tc_verifier._capture_from_qr.assert_not_called()
 
 
-def test_tc_code_verification_esc_key(amigo, mocker):
+def test_tc_code_verification_scan_qr_mode(amigo, mocker):
```
qlrd: same

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
+
+    qr_capture_loop.assert_called_once()
+    tc_verifier.capture_from_keypad.assert_not_called()
+
+
+def test_tc_code_verification_manual_empty_input_is_invalid_code(amigo, mocker):
+    from krux.krux_settings import Settings, TAMPER_CHECK_INPUT_MANUAL
+    from krux.pages.tc_code_verification import TCCodeVerification
+
+    _patch_device_secret(mocker)
```
qlrd: same

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
+    tc_verifier.capture_from_keypad = mocker.MagicMock()
+
+    with patch("builtins.open", mock_open(read_data=TC_CODE_EXTENDED_HASH)):
+        assert tc_verifier.capture() is True
+
+    qr_capture_loop.assert_called_once()
+    tc_verifier.capture_from_keypad.assert_not_called()
+
+
+def test_tc_code_verification_manual_empty_input_is_invalid_code(amigo, mocker):
```
qlrd: same

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
+
+    with patch("builtins.open", mock_open(read_data=TC_CODE_EXTENDED_HASH)):
+        assert tc_verifier.capture() is False
+
+    _assert_keypad_called_for_tc_code(tc_verifier)
+    tc_verifier._capture_from_qr.assert_not_called()
+    tc_verifier.flash_error.assert_called_once_with("Invalid Tamper Check Code")
+
+
+def test_tc_code_verification_ask_every_time_routes_manual(amigo, mocker):
```
qlrd: same

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
+    from krux.krux_settings import (
+        Settings,
+        TAMPER_CHECK_INPUT_ASK_EVERY_TIME,
+        TAMPER_CHECK_INPUT_MANUAL,
+    )
+    from krux.pages.tc_code_verification import Menu, TCCodeVerification
+
+    _patch_device_secret(mocker)
```
qlrd: same

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
+
+def test_tc_code_verification_ask_every_time_routes_scan_qr(amigo, mocker):
+    from krux.krux_settings import (
+        Settings,
+        TAMPER_CHECK_INPUT_ASK_EVERY_TIME,
+        TAMPER_CHECK_INPUT_SCAN_QR,
+    )
+    from krux.pages.tc_code_verification import Menu, TCCodeVerification
+
+    _patch_device_secret(mocker)
```
qlrd: same

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,283 @@
...
+    tc_verifier._capture_from_qr = mocker.MagicMock()
+
+    with patch("builtins.open", mock_open(read_data=TC_CODE_EXTENDED_HASH)):
+        assert tc_verifier.capture() is True
+
+    _assert_keypad_called_for_tc_code(tc_verifier)
+    tc_verifier._capture_from_qr.assert_not_called()
+
+
+def test_tc_code_verification_ask_every_time_routes_scan_qr(amigo, mocker):
```
qlrd: same

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog 26.05.0 - May 2025
 
+### Tamper Check Code Input Modes
+Tamper Check Code verification can now use Manual, Scan QR, or Ask Every Time
+input modes. Manual remains the default. QR input reduces typing friction and
+permits convenient use of longer user-known codes. Changing the Tamper Check
+Code still requires manual input.
+
```
qlrd: Just to be less differ, non-blockable:

```suggestion
### Tamper Check Code Input Modes
Tamper Check Code verification can now use Manual, Scan QR, or Ask Every Time input modes. Manual remains the default. QR input reduces typing friction and permits convenient use of longer user-known codes. Changing the Tamper Check Code still requires manual input.
```

qlrd (commented):
LGTM while i did some first-round on code-quality aspects (less diff as possible, lints, etc.), so some non-blockable nits:

CHANGELOG.md
oroderico: Applied manually in c717e842. Also added you as co-author, thanks!

qlrd: No need to add me co-author, i'm just doing a reviewer job. The merit is yours.

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,284 @@
...
+    tc_verifier._capture_from_qr.assert_not_called()
+
+
+def test_tc_code_verification_esc_key(amigo, mocker):
     from krux.input import BUTTON_ENTER, BUTTON_PAGE_PREV
+    from krux.pages.tc_code_verification import TCCodeVerification
 
-    # Simulate user pressing ESC key
     cases = [
-        # 1 - Simple ESC key after entering TC Code then pressing "Yes" in prompt
```
qlrd, line 86: optional: those numbers are somewhat a guide for coding tests, IMHO no need to remove them (i saw them at first when started too, but then made sense from jeff's and other contributors perspective on long term).

tests/pages/test_tc_code_verification.py
```diff
@@ -1,96 +1,284 @@
...
+
+    _assert_keypad_called_for_tc_code(tc_verifier, changing_tc_code=True)
+    tc_verifier._capture_from_qr.assert_not_called()
+
+
+def test_tc_code_verification_esc_key(amigo, mocker):
     from krux.input import BUTTON_ENTER, BUTTON_PAGE_PREV
+    from krux.pages.tc_code_verification import TCCodeVerification
 
-    # Simulate user pressing ESC key
```
qlrd, line 84: optional: IMHO no need to remove, it work like a document title. One important aspect of coding tests are the comments (we do not add docstrings for specific reasons but comments are valuable and help understand the context while coding).

qlrd (commented):
Second round :), optional nits

tests/pages/test_tc_code_verification.py
oroderico: Addressed in c01c6062. Thanks for the guidance!

qlrd:
> The Tamper Check (TC) Code is not intended to defend against attackers or protect confidential data. Its purpose is to make firmware tampering obvious and to prevent casual bypass by normal users, not to provide strong access control. The device is intentionally open to modification and scrutiny.

just linking clarifiction issue to better review (see https://github.com/selfcustody/krux/issues/814).

oroderico:
Thanks for linking it. I kept this PR scoped to input selection and preserved the current TC verification behavior. If maintainers prefer to address #814 here, I can adjust, but I think it is a separate behavior change.

odudex:
IMO Krux's biggest flaw is an overwhelming UI: long flows, too many options and settings. The "Type or Scan" choice screen is in the way of all passphrase and key inputs, and with this PR, also TC Code, even though the great majority of users just expect to type.

I think we can make a lemonade from this lemon. Proposal: add a new key to the keypad, with a "QR code glyph", that launches the scanner when pressed.

- Works for TC Code, passphrase and KEF keys, one consistent mechanism everywhere text input can accept scanned data.
- Removes the "Type/Scan" step from all related flows. The cost of the choice is paid only by scan users (finding one key), not by everyone on every use.
- Removes the need for an input-method setting.
- Discoverable in context: the glyph is visible exactly when scanning is relevant, no need to know a setting exists.

Implementation details to agree on:
- Placement: alongside the existing special keys (shift, delete, ABC/123 toggle), on the first keypad page. Burying it in the symbols page would defeat the discoverability. On the smallest screens, if real estate doesn't allow it, we can discuss alternatives (e.g. long-press), but visible-by-default should be the target.
- Mid-entry behavior: if the user already typed something, scanning should replace the input, with a confirmation prompt. Append is almost never what someone wants, silently replacing a partial passphrase isn't either.
- Cancel/failure path: aborting or failing a scan returns to the keypad with the previous input intact.
- TC Code note: scanning shouldn't weaken what the typed TC Code was designed for. Users who print their TC Code as a QR and store it next to the device are undermining the tamper check, worth a line in the docs, but I don't think it blocks the feature.

Your PR proved the demand for scanning these inputs, this proposal just gives that capability a permanent home in the keypad instead of an extra screen. Would you be up for evolving the PR in this direction?

I'd also like to bring @Naman015 into this discussion, as a design specialist, his help on the glyph itself and keypad layout per device class would be valuable here.

oroderico:
Yes, I am definitely open to evolving the PR in this direction.

The keypad-based approach makes sense to me. It avoids another setting, removes the extra Type/Scan choice screen, and keeps scan as an in-context action for users who need it.

This is a larger change than my original PR, so I may need some guidance while implementing it, especially around the keypad layout and QR glyph details.

I will wait for layout/glyph guidance, especially from @Naman015, before changing the keypad layout.

Naman015:
Hi @odudex , thanks for bringing me in. I tested the two layout options on the TZT and M5StickV simulators and annotated the results here with mockups:
https://www.figma.com/design/pqiyvCdIViDLbk4Hsy95SX/Krux-QR-Scan?node-id=18-3&t=gMjAd02iibj6lKXf-0

Both layouts fit TZT... on M5StickV, 5×7 keeps abc, qr, esc and go separate, while 6×6 merges the fixed labels visually. Since 5×7 also follows the current Keypad layout calculation, i recommend using it without adding a device-specific layout.

Could you have a look and confirm the 5x7 direction please?

What i am curious about is... after a successful scan, should the scanned value return to the editable keypad buffer for review or immediately behave as if Go was pressed? My preference is review first so user can check the value before accepting it.

Intended flow in my opinion should be

QR glyph pressed → confirm replacement if needed → scan → review/Go on success OR restore original buffer on cancel/failure

Once those are confirmed, I’ll work on the final scan glyph at native device size and coordinate with @oroderico on the keypad/layout implementation.

odudex:
I like proposition A over B, but maybe we need a C:
A currently creates an extra line on the grid, reducing all keys surface area, which are already very small in the 2" devices. So we need to evaluate the idea of placing it together with special characters keypads. I know it's "farther" to reach it. But we must remember, this is the unusual input method.
Regarding the flow, yes, UI should allow edition of the scanned content (some users like to change one letter or append something, as an extra secret layer)

odudex:
Reaching especial characters is easy in touch devices, just a swipe right and you're there. For M5stickV and Cube, you have to press "prev" until it reaches the keyboard toggle a few times. But we also must remember, both devices are discontinued, and there are only a few users out there, against maybe a few thousands with 2" touch devices(sorry @oroderico :grimacing:)

Naman015:
Hi @odudex, so i tested option C in different simulators. Placing QR in an unused cell on the last special character page preserves the existing 5×6 grid and key sizes on devices. A right swipe also reaches it on touch devices.

It also remains reachable through the keypad toggle on M5StickV and Cube with no clipping or focus navigation issues.

I understood “swipe right and you’re there” to mean the scan action should appear only on the second special-character page. is that ok or would you prefer it on both symbol pages? (I prefer just the last special character page) once this is confirmed i will update the existing mockups with option C and show the scanned content returning to the editable keypad before moving further.

Naman015:
> But we also must remember, both devices are discontinued, and there are only a few users out there, against maybe a few thousands with 2" touch devices

On a side note, do we have a rough sense of which Krux devices are most commonly used today and which ones should be treated as legacy compatibility targets? I couldn’t find any public device-usage figures. Even a rough estimate would help me weigh design trade-offs and simulator coverage.

odudex:
It's not possible to know the amount of devices and models out there. But we do know Sipeed stopped producing Cube, and M5stack stopped producing M5stickV years ago. And we also do know Yahboom sold a lot because it is the cheaper, and TZT, WonderMV share the same screen size and resolution. Embed Fire, WonderK, although larger, also share same resolution and touch UI, so touch 240x320 is, by far, the most usual UI.

qlrd:
> Thanks for linking it. I kept this PR scoped to input selection and preserved the current TC verification behavior. If maintainers prefer to address #814 here, I can adjust, but I think it is a separate behavior change.

If possible, i advocate to put a secondary message on commit message like `Fix #814` if you choose to link it (from what i understood not, right?) and some rational about (so it could be documented on PR history, just a resume i think). This could be used by github to close automatically some issues (idk if it's enabled) and other git tools. But is optional :)

tests/pages/test_tc_code_verification.py
```diff
@@ -1,85 +1,274 @@
...
+)
+INVALID_EXTENDED_HASH = (
+    b"\x98\x99kJ\x04\x98r\xec \x9d\xd6\xbaG\xc2P\xbb9\x00\xe53("
+    b"\x98\xb9\x1a,\x13-.\x1e\xe6Z\xc8"
+)
+
+
+@pytest.fixture
+def device_secret(mocker):
+    mocker.patch("machine.unique_id", return_value=b"\x01" * 32)
```
qlrd, line 8: One thing that came to mind is that i wans't understanding why `mocker`ing it inside a loop with the same result, but then thought that doing it many times could be desirable if this was the initial intention, so if you need those initial intentions, the secrets can be wrapped on a function inside. i think this could be more re-usable.

```python
@pytest.fixture
def device_secret(mocker):
    def _wrap(value):
        mocker.patch("machine.unique_id", return_value=b"<the value>" * 32)
    return _wrap
```

And then call it as function on test something like:

```python
def test_test(amigo, mocker, device_secret):
    # ...
    for i, case in cases:
        device_secret(i)
```

qlrd (commented):
Since you already addresed it, it's just a nice to have one:

Naman015:
> A currently creates an extra line on the grid, reducing all keys surface area, which are already very small in the 2" devices. So we need to evaluate the idea of placing it together with special characters keypads. I know it's "farther" to reach it. But we must remember, this is the unusual input method.

Hi, so I updated the option C mockup and added three options for qr glyph and designed the preferred one at the 14 px, 16 px and 24 px sizes used across the device profiles here: https://www.figma.com/design/pqiyvCdIViDLbk4Hsy95SX/Untitled?node-id=94-3

I recommend the simple 11×11 pixel matrix... it remains clear at 14 px on M5StickV and Cube, scales cleanly to 16 px and 24 px and also doesn’t need scan corners or a QR label. Keeping it on last special character page also preserves the existing 5×6 grid and key sizes.

@odudex, does this match what you had in mind? 

C editable scanned buffer capture

odudex:
> @odudex, does this match what you had in mind?

Yes, this is what I though. Would just add a few blocks at top center, and re-arrange blocks in a way that, even if in unconscious, will better resemble a real QR code:
image

Naman015:
That makes sense, adjusted the top centre and blocks using this reference while keeping the glyph symbolic and readable at 14px in this figma frame: https://www.figma.com/design/pqiyvCdIViDLbk4Hsy95SX/Untitled?node-id=115-2

odudex:
> That makes sense, adjusted the top centre and blocks using this reference while keeping the glyph symbolic and readable at 14px in this figma frame: https://www.figma.com/design/pqiyvCdIViDLbk4Hsy95SX/Untitled?node-id=115-2

Glyph A (no borders) looks good to me in all devices!

qlrd:
ACK Glyph A too

oroderico:
Hey @Naman015, nice work on this. Since we’re both contributing around the same flow, it would be good to stay in touch and coordinate ideas. Do you have a preferred contact outside GitHub?

Naman015:
> Hey @Naman015, nice work on this. Since we’re both contributing around the same flow, it would be good to stay in touch and coordinate ideas. Do you have a preferred contact outside GitHub?

Thank you! Yes i would be happy to collaborate, preferably on Telegram? If it is okay with you. I go by @Dexxtrose on there.

oroderico:
I’m working on the keypad-based implementation in a separate branch while coordinating with @Naman015. It’s still a WIP, so I haven’t updated this PR yet, but the current code is available here for anyone interested:
https://github.com/oroderico/krux/tree/feat/keypad-qr-scan

Once the design review is complete, I’ll bring the changes into this PR.

oroderico:
The keypad-based implementation discussed above is now available in this PR.

It replaces the original TC-specific input-mode approach with a reusable QR action inside text keypads. The implementation follows the agreed flow:

- QR action on the last special-character page
- confirmation before replacing existing input
- scanned content returned to the editable buffer
- previous input preserved on cancellation or failure
- no changes to existing keypad layouts or key sizes

It is currently used for TC Codes, BIP39 passphrases, and encryption keys.

Thank you @odudex for proposing and guiding the keypad-based direction, @Naman015 for the layout, glyph, and interaction design work, and @qlrd for the review, Tamper Check context, and glyph feedback.

The four implementation commits include Naman Gupta as coauthor.

I converted the PR to draft while the updated implementation is reviewed. @Naman015, @odudex, and @qlrd, the latest version is ready for your feedback.

oroderico:
Fixed a scanned-KEF decline path that could re-prompt for decryption; added regression coverage.

Commit: 4e2975d8

docs/getting-started/features/encryption/encryption.en.md
```diff
@@ -8,7 +8,14 @@ During encryption, users may be prompted to override their preferences or select
...
 
-When Krux detects data resembling a KEF-encrypted envelope, it prompts the user to *"Decrypt?"*, showing the KEF version, envelope ID (or label), and the PBKDF2 iteration count used during creation. To decrypt, the same key (typed or scanned) must be provided. Once unlocked, Krux uses the plaintext within context. If decryption is declined, the raw envelope is used instead—usually resulting in an error, since KEF data is meaningless without decryption.
+When Krux detects data resembling a KEF-encrypted envelope, it prompts the user
+to *"Decrypt?"*, showing the KEF version, envelope ID (or label), and the PBKDF2
+iteration count used during creation. To decrypt, provide the same key by typing
+it or selecting the QR glyph on the last symbols keypad. Scanned text returns to
+the keypad for review and editing; binary keys remain binary and proceed to
+confirmation. Once unlocked, Krux uses the plaintext within context. If
+decryption is declined, the raw envelope is used instead—usually resulting in an
+error, since KEF data is meaningless without decryption.
```
qlrd: Back again. Just to be less diff:

```suggestion
When Krux detects data resembling a KEF-encrypted envelope, it prompts the user to *"Decrypt?"*, showing the KEF version, envelope ID (or label), and the PBKDF2 iteration count used during creation. To decrypt, provide the same key by typing it or selecting the QR glyph on the last symbols keypad. Scanned text returns to the keypad for review and editing; binary keys remain binary and proceed to confirmation. Once unlocked, Krux uses the plaintext within context. If decryption is declined, the raw envelope is used instead—usually resulting in an error, since KEF data is meaningless without decryption.
```

docs/getting-started/features/encryption/encryption.en.md
```diff
@@ -95,4 +102,3 @@ While all KEF envelopes share the above format, each version differs - offering
 Storage of encrypted secrets on the device or SD cards are meant for convenience only and should not be considered a long-term form of backup. Always make a physical backup of your keys that is independent from electronic devices and test recovering your wallet from this backup before you send funds to it. Flash storage can degrade over time and may be subject to permanent damage, resulting in the loss of stored information.
 
 Remember that any encrypted secret is protected by the key you defined to encrypt it. If the defined [key is weak](https://www.hivesystems.com/blog/are-your-passwords-in-the-green), your encrypted mnemonic and other secrets will **not be protected**. If you have stored sensitive secrets in the device's internal flash memory using a [weak key](https://www.hivesystems.com/blog/are-your-passwords-in-the-green), the best way to undo this is to [erase user's data](../tools.md/#erase-users-data).
-
```
qlrd, line 98: This remotion is needed?

docs/getting-started/features/tamper-detection.en.md
```diff
@@ -38,6 +38,11 @@ When you enable the *TC Flash Hash at Boot* feature, the device will require you
 ## Tamper Check Code (TC Code)
 Before being stored in the device’s flash, the *TC Code* is hashed together with the K210 chip’s unique ID and stretched using PBKDF2. This ensures the *TC Code* is not retrievable via a flash dump and can only be brute-forced outside the device if the attacker also has access to the device’s unique ID (UID). By allowing letters, special characters, and running 100k iterations of PBKDF2, brute-forcing the *TC Code* from dumped data becomes more time-consuming and resource-intensive.
 
+Creating or changing the *TC Code* always requires manual input. When running
+*TC Flash Hash*, you can instead select the QR glyph on the last symbols keypad,
+scan the code, and review or edit it before pressing `Go`. Do not store a QR
+containing the *TC Code* with the device, as this undermines the tamper check.
+
```
qlrd, line 47: Just to be less diff:

```suggestion
Creating or changing the *TC Code* always requires manual input. When running *TC Flash Hash*, you can instead select the QR glyph on the last symbols keypad, scan the code, and review or edit it before pressing `Go`. Do not store a QR containing the *TC Code* with the device, as this undermines the tamper check.

```

docs/getting-started/usage/loading-a-mnemonic.en.md
```diff
@@ -158,10 +158,13 @@ Informs if the wallet has a passphrase. Adding or changing the passphrase result
...
-<img src="../../../img/maixpy_m5stickv/passphrase-load-options-250.png" align="right" class="m5stickv">
-<img src="../../../img/maixpy_amigo/passphrase-load-options-300.png" align="right" class="amigo">
 
-You can type or scan a BIP39 passphrase. When typing, swipe left :material-gesture-swipe-left: or right :material-gesture-swipe-right: to change keypads if your device has a touchscreen. Hold `PAGE` or `PREVIOUS` to move quickly through letters, and hold `ENTER` to jump to the next keypad.
+You can type or scan a BIP39 passphrase from the same keypad. Swipe right
+:material-gesture-swipe-right: to reach the last symbols keypad on touchscreen
+devices, or use the keypad toggle on button devices, then select the QR glyph.
+After a successful scan, the passphrase returns to the keypad so you can review
+or edit it before pressing `Go`. If text is already present, Krux asks before
+replacing it; cancelling or failing a scan preserves the existing text.
```
qlrd, line 170:
```suggestion
You can type or scan a BIP39 passphrase from the same keypad. Swipe right :material-gesture-swipe-right: to reach the last symbols keypad on touchscreen devices, or use the keypad toggle on button devices, then select the QR glyph. After a successful scan, the passphrase returns to the keypad so you can review or edit it before pressing `Go`. If text is already present, Krux asks before replacing it; cancelling or failing a scan preserves the existing text.
```

docs/getting-started/usage/navigating-the-main-menu.en.md
```diff
@@ -55,7 +55,12 @@ Please refer to [Transcribing QR Codes](../features/QR-transcript-tools.md) for
 
 This feature allows you to back up your mnemonic by encrypting it and storing it on the device's flash memory, on an SD card, or in QR code format. You can customize the encryption method and parameters in the [settings](../settings.md/#encryption).
 
-When using any of the encryption methods, you will be prompted to enter an encryption key. This key can be provided in text or QR code format. Additionally, you have the option to set a custom ID for easier management of your mnemonics. If a custom ID is not specified, the current loaded wallet fingerprint will be used.
+When using any of the encryption methods, you will be prompted to enter an
+encryption key. Type the key normally, or open the last symbols keypad and
+select the QR glyph to scan it. Scanned text returns to the keypad for review
+and editing. Additionally, you have the option to set a custom ID for easier
+management of your mnemonics. If a custom ID is not specified, the current
+loaded wallet fingerprint will be used.
```
qlrd, line 63: Just to be less diff:

```suggestion
When using any of the encryption methods, you will be prompted to enter an encryption key. Type the key normally, or open the last symbols keypad and select the QR glyph to scan it. Scanned text returns to the keypad for review and editing. Additionally, you have the option to set a custom ID for easier management of your mnemonics. If a custom ID is not specified, the current loaded wallet fingerprint will be used.
```

docs/getting-started/usage/navigating-the-main-menu.en.md
```diff
@@ -180,12 +185,13 @@ Krux also allows you to verify a descriptor's receive and change addresses witho
...
 
 #### Passphrase
-<img src="../../../img/maixpy_m5stickv/passphrase-load-options-250.png" align="right" class="m5stickv">
-<img src="../../../img/maixpy_amigo/passphrase-load-options-300.png" align="right" class="amigo">
 
-If you forgot to load a passphrase while loading your wallet, or if you use multiple passphrases with the same mnemonic, you can add, replace, or remove a passphrase here. Simply choose between typing or scanning it.
+If you forgot to load a passphrase while loading your wallet, or if you use
+multiple passphrases with the same mnemonic, you can add, replace, or remove a
+passphrase here. Type it normally, or select the QR glyph on the last symbols
+keypad to scan it.
```
qlrd, line 192: Just to be less diff:

```suggestion
If you forgot to load a passphrase while loading your wallet, or if you use multiple passphrases with the same mnemonic, you can add, replace, or remove a passphrase here. Type it normally, or select the QR glyph on the last symbols keypad to scan it.
```

docs/getting-started/settings.en.md
```diff
@@ -181,6 +181,10 @@ When `True`, Krux will disable the [New Mnemonic](./usage/generating-a-mnemonic.
 
 Create or modify a Tamper Check Code. This code will be required every time [Tamper Check Flash Hash](features/tamper-detection.md) is executed.
 
+Creating or changing the code always uses manual keypad input. During
+verification, the QR glyph on the last symbols keypad can scan a code for review
+and editing before it is submitted.
+
```
qlrd, line 187: Just to be less diff:

```suggestion
Creating or changing the code always uses manual keypad input. During verification, the QR glyph on the last symbols keypad can scan a code for review and editing before it is submitted.

```

qlrd (commented):
Some optional docs nits:

odudex:
got a Bug: scanned binary data raises an uncaught ValueError

The QR code glyph must be created within the bdf files of firmware/font, for 14, 16 and 24 heights, then baked into Maixpy

There are more nits down the road but the above is priority.

oroderico:
@odudex  Thanks! I tried to reproduce the binary issue by scanning a raw 0x8f QR in the KEF decryption flow, but it
worked and showed 0x8f. Do you remember the payload and exact flow you used? A traceback and device model would
also help.

About the glyph: should I use a private codepoint such as U+E000 and add versions for the existing 8×14, 8×16, and 12×24 font cells to the three Terminus BDF files, then rebuild the MaixPy fonts?

I also noticed a visual issue: after rejecting Proceed? and selecting QR again, the Overwrite? prompt is drawn over the keypad. I’ll fix that as well once we confirm the points above.

odudex:
> I tried to reproduce the binary issue by scanning a raw 0x8f QR in the KEF decryption flow, but it
> worked and showed 0x8f. Do you remember the payload and exact flow you used? A traceback and device model would
> also help.

Sorry I defined it poorly. The keypad scanner must block loading binary data. For passphrases it did block in my tests, but for KEF key it didn't. Only `str` should be allowed.

> should I use a private codepoint such as U+E000 and add versions for the existing 8×14, 8×16, and 12×24 font cells to the three Terminus BDF files, then rebuild the MaixPy fonts?

Yes

oroderico:
Both fixed, rebased on develop. Built and flashed on a TZT.

Binary is rejected everywhere now (0a4c0da9). Wider than you saw: passphrases only blocked KEF data, because `.decode()` sat inside the `try` catching ValueError, so non-KEF binary skipped it. The keypad was also returning bytes straight out.

Glyph baked into the .bdf files at U+E000, 14/16/24px (f404ade2). Also fixed the Overwrite? prompt drawing over the keypad.

The baked fonts live in MaixPy and I can't push there, so building this branch needs `poe update-glyphs` first or you get a blank key. Want me to open a PR on selfcustody/MaixPy?

docs/getting-started/features/encryption/encryption.en.md
oroderico: Applied, except the binary sentence. That was right when you wrote it, but odudex asked for binary to be rejected instead, so the docs now say that. Rest is your wording.

docs/getting-started/features/encryption/encryption.en.md
oroderico: No, that was accidental. Restored in 167a27c2.

tests/pages/test_tc_code_verification.py
oroderico: Fair point, it's the same secret every iteration so re-patching in the loop does nothing. Moved it to a fixture in 230e478a. Skipped the wrapper variant since no case varies the secret.

qlrd:
> Both fixed, rebased on develop. Built and flashed on a TZT.
> 
> Binary is rejected everywhere now ([0a4c0da](https://github.com/selfcustody/krux/commit/0a4c0da9cbe15f568c571b80b9b9605f9c16c007)). Wider than you saw: passphrases only blocked KEF data, because `.decode()` sat inside the `try` catching ValueError, so non-KEF binary skipped it. The keypad was also returning bytes straight out.
> 
> Glyph baked into the .bdf files at U+E000, 14/16/24px ([f404ade](https://github.com/selfcustody/krux/commit/f404ade229b94e765bc0fd47816b3c5e16d6629a)). Also fixed the Overwrite? prompt drawing over the keypad.
> 
> The baked fonts live in MaixPy and I can't push there, so building this branch needs `poe update-glyphs` first or you get a blank key. Want me to open a PR on selfcustody/MaixPy?

Hi @oroderico, i was doing a review on this PR and when I typed:

```bash
```

I got:

```bash
fatal: Fetched in submodule path 'firmware/Maixpy', but it not did contain ca57b2d... . Direct fetching of that commit failed
```

I fixed with something like that -- localy -- so i could start to do devices tests, but don't know if it's a error from my side or something added wen you did the 230e478 or c21409a:

```bash
cd firmware/MaixPy && git checkout 4c1c0880 && cd ../..
git add firmware/MaixPy
git commit -m "fix: fix me"
```

With this i was able to test latest updates on both `amigo` and `wonder_mv` as described in PR description. Just need to know the issue talked above.

oroderico:
@qlrd not your fault — the submodule pointer was the problem, and your workaround was the right call.

`firmware/MaixPy` was pinned to `ca57b2dd4`, the glyph commit, which only ever existed in my local clone. That is why `git submodule update --recursive` failed for you, and it is also why all seven CI jobs were red: they die at checkout, not in any test.

```
fatal: remote error: upload-pack: not our ref ca57b2dd41ee...
```

Opened selfcustody/MaixPy#58 with that commit rebased onto `e1e54e0d`, the pointer develop already uses. Once it merges the submodule resolves from `selfcustody/MaixPy` for everyone and CI can run. @odudex, that answers my question from 11/08 — I went ahead and opened it.

The pointer here now references the rebased commit, so CI stays red on checkout until #58 lands. Everything else is ready.

### Merged develop

Conflicts with #911 and #935 resolved:

- `capture_from_keypad` keeps both `scan_fn` and the new `buffer_suffix` / `buffer_short_suffix`.
- `PassphraseEditor._load_qr_passphrase` adopts the silent back-out from #935, returning `None` instead of `MENU_CONTINUE`. It is a keypad `scan_fn` now, not a submenu handler, and `None` is what preserves the buffer. Same user-visible behavior you asked for: leaving the scanner is not a failure and flashes nothing.
- `test_qr_passphrase_camera_back_is_silent` and `test_encryption_key_camera_back_is_silent` rewritten for the keypad scan action. They navigated the Type/Scan submenu, which this PR removes, so the button sequences no longer applied. Both assertions are intact, including `assert_not_flashed`.

### Validation

- `uv run poe test-simple` — 1195 passed, 1 xfailed
- `uv run poe lint` — 10.00/10 on both `src` and scripts
- `uv run black --check` — clean
- `uv run poe i18n validate` — OK
- `git diff --check` — clean

Staying in draft until #58 merges and CI goes green here.

jdlcdl:
I was hiding under a rock while you all were rock'n it!  I see this pr finally.

docs/getting-started/features/encryption/encryption.en.md
```diff
@@ -8,7 +8,14 @@ During encryption, users may be prompted to override their preferences or select
 <img src="../../../../img/maixpy_m5stickv/load-mnemonic-kef-via-qr-250.png" align="right" class="m5stickv">
 <img src="../../../../img/maixpy_amigo/load-mnemonic-kef-via-qr-300.png" align="right" class="amigo">
 
-When Krux detects data resembling a KEF-encrypted envelope, it prompts the user to *"Decrypt?"*, showing the KEF version, envelope ID (or label), and the PBKDF2 iteration count used during creation. To decrypt, the same key (typed or scanned) must be provided. Once unlocked, Krux uses the plaintext within context. If decryption is declined, the raw envelope is used instead—usually resulting in an error, since KEF data is meaningless without decryption.
+When Krux detects data resembling a KEF-encrypted envelope, it prompts the user
+to *"Decrypt?"*, showing the KEF version, envelope ID (or label), and the PBKDF2
```
qlrd: I think a trusthworthy link here overexplaining what is PBKDF2 here would be neat:

```suggestion
to *"Decrypt?"*, showing the KEF version, envelope ID (or label), and the [PBKDF2](<somelink>)
```

src/krux/pages/encryption_ui.py
```diff
@@ -542,11 +528,38 @@ def load_key(self, data=""):
...
+            return None
+        except KeyError:
+            self.flash_error(t("Failed to decrypt"))
+            return None
+        except ValueError:
+            # ValueError=not KEF or declined to decrypt
+            if isinstance(key, str):
+                return key
+            self.flash_error(t("Failed to load"))
+            return None
```
qlrd, line 561: Just thoughs: any other `Exception` could be raised here? if yes, it need a general `raise Exception`.

src/krux/pages/encryption_ui.py
```diff
@@ -558,12 +571,17 @@ def load_qr_encryption_key(self):
...
+        return self._decode_key(data)
+
+    @staticmethod
+    def _decode_key(key):
+        """Decodes a UTF-8 byte key while preserving non-text data."""
+        if not isinstance(key, bytes):
+            return key
+        try:
+            return key.decode()
+        except (UnicodeError, TypeError):
```
qlrd, line 583: Just thoughs: any other Exception could be raised here? if yes, it need a general raise Exception.

oroderico:
No worries, thanks for taking a look! Status here: this is in draft waiting on selfcustody/MaixPy#58 (the submodule pointer fix) to merge — once that lands I'll update the `firmware/MaixPy` pointer here and CI should go green. Everything else (tests, lint, i18n) is passing locally.

jdlcdl:
> this is in draft waiting on [selfcustody/MaixPy#58](https://github.com/selfcustody/MaixPy/pull/58)

I've created a new branch in the selfcustody/MaixPy repo named "v26.09.0_plus_qrglyph" which is a different commit hash containing the exact changes of pr58 (except that instead of against "master" it is against "develop" which is already the expected MaixPy HEAD for krux v26.09.0.   I believe you can update this pr pointing to that new branch so that others can "submodule update" and pick-up the new pointer to an existing branch.

odudex:
Focusing a bit on the glyph itself, I think it deserves a some more debate. When I tested it on device my impression was it is "over-detailed". 1 pixel details might look like noise. How about something simpler, like FontAwesome's QR code icon:
image

qlrd:
> Focusing a bit on the glyph itself, I think it deserves a some more debate. When I tested it on device my impression was it is "over-detailed". 1 pixel details might look like noise. How about something simpler, like FontAwesome's QR code icon: 

liked, very aesthetical

oroderico:
Applied the simplified glyph design odudex suggested (finder-pattern corners + checkerboard, similar spirit to FontAwesome's QR icon) across all three font sizes (14/16/24px). Updated `firmware/font/qr-u{14,16,24}.bdf` (`cf049be2`) and the corresponding `font_device.h` in the MaixPy submodule (selfcustody/MaixPy#58, `ca133f4d`), with the submodule pointer updated here (`4ad288d7`). `tests/test_font_glyphs.py` updated to match the new bitmap.

Also merged `develop` into this branch (no conflicts) — was 20 commits behind, now in sync. Picked up the format-first mnemonic loading feature, an embit bump, and docs/CI changes; none of it touches the code this PR changes.

Full validation:
- `uv run poe test-simple` — 1203 passed, 1 xfailed
- `uv run poe lint` — 10.00/10
- `uv run poe i18n validate` — OK
- Built and flashed on a TZT — glyph confirmed legible at all three sizes

Still on the submodule pointer issue from before — will push once that's sorted.

qlrd (commented):
doubts:

PR #819, opened 2026-01-12
feat(docs): add sphinx doc builder

qlrd:
### What is this PR for?

Krux lack of a techinical documentation. This commit add `sphinx` module
to generate a proper [readthedocs](https://readsthedocs.com/)
documentation for Krux firmware (and not a user documentation) and a
setup for install aditional deps, generation of rst files, build to html
files and serve.

### Changes made to:
- [ ] Code
- [ ] Tests
- [x] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [x] Docs update
- [ ] Other


tadeubas:
Thanks for working on this! Just a small clarification:

ReadTheDocs is a great tool for API documentation, where there are stable endpoints, request/response schemas, etc. However, Krux is not an API or python / micropython library (lib) - it’s just a firmware for an air-gapped Bitcoin signing device.

Because of that distinction, I’d suggest to focus the documentation on technical firmware internals, build instructions, and usage details rather than framing it like an API / lib reference.

odudex:
I don't think it's needed, especially for python code, which speaks for itself more clearly.
This PR, together with #821 adds more than 100 files, and thousands of lines to be maintained.

I suggest to open an issue to keep the idea open to discussion, and in case it's demanded, PR can be re-opened.

PR #820, opened 2026-01-13
Simplify internal key validation for taproot

### LINKED ISSUE #807: Key path spend disabling in miniTapScript
Hi!
I was trying to play with Krux and Nunchuck-Android wallet.
I came across a situation that I described on an issue in [their repo](https://github.com/nunchuk-io/nunchuk-android/issues/99).
It might be a problem on Krux, or on Nunchuck, or I am just being stupid and trying to use the wallets off spec.

I will paste below what I opened there:

> Hi, I came across an issue with the Nunchuck Android wallet (version 2.2.0 311, from Google Play), but I am not sure if it is really an issue in this codebase, another codebase ([Krux](https://github.com/selfcustody/krux)) or just me being reckless and trying to use the wallet off spec.
> But maybe it is indeed an issue here, sorry if not.
> 
> In this [blog post](https://nunchuk.io/blog/miniscript-programmable-bitcoin) you say that Taproot Miniscript is supported for Coldcard, Ledger and Specter DIY. I beleive this is outdated (Jade is also already supported, right?). Actually, as per my reckless poking around, Jade was the only option that let me add a Taproot Miniscript key.
> 
> But I dont have a Jade. I am using a Krux wallet.
> 
> So, here is what I've done: on Krux, I create a seed and configure it for a [mainnet, miniscript, taproot, m/48h/0h/0h/2h] wallet descriptor. Export xpub.
> 
> On Nunchuck, import key (the xpub above). I did it first as a Generic Airgap type of key, more on that later.
> Then, create wallet: miniscript -> taproot -> select any kind of miniscript -> add key.
> Here it says, greyed out, "Key path disabled".
> To add keys, the Krux key imported will also be grayed out: "Keys not yet supporting Taproot".
> 
> I went back to the key import. I tried all the ones that would allow me to import an xpub from QR: Coldcard, Jade, SeedSigner, Keystone, Foundation.
> Of all of these, the only option that would make the key available for use on a miniTapScript was the Jade.
> I am really lost in this Kotlin thing, can't understant it, so I couldn't find where in the code this would be set.
> 
> Well, I went ahead with the "trans Jade" (actually a Krux miniTapScript xpub) and created a wallet. Then, export Descriptor. And try to import it on the Krux.
> 
> The Krux gives me "Internal key not provably unspendable".
> I can see how they check for it: [Krux tapkeypath NUMS (Nothing Up My Sleeve)](https://github.com/selfcustody/krux/blob/main/src/krux/wallet.py#L201)
> 
> I also exported the descriptor from Nunchuck as text and base58-decoded the first xpub (the key path). It does use the [BIP-341 NUMS](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs) 0x50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0
> 
> So, in summary, both Nunchuck and Krux are apparently doing it alright: disable the taproot key path with the BIP341 suggested NUMS.
> But, for some reason, they are disagreeing somewhere. I can see how Krux did it (hash the other keys in series and use that as chaincode for the xpub). But I can't understant Kotlin and I don't know how Nunchuck did it.
> 
> Sorry to maybe bother with my ignorance, but maybe you can find the origin of this. Thank you for your attention to this matter!

rafaelpac:
Fix #807 

### What is this PR for?
Simplify key path validation for taproot miniscript.
When a wallet is defined as a taproot miniscript, it is possible to disable the key path spend using the [BIP-0341 NUMS](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs).
There is no standard for how to generate the taproot keypath chaincode to be used with the NUMS pubkey.
Liana does it by hashing the taptree keys;
Coldcard uses a random chaincode;
Nunchuck hashes the taptree keys, but remove duplicates and sort the keys before hashing.

As far as a hardware signer is concerned, one which doesn't generate wallets like Krux, it is not necessary to check the chaincode, because the NUMS pubkey alone guarantees that the key path is disabled.

This fix changes the verification logic to check only the NUMS pubkey and ignore the chaincode.

### Changes made to:
- [X] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

- [X] No, I don't have a device here with me, but I tested the wallet.load function with some test vectors, all successful. Please, test it on a device during review.

- [X] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other


odudex:
Nice work.
Please checkout and target  the PR to `develop` branch.

odudex:
Thank you!

PR #821, opened 2026-01-17
docs: add source documentation for `readthedocs` poe task

qlrd:
### What is this PR for?

Add a big set of techinical documentation for every package, module and function of krux `firmware`. It's important to say that the `src.boot` isnt autogenerated and `src.krux` is. This happen because of issues during setup that could be fixed in a followup.

Depends on #819 

### Changes made to:
- [ ] Code
- [ ] Tests
- [x] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [x] Docs update
- [ ] Other


odudex:
As said in #819 , we need more demand to justify the addition of a so big infrastructure

PR #842, opened 2026-03-25
Remove unreachable code after while true

tadeubas:
### What is this PR for?
Remove unreachable code after while true, as pointed out by `vulture` with 100% confidence.

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/src/krux/pages/encryption_ui.py
+++ b/src/krux/pages/encryption_ui.py
@@ -114,7 +114,6 @@ def decrypt_kef(ctx, data):
         kef_envelope = KEFEnvelope(ctx)
         if not kef_envelope.parse(data):
             return data
-    raise ValueError(err)
 
 
 def prompt_for_text_update(


tadeubas:
Forced-push to fix PR with just one removal. Opened an issue on vulture for the false positive: https://github.com/jendrikseipp/vulture/issues/413

src/krux/pages/encryption_ui.py
jdlcdl, line 111: Would it be more clear to future readers to `break` here and still `raise ValueError(err)` at bottom so that it doesn't look like falling off end of function is possible?

tadeubas: I was think it was a good idea, but than called AI for help, plz check the response: https://chatgpt.com/share/69c3e205-b8fc-83e9-b278-7894b33fb241

odudex:
Thank you!

Issue #843, opened 2026-03-30
Security Audit: 8 Critical + 12 High findings (AI-assisted audit)


smlbnr:
## Security Audit Report — Krux Bitcoin Hardware Wallet

**Date:** 2026-03-29  
**Commit:** 76feb7b (release-26.03.0)  
**Auditors:** 6 AI agents (Claude Opus 4.6 ×4, Gemini 2.5 Pro, GPT-4.1)  
**Scope:** Full source code — 212 Python files, all critical paths

### Summary

**47 findings:** 8 Critical, 12 High, 18 Medium, 9 Low

---

### CRITICAL FINDINGS

| # | File | Description | Impact |
|---|------|-------------|--------|
| C1 | `key.py:159-178` | Mnemonic, passphrase, root key persist as plaintext in RAM — no zeroization possible (MicroPython immutable strings) | Cold-boot / JTAG memory dump extracts seed |
| C2 | `psbt.py:436,475` | **Sighash flags not verified before signing.** If PSBT requests `SIGHASH_NONE`, no user alert. Fully delegated to embit. | `SIGHASH_NONE` = attacker redirects ALL outputs after signing. **Total loss of funds.** |
| C3 | `sign_message_ui.py:255-256` | 32-byte data signed directly as raw hash (no SHA256 prefix applied). Attacker sends transaction sighash as "message". | Fund theft via transaction signature disguised as message signing |
| C4 | `bbqr.py:96-105` | `deflate_decompress` calls `d.read()` without output size limit. Zip bomb via BBQR encoding "Z". | OOM crash on K210 (6MB RAM) |
| C5 | `qr.py:166-193` | `QRPartParser` accepts unlimited `part_total` in pMofN. No cap on number of parts accumulated. | OOM crash, device DoS |
| C6 | `settings.py:193-199` | `settings.json` from SD card loaded without schema validation. Attacker can disable `auto_shutdown`, `boot_flash_hash`, set `hide_mnemonic: false`. | Security configurations silently disabled via physical SD access |
| C7 | `psbt.py:288` | `fee_percent` divides by `out_amount` — ZeroDivisionError if all outputs are zero-value (OP_RETURN only PSBT). | Unhandled firmware crash during transaction review |
| C8 | `wallet.py:363` | `parse_key_value_file` accepts `m=0` in multisig policy "0 of 3" → `sortedmulti(0,...)` requires zero signatures. | **Anyone can spend funds** sent to this address |

---

[...]

PR #845, opened 2026-03-30
Fix: Reject multisig policies with m=0 or m>n and guard against ZeroDivisionError in fee calculation and 

odudex:
### What is this PR for?

- Reject multisig policies with m=0 (zero signatures = anyone can spend) or m>n (impossible quorum) when loading wallets from BlueWallet INI format. Also adds defense-in-depth validation in descriptor loading path. Addresses security audit C8 of #843.

- Handles the case where all PSBT outputs are zero-value (e.g. OP_RETURN only transactions) by checking out_amount > 0 before dividing. Sets fee_percent to 100.0% when output amount is zero. Addresses audit C7 of #843..

### Changes made to:
- [X] Code
- [X] Tests
- [ ] Docs
- [X] CHANGELOG

- [X] Yes, build and tested on Amigo
- 100% fee Tx
- 0ofM multisig rejection
- NofM where N>M multisig rejection

- [X] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other


joaozinhom (commented):
CACK e389f4ed44931acbc9c73200c01e4e627ff4af1a

PR #850, opened 2026-04-07
fix: validate settings.json on load to prevent OOM and type confusion

odudex:
### What is this PR for?
Enforces a maximum file size and rejects non-object payloads when loading settings.json from SD/flash.

### Changes made to:
- [X] Code
- [X] Tests
- [ ] Docs
- [X] CHANGELOG

### Did you build the code and tested on device?
- [X] Yes, build and tested on Amigo

### What is the purpose of this pull request?
- [X] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

PR #848, opened 2026-03-31
fix: add zip bomb protection and QR part limit enforcement

odudex:
### What is this PR for?
DeflateIO now enforces a 100KB max decompressed size at the C level, protecting all callers (BBQR and KEF) from zip bomb OOM on K210. pMofN parser validates part_total (1-99) and part_index range. BBQR parser validates part_total >= 1.
Issues C4 and C5 from #843

### Changes made to:
- [X] Code
- [X] Tests
- [ ] Docs
- [X] CHANGELOG

### Did you build the code and tested on device?
- [X] Yes, build and tested on Amigo

### What is the purpose of this pull request?
- [X] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

PR #844, opened 2026-03-30
fix: reject PSBT inputs with non-standard sighash types before signing

odudex:
### What is this PR for?

Although Embit already refuses to sign inputs with non-ALL sighash by default, this adds a redundant pre-sign validation that refuses to sign if any input requests SIGHASH_NONE, SIGHASH_SINGLE, or ANYONECANPAY, which could allow an attacker to redirect funds after signing. Addresses security audit C2 of #843

### Changes made to:
- [X] Code
- [X] Tests
- [ ] Docs
- [X] CHANGELOG

### Did you build the code and tested on device?
- [X] Yes, build and tested on Amigo + Sparrow

### What is the purpose of this pull request?
- [X] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

PR #841, opened 2026-03-25
Remove duplicated HEX and OCT digit constants

tadeubas:
### What is this PR for?

Remove duplicated HEX and OCT digit constants (were moved to `src/krux/pages/mnemonic_loader.py`)

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

### Did you build the code and tested on device?
- [ ] Yes, build and tested on 

### What is the purpose of this pull request?
- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

PR #855, opened 2026-04-22
Release 26.04.0

odudex:
### What is this PR for?
Release 26.04.0 — primarily security hardening following the audit tracked in issue #843
#### Security fixes

**Signing-path hardening**
- PSBT — reject non-standard sighash types (#844, audit C2): refuses to sign inputs that request SIGHASH_NONE, SIGHASH_SINGLE, or ANYONECANPAY, which could let an attacker redirect funds after signing.
- Message signing — warn on raw hashes (#846, audit C3): detects 32-byte raw hashes / 64-char hex strings and shows a warning before signing, mitigating the risk of a transaction sighash disguised as a message.
- Multisig quorum validation (#845): rejects key-value wallet files with m=0 or m>n, and guards against ZeroDivisionError in fee calculation for zero-value outputs.
- PSBT multi-key descriptors: reject more than one origin-less xpub (would previously silently overwrite and lose cosigner identity); the taproot internal-key exception is preserved.

**Memory / DoS hardening**
- Zip-bomb protection (#848): DeflateIO enforces a 100 KB max decompressed size at the C level, protecting all callers (BBQR "Z" encoding, KEF decryption) from OOM on K210.
- QR part-count limits (#848): pMofN validates part_total (1–99) and part_index; BBQR parser requires part_total ≥ 1, preventing OOM via unbounded part accumulation.
- settings.json validation (#850): enforces a max file size and rejects non-object payloads on load, preventing OOM and type confusion from a malicious SD card.

**KEF / key-stretching hardening (#851)**
- Enforce a minimum of 10 000 PBKDF2 iterations on kef.unwrap — a malicious envelope declaring iterations=0  
can no longer bypass key stretching.                                                                         
- Add in-session exponential backoff (1 s, 2 s, 4 s … capped at 30 s) on failed KEF decryption attempts.     
Counter lives in RAM only; a power cycle resets it (accepted trade-off to avoid writing lockout state to     
flash).

**Filesystem hardening (#851)**
- File manager filters ., .., and entries containing path separators from os.listdir results on SD, blocking directory traversal via crafted FAT entries. The intended .. up-navigation still works (added explicitly by the page).

#### Refactors

- Remove unreachable code after while True (#842).
- Deduplicate HEX/OCT digit constants (#841).


jdlcdl (approved):
I have visually reviewed the proposed changes in "Files changed" again as a review and originally when most PRs were submitted since last month.

I have "played" randomly on Amigo and Yahboom_devkit for the past 2 days without any problems.

ACK for release (as others confirm reproducibility)

qlrd:
ACK

joaozinhom (approved):

PR #857, opened 2026-04-28
chore: drop translate dev dep to remove exposure

odudex:
### What is this PR for?
Removes the `translate` package (which often brings vulnerabilities to development environment through its dependencies) and its i18n auto-fill helper (print_missing/post_process_translation and the fill* CLI actions).

Contributors can use LLMs from their preference to create new translations.

### Changes made to:
- [X] Code (translation development script)
- [ ] Tests
- [X] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other


qlrd:
This has relation with recent multiple attacks? I mean, remotion could avoid a supply chain attack, right?

odudex:
Krux firmware does not use Python packages, so it can't suffer a supply chain attack. But as we use Python packages in our development tools, we are constantly annoyed by vulnerabilities. This is why we must keep only essential tools and drop those that cause more problems than solutions.

Issue #865, opened 2026-05-29
[Bug] align BIP137 and implement BIP322 message signing


qlrd:
During a verification on BIP-322 i noticed a not complete BIP-137 implementation (but still sign correctly).

`src/krux/pages/home_pages/sign_message_ui.py::_sign_at_address` emit a P2PKH-compressed header byte `31` for any legacy signed message for `singlesig` wallets.
If you use P2SH-P2WPKH or P2WPKH, those header bytes differ from a compliant implementation (see this [gist](https://gist.github.com/qlrd/d61228c1b33d2e93194734233df32ecd)).

Sparrow's permissive Electrum mode accepts wrong byte headers -- signature is correct anyway, masking the spec violation.

**Device(s) affected**
- [x] Amigo
- [x] M5stickV
- [x] Cube
- [x] Dock
- [x] Bit
- [x] Yahboom
- [x] WonderMV

**Version affected**
- [x] Official release (selfcustody/krux): v26.04.0 
- [x] Beta release (odudex/krux_binaries): v26.05.0-beta*

PR #866, opened 2026-05-29
 fix(bip137):  align `recid` with the script type

qlrd:
### What is this PR for?

Minor bug fix. 

Previously, the header byte could not reflect the script type's recid when using `sparrow <2.5.x` for sign messages on some script types (mostly `p2sh-p2wpkh`). Through some manual experiments with different versions of `sparrow` , a possibility that a header byte could mismatched to the script type causes verification to fail on those versions.

This commit add a check of a `recid` as well structure a `bip137.py` that is strict on context of `sparrow 2.5.x` and lenient for `sparrow <2.5.x` (similar what used in `electrum` clients), as well routes `raw hashes` to be not signed as `bip137` but instead as `der` one.

### Changes made to:
- [x] Code
- [x] Tests
- [x] Docs
- [x] CHANGELOG

- [x] Yes, build and tested on Amigo / Sparrow / krux-file-signer:
  - Sparrow 2.3.1
  - Sparrow 2.4.2
  - Sparrow 2.5.2
  - krux file signer with and without `sign [-u]` flag
  
- [x] Bug fix
- [ ] New feature
- [x] Docs update
- [ ] Other

# Context

The bip322 will be added to a follow up so the diffs could be simplified

# Steps to reproduce (WIP)

Context: built different versions of Sparrow (at least 2 with version `<2.5.x`) from source and made `P2PKH`, `P2SH_P2WPKH` and `P2WPKH` sign message procedures with a flashed firmware on device with a simple `oi` (`hi` on portuguese) on Sparrow on given address. 

Below was noted some different behaviours for some different Sparrow's versions. Since used different mnemonics on `testnet` and `testnet4` networks, i think any mnemonic here will be sufficient.

While reproduce expects these results below -- deselecting any mode on Sparrow (`Standard(electrum)`, `trezor(BIP137)` and `BIP322` -- when available):

**Sparrow 2.3.1**
- [x] P2PKH: raw / electrum / trezor modes -> all **electrum** after verify;
- [x] P2SH-P2PWPKH: raw / electrum / trezor modes -> all **trezorBIP137** after verify;
- [x] P2PWPKH -> raw / electrum / trezor modes -> all **electrum** after verify;
   
**Sparrow 2.4.2**
- [x] P2PKH: raw / electrum / trezor modes -> all **trezorBIP137** after verify;
- [x] P2SH-P2PWPKH: raw / electrum / trezor modes -> all **trezorBIP137** after verify;
- [x] P2PWPKH -> raw / electrum / trezor modes -> all **trezorBIP137** after verify;
  
**Sparrow 2.5.2**
- [x] P2PKH: raw / electrum / trezor modes -> all **electrum** after verify[^1];
- [x] P2SH-P2PWPKH: raw / electrum / trezor modes -> all **trezorBIP137** after verify[^2];
- [x] P2PWPKH: raw / electrum / trezor modes -> all **electrum** after verify;
   
**Krux file signer**

Download and update to `main` branch; and run the following command:

```bash
uv run poe sign -f README.md # follow the procedures, krux will show `raw hash`.
```

Then the command below should produce:

```bash
uv run poe verify -f README.md -s README.md.sig -p pubkey.pem
Poe => python src/ksigner.py verify -f README.md -s README.md.sig -p pubkey.pem
[HH:MM:ss MM/DD/YY time] Verifying signature:
Signature Verified Successfully
```

Also need to verify with `--uncompressed` flag:

```bash
uv run poe sign -f README.md -u -s README.md.u.sig
```

```bash
uv run poe verify -f README.md -s README.md.u.sig -p pubkey.pem 
Poe => python src/ksigner.py verify -f README.md -s README.md.sig -p pubkey.pem
[HH:MM:ss MM/DD/YY time] Verifying signature:
Signature Verified Successfully
```

Note that the `-u` flag should produce different pubvkeys.

--
[^1]: Sparrow disables trezor(BIP137) and BIP322 for P2PKH (not disabled on previous versions)
[^2]: Sparrow disable BIP322 for P2SH_P2WPKH (not disabled on previous versions)

### DIFF
--- a/src/krux/bip137.py
+++ b/src/krux/bip137.py
@@ -1,0 +1,77 @@
+# The MIT License (MIT)
+
+# Copyright (c) 2021-2026 Krux contributors
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+P2PKH_UNCOMPRESSED_HEADER = 27
+P2PKH_HEADER = 31
+P2SH_P2WPKH_HEADER = 35
+P2WPKH_HEADER = 39
+MESSAGE_MAGIC = b"\x18Bitcoin Signed Message:\n"
+RECID_OFFSET = P2PKH_HEADER - P2PKH_UNCOMPRESSED_HEADER
+
+
+def message_commitment(message):
+    """Double-SHA256 commitment over the BIP-137 magic"""
+    from embit import compact
+
+    try:
+        import uhashlib as hashlib
+    except ImportError:
+        import hashlib
+
+    # BIP137 commitment message:
+    # `SHA256(SHA256(MAGIC // varint // message))`
+    varint = compact.to_bytes(len(message))
+    _message = MESSAGE_MAGIC + varint + message
+    return hashlib.sha256(hashlib.sha256(_message).digest()).digest()
+
+
+def build_header(raw_sig, script_type, compressed=True):
+    """Build header byte from raw signature and script_type"""
+    # Avoid some unexpected header
+    rsig = raw_sig[0]
+    if not P2PKH_UNCOMPRESSED_HEADER <= rsig <= P2PKH_HEADER + 3:
+        raise ValueError("Invalid sig header: %d" % rsig)
+
+    # grab the 2 least significant bits as recId
+    # and normalize with a minimum p2pkh (uncompressed) flag
+    recid = (raw_sig[0] - P2PKH_UNCOMPRESSED_HEADER) & 3
+
+    if script_type == "p2sh-p2wpkh":
+        return P2SH_P2WPKH_HEADER + recid
+    if script_type == "p2wpkh":
+        return P2WPKH_HEADER + recid
+
+    # compressed=False is only meaningful for p2pkh
+    return (
+        P2PKH_UNCOMPRESSED_HEADER
+        + recid
+        + (0 if not compressed and script_type == "p2pkh" else RECID_OFFSET)
+    )
+
+
+def sign(message, key, derivation, script_type="p2pkh", compressed=True):
+    """Sign a BIP137 `message` with `key` at `derivation` for some `script_type`"""
+    commitment = message_commitment(message)
+    raw_sig = key.sign_at(derivation, commitment)
+    header = build_header(raw_sig, script_type, compressed)
+    sig = bytes([header]) + raw_sig[1:]
+    return (commitment, sig)
--- a/src/krux/pages/home_pages/sign_message_ui.py
+++ b/src/krux/pages/home_pages/sign_message_ui.py
@@ -20,7 +20,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 # THE SOFTWARE.
 
-from embit import bip32, compact, script
+from embit import bip32, script
 from embit.networks import NETWORKS
 import hashlib
 import binascii
@@ -133,7 +133,8 @@ class SignMessage(Utils):
         return addr
 
     def _sign_at_address(self, message, derivation_str, address=""):
-        """Signs a message at a derived Bitcoin address"""
+        """Signs a BIP137 message at a derived Bitcoin address"""
+        from krux import bip137
 
         derivation = bip32.parse_path(derivation_str)
         self._display_message_sign_prompt(
@@ -143,15 +144,8 @@ class SignMessage(Utils):
         if not self.prompt(t("Sign?"), BOTTOM_PROMPT_LINE):
             return None
 
-        message_hash = hashlib.sha256(
-            hashlib.sha256(
-                b"\x18Bitcoin Signed Message:\n"
-                + compact.to_bytes(len(message))
-                + message
-            ).digest()
-        ).digest()
-
-        sig = self.ctx.wallet.key.sign_at(derivation, message_hash)
+        script_type = self.get_script_type_from_path(derivation_str) or "p2pkh"
+        _, sig = bip137.sign(message, self.ctx.wallet.key, derivation, script_type)
         self._display_signature(base_encode(sig, 64))
         return sig
 
@@ -234,6 +228,8 @@ class SignMessage(Utils):
 
     def sign_standard_message(self, data):
         """Signs a standard message"""
+        from krux import bip137
+
         message_hash, is_raw_hash = self._compute_message_hash(data)
         if message_hash is None:
             return ""
@@ -248,6 +244,8 @@ class SignMessage(Utils):
             )
             if not self.prompt(t("Proceed?"), BOTTOM_PROMPT_LINE):
                 return ""
+        else:
+            message_hash = bip137.message_commitment(data)
 
         self.ctx.display.clear()
         self.ctx.display.draw_centered_text(
@@ -257,7 +255,13 @@ class SignMessage(Utils):
         if not self.prompt(t("Sign?"), BOTTOM_PROMPT_LINE):
             return ""
 
-        sig = self.ctx.wallet.key.sign(message_hash).serialize()
+        key = self.ctx.wallet.key
+        if is_raw_hash:
+            sig = key.sign(message_hash).serialize()
+        else:
+            _, sig = bip137.sign(
+                data, key, bip32.parse_path(key.derivation), key.script_type
+            )
         self._display_signature(base_encode(sig, 64))
         return sig
 
--- a/tests/test_bip137.py
+++ b/tests/test_bip137.py
@@ -1,0 +1,298 @@
+import pytest
+
+from .pages.home_pages.test_home import tdata
+
+P2PKH_DERIV = "m/44h/0h/0h/0/3"
+P2SH_P2WPKH_DERIV = "m/49h/0h/0h/0/3"
+P2WPKH_DERIV = "m/84h/0h/0h/0/3"
+P2TR_DERIV = "m/86h/0h/0h/0/3"
+
+MESSAGE = b"krux+bip137"
+
+
+@pytest.fixture
+def address_for():
+    def _addr(pub, script_type, network):
+        from embit import script
+
+        if script_type == "p2pkh":
+            return script.p2pkh(pub).address(network=network)
+        if script_type == "p2sh-p2wpkh":
+            return script.p2sh(script.p2wpkh(pub)).address(network=network)
+        if script_type == "p2wpkh":
+            return script.p2wpkh(pub).address(network=network)
+        if script_type == "p2tr":
+            return script.p2tr(pub).address(network=network)
+        return None
+
+    return _addr
+
+
+@pytest.fixture
+def fake_raw_sig():
+    def _sig(recid):
+        import secrets
+        from krux.bip137 import P2PKH_HEADER
+
+        return bytes([P2PKH_HEADER + recid]) + secrets.token_bytes(64)
+
+    return _sig
+
+
+@pytest.fixture
+def lenient_sign():
+    def _sign(key, derivation, message):
+        import hashlib
+        from embit import compact
+        from krux.bip137 import MESSAGE_MAGIC
+
+        commitment = hashlib.sha256(
+            hashlib.sha256(
+                MESSAGE_MAGIC + compact.to_bytes(len(message)) + message
+            ).digest()
+        ).digest()
+        sig = key.sign_at(derivation, commitment)
+        return (commitment, sig)
+
+    return _sign
+
+
+# An idealized strict verifier (Sparrow >=2.5.x).
+#
+# The numbers are defined in BIP137 and means and
+# each one produces different check addresses:
+#
+# - 27-30: uncompressed-pubkey p2pkh
+# - 31-34: compressed-pubkey p2pkh
+# - 35-38: compressed-pubkey p2sh-p2wpkh (p2sh ones are found here)
+# - >38: compressed-pubkey p2wpkh and beyond
+@pytest.fixture
+def strict_verify(address_for):
+    def _verify(sig, commitment, expected_address, network):
+        from embit import ec
+        from embit.util import secp256k1
+
+        header = sig[0]
+        if not 27 <= header <= 42:
+            return False
+
+        recid = (header - 27) & 3
+        if 27 <= header <= 30:
+            script_type, flag = "p2pkh", secp256k1.EC_UNCOMPRESSED
+        elif 31 <= header <= 34:
+            script_type, flag = "p2pkh", secp256k1.EC_COMPRESSED
+        elif 35 <= header <= 38:
+            script_type, flag = "p2sh-p2wpkh", secp256k1.EC_COMPRESSED
+        else:
+            script_type, flag = "p2wpkh", secp256k1.EC_COMPRESSED
+
+        parsed = secp256k1.ecdsa_recoverable_signature_parse_compact(sig[1:], recid)
+        raw = secp256k1.ecdsa_recover(parsed, commitment)
+        pub = ec.PublicKey.parse(secp256k1.ec_pubkey_serialize(raw, flag))
+        addr = address_for(pub, script_type, network)
+        return addr is not None and addr == expected_address
+
+    return _verify
+
+
+# Idealized lenient verifier (Sparrow  <2.5.x/ Electrum).
+#
+# By lenient we mean that the verification still occurs for recids and flags,
+# but the script type and thus address, will be trusted by a provided script
+# type, not by a checked one.
+@pytest.fixture
+def lenient_verify(address_for):
+    def _verify(sig, commitment, expected_address, network, script_type):
+        from embit import ec
+        from embit.util import secp256k1
+
+        header = sig[0]
+        if not 27 <= header <= 42:
+            return False
+        recid = (header - 27) & 3
+        flag = (
+            secp256k1.EC_UNCOMPRESSED if 27 <= header <= 30 else secp256k1.EC_COMPRESSED
+        )
+
+        parsed = secp256k1.ecdsa_recoverable_signature_parse_compact(sig[1:], recid)
+        raw = secp256k1.ecdsa_recover(parsed, commitment)
+        pub = ec.PublicKey.parse(secp256k1.ec_pubkey_serialize(raw, flag))
+        addr = address_for(pub, script_type, network)
+        return addr is not None and addr == expected_address
+
+    return _verify
+
+
+def test_message_commitment(mocker, m5stickv, tdata):
+    import binascii
+    from krux.bip137 import message_commitment
+
+    assert binascii.hexlify(message_commitment(MESSAGE)) == (
+        b"6f47a0896ff0eb30d36d73ef8783f9796abc01328807835eb258ee042094df22"
+    )
+
+
+def test_build_header(mocker, m5stickv, fake_raw_sig):
+    from krux.bip137 import (
+        build_header,
+        P2PKH_HEADER,
+        P2SH_P2WPKH_HEADER,
+        P2WPKH_HEADER,
+    )
+
+    # (script_type, expected_header_base)
+    cases = [
+        ("p2pkh", P2PKH_HEADER),
+        ("p2sh-p2wpkh", P2SH_P2WPKH_HEADER),
+        ("p2wpkh", P2WPKH_HEADER),
+        # lenient ones could sign p2tr using this conversion
+        ("p2tr", P2PKH_HEADER),
+        ("p2wsh", P2PKH_HEADER),
+    ]
+    for i, case in enumerate(cases):
+        print("Case: %d", i)
+        for recid in range(4):
+            assert build_header(fake_raw_sig(recid), case[0]) == case[1] + recid
+
+
+def test_fail_build_header(mocker, m5stickv):
+    from krux.bip137 import build_header
+
+    for bad in (0, 26, 35, 255):
+        raw = bytes([bad]) + b"\x00" * 64
+        with pytest.raises(ValueError, match="Invalid sig header"):
+            build_header(raw, "p2pkh")
+
+
+def test_lenient_signatures(mocker, m5stickv, tdata, lenient_sign):
+    from embit import bip32
+    from krux.bip137 import (
+        sign,
+        P2PKH_HEADER,
+        P2SH_P2WPKH_HEADER,
+        P2WPKH_HEADER,
+    )
+
+    # (derivation, script_type, current_header_base)
+    cases = [
+        (P2PKH_DERIV, "p2pkh", P2PKH_HEADER),
+        (P2SH_P2WPKH_DERIV, "p2sh-p2wpkh", P2SH_P2WPKH_HEADER),
+        (P2WPKH_DERIV, "p2wpkh", P2WPKH_HEADER),
+        # lenient ones could sign p2tr using this conversion
+        (P2TR_DERIV, "p2tr", P2PKH_HEADER),
+    ]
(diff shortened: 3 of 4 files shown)


odudex (changes requested):
The header-byte fix is correct: single-sig P2SH-P2WPKH and P2WPKH wallets were
emitting legacy-message signatures with the P2PKH-compressed header byte (31-34)
instead of the BIP-137 script-type headers (35-38 / 39-42). Signing now routes
through `bip137.sign`, which preserves the recovery id and rebases the header on
the script type. Verified against the regenerated vectors:

  - bc1qgl5... (P2WPKH):       32 -> 40  (39 + recid)  OK
  - 38CahkV... (P2SH-P2WPKH):  31 -> 35  (35 + recid)  OK
  - tb1qynp... (P2WPKH tnet):  32 -> 40  (39 + recid)  OK

`sign_at` emits `31 + recid`, so `recovery_id = raw_header - 31` is correct and
the re-encoding only swaps the script-type base. Tests pass (5 passed).

## Blocking: taproot message signing shouldn't stop working

This PR makes single-sig taproot (m/86') message signing fail. From a user's
perspective this is a regression / introduced bug, not a tightening: taproot
signing worked before and Sparrow verified the result as valid.

Mechanism: `get_script_type_from_path("m/86'/...")` returns "p2tr", which is
passed to `bip137.sign` -> `build_header`. `build_header` has no p2tr branch, so
it raises ValueError, which the menu catches and shows as an error screen.

Why it really was working (so this is a real loss of function, not removal of a
broken path):
  - Krux signs the standard BIP-137 commitment with an ECDSA-recoverable
    signature using the taproot internal key.
  - Lenient verifiers (Sparrow / Electrum-style) recover the pubkey and
    reconstruct the address of the claimed type from it - for taproot they apply
    the taproot tweak (same as script.p2tr) and the recovered+tweaked key matches
    the bc1p... address. They effectively ignore the header's script-type bits.
  - So the signature genuinely proves control of the taproot key and verifies in
    practice. It's just not a formally specified scheme (BIP-137 has no taproot;
    BIP-322 is the eventual standard).

Required outcome: taproot message signing must keep working. Acceptable options:
  1. Keep the previous behavior - give build_header a p2tr branch that returns a
     valid recoverable header so Krux keeps emitting the Sparrow-verifiable
     signature (this is the no-regression path). The segwit header fix (#865) is
     independent and does not require dropping taproot.
  2. Implement proper BIP-322 signing for taproot.

What is NOT acceptable: shipping a version where taproot signing errors out.

## Other required changes

3. Misleading error message in build_header.
   `"%s legacy sign not supported"` implies the script type itself is
   unsupported. p2tr is a supported wallet type; the limitation is that BIP-137
   legacy message signing has no taproot scheme. If any reject path remains,
   reword to make that clear and point at BIP-322. Also drop the trailing space
   (currently shown verbatim as `ValueError('p2tr legacy sign not supported ')`).

4. Add the MIT license header to src/krux/bip137.py.
   Every other src/krux/*.py file opens with the MIT block; this module starts
   straight at `P2PKH_HEADER = 31`.

5. Dead module-level constants. P2PKH_HEADER / P2SH_P2WPKH_HEADER / P2WPKH_HEADER
   are defined but never used; build_header hardcodes the literals 31/35/39.
   Use the constants or drop them so they cannot drift.

6. Unreachable assertion in test_sign_message_p2tr_bip137.
   The `assert ctx.input.wait_for_button.call_count == len(btn_seq)` sits after
   the raising call inside the `with pytest.raises(...)` block, so it never runs.
   Move it after the block or delete it. (This test should change anyway once
   taproot signing is restored.)

## Suggested (non-blocking)

7. Commit says "add test vectors for bip137 module" but there are no direct unit
   tests for bip137.py; coverage is only via the UI. The API-boundary guards in
   build_header / sign are never exercised. Add tests/test_bip137.py.

odudex:
Why a menu is needed?
Krux should just show the message and sign, everything else should be under the hood.

odudex:
This needs to be simplified.
For example, there's no need for translated error messages. A simple, single generic raised error is enough.

src/krux/bip137.py
```diff
@@ -0,0 +1,62 @@
...
+    return hashlib.sha256(
+        hashlib.sha256(
+            b"\x18Bitcoin Signed Message:\n" + compact.to_bytes(len(message)) + message
+        ).digest()
+    ).digest()
+
+
+def build_header(raw_sig, script_type):
+    """Build header byte from raw signature and script_type"""
+    recid = raw_sig[0] - P2PKH_HEADER
```
jdlcdl: un-settled thoughts:  wouldn't we just ~~left-fshift 6 bits~~ (correction: `byte0 & 3`) so that we grab the 2 least significant bits as recId?

tests/test_bip137.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+@pytest.fixture
+def bip137_verify(address_for):
+    def _verify(sig, commitment, expected_address, network):
+        from embit import ec
+        from embit.util import secp256k1
+
+        header = sig[0]
+        if 31 <= header <= 34:
```
jdlcdl: Is there a 27-30 case missing (for uncompressed p2pkh)?

tests/test_bip137.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+@pytest.fixture
+def old_bip137_verify(address_for):
+    def _verify(sig, commitment, expected_address, network, script_type):
+        from embit import ec
+        from embit.util import secp256k1
+
+        header = sig[0]
+        if 31 <= header <= 34:
```
jdlcdl: missing? (27-30 case for uncompressed p2pkh address)

tests/test_bip137.py
```diff
@@ -0,0 +1,287 @@
...
+        P2SH_P2WPKH_HEADER,
+        P2WPKH_HEADER,
+    )
+
+    # (derivation, script_type, current_header_base)
+    cases = [
+        (P2PKH_DERIV, "p2pkh", P2PKH_HEADER),
+        (P2SH_P2WPKH_DERIV, "p2sh-p2wpkh", P2SH_P2WPKH_HEADER),
+        (P2WPKH_DERIV, "p2wpkh", P2WPKH_HEADER),
+        (P2TR_DERIV, "p2tr", P2PKH_HEADER),
```
jdlcdl, line 182: supported? p2tr for bip137

and same for line 258

RESOLVED: via odudex's comment in convo... p2tr currently works via tapkey used for ecdsa, so don't break something that works.

tests/test_bip137.py
qlrd: yes, but this would be unreachable iirc (as we load compressed)
 Maybe a comment to explain Why i removed? I thought to add uncompressed keys tests, but also thought "let's wait a review"

tests/test_bip137.py
qlrd: yes, but this would be unreachable iirc

src/krux/bip137.py
qlrd: Good!

tests/test_bip137.py
qlrd: `tests/test_bip137` (L276-279) i tried not to support, but asserts p2tr fails BIP-137 verification with a `assert not`. So just to be clear, i think i'm not: I'm not claiming this p2tr bip137 compliant, just checking it, by spec, it will not work.

odudex:
Did you test file/hash signing (and release signing)? I didn't but it seems this PR breaks it.

src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -366,14 +408,89 @@ def _export_to_sd(self, sig, pubkey, message_filename, message="", address=""):
                 False,
             )
 
+    # pylint: disable=too-many-return-statements,too-many-branches
```
odudex: Try not to disable lint checks

odudex (commented):
Some AI caught nits, other than the file signing issue:

## 1. Non-BIP322 PSBT raises an uncaught `ValueError` instead of flashing
`src/krux/pages/home_pages/sign_message_ui.py:436‑437`

```python
if not bip322.check_bip322_psbt(parsed_psbt):
    raise ValueError("Invalid BIP-322 PSBT")
```

A normal fundable PSBT base64-parses fine, fails the BIP322 shape check, and throws. This is
tested (`test_sign_non_bip322_psbt_fail`) so it's intentional, but it's inconsistent with every
other failure in this method, which uses `self.flash_error(t("Failed to load"))` + return
`MENU_CONTINUE` (lines 419, 508). On-device, an uncaught exception is a harsher UX (error
screen) than a flash. Consider `flash_error` + return here for consistency.

---

## 2. BIP137 fallback inside the BIP322 branch does not validate the key-origin fingerprint
`src/krux/pages/home_pages/sign_message_ui.py:455‑462`

For p2pkh / p2sh-p2wpkh BIP322 PSBTs the code falls back to BIP137 and blindly takes
`next(iter(derivations.items()))` without checking the key-origin fingerprint against the wallet
— unlike `bip322.sign` (`bip322.py:117‑129`), which skips non-matching fingerprints and raises
if none match. A p2pkh PSBT from another wallet would be signed with *this* wallet's key at that
path and exported against a displayed address it doesn't own, with no warning. Minor (the
signature simply won't verify), but the two paths should be consistent.

---

## 3. `serialize_bip322` patches fixed byte offsets with no validation
`src/krux/bip322.py:178‑191`

The offset arithmetic is correct for the fixed BIP322-simple shape and the comment is good, but
the version (bytes 8‑11) and sequence (`seq_pos..+4`) are overwritten unconditionally without
confirming they currently hold the embit-substituted defaults. If embit ever changes
serialization ordering, this silently corrupts the PSBT.

Firmware does not use `assert`, so add a defensive check in the module's existing style: before
zeroing, verify the bytes about to be overwritten match the expected substituted version
(`0x01000000`) / sequence, and `raise ValueError("Unexpected BIP-322 PSBT layout")` on mismatch.
That converts a silent corruption into a loud, debuggable failure. (`test_serialize_bip322_zeros_version_and_sequence`
only checks the post-condition, so layout drift wouldn't be caught there either.)

---

## 4. Minor — confusing return naming in `build_bip322_txs`
`src/krux/bip322.py:25,62`

Returns `(to_message, to_spend)`, but per the BIP322 spec these are the *to_spend* and *to_sign*
transactions respectively. The logic is correct (callers use them consistently), but the names
invert the spec vocabulary and make the code harder to verify against the BIP. A rename to
`(to_spend, to_sign)` would match the spec.

---

## Pre-existing, now untested
The P2TR at-address case was removed from `test_sign_message_at_address` (was case 2)

src/krux/bip137.py
qlrd: from what i found, it need a "normalization" (`(byte0 - 27) & 3` -- the p2pkh one), an this works, confirm?

qlrd:
> Did you test file/hash signing (and release signing)? I didn't but it seems this PR breaks it.

I flashed and tested, but not the release signing, will check.

qlrd:
> rify the bytes about to be overwritten match the expected substituted version
> (`0x01000000`)

Good catching, will apply here

tests/test_bip137.py
qlrd: Changed here

odudex:
> I flashed and tested, but not the release signing, will check.

I gave our file signer a fresh blood (and a deep clean). It is useful to test.
https://github.com/selfcustody/krux-file-signer

tests/test_bip137.py
qlrd: changed with your suggestion

tests/test_bip137.py
qlrd: changed with your suggestion

qlrd:
> Did you test file/hash signing (and release signing)? I didn't but it seems this PR breaks it.

I think i(we) found the reason and will re-do, basically i ignored the `raw_hash`.

qlrd:
> > I flashed and tested, but not the release signing, will check.
> 
> I gave our file signer a fresh blood (and a deep clean). It is useful to test. https://github.com/selfcustody/krux-file-signer

After your warning, tested manually the part that must branch the `raw_hash` from bip ones. It procuded a proper DER signature/shas (as before). But `krux-file-signer` had a little bug, and after fixed it produced a verified signature:

Reproduce (with cherry picking my prs on `krux-file-signer` or using raw commands:

```bash
uv run poe sign -f ../krux/README.md -o ../krux/<file>.pem 
Poe => python src/ksigner.py sign -f ../krux/README.md -o ../krux/qlrd.pem
some procedure...
some qrcode...
```

Expected result:

```bash
uv run poe verify -f ../krux/README.md -s ../krux/README.md.sig -p ../krux/<file>.pem
Poe => python src/ksigner.py verify -f ../krux/README.md -s ../krux/README.md.sig -p ../krux/qlrd.pem
[HH:MM:SS mm/dd/yy +-GMT] Verifying signature:
Signature Verified Successfully
```

The stubbed `qlrd.pem`:

```bash
-----BEGIN PUBLIC KEY-----
MDYwEAYHKoZIzj0CAQYFK4EEAAoDIgACB5jsRnBQaWpX1V+drUfen5sQjx8YeEkdMyR67bh1LGw=
-----END PUBLIC KEY-----
```

src/krux/bip137.py
jdlcdl: I suspect I'm missing an important aspect of bip137, but I'll explain my understanding of at least the header byte, so that I can be corrected on my misunderstanding.

1) header byte was set by some ecdsa library when the signature was created,
2) we're just trying to dissect the header so that we know what type of address this is, without modifying the signature,
3) the 2 least significant bits are the recId and we can have this from any sample_byte between 0-255 via sample_byte & 3 (0b00000011 and any number will result 0-3)
4) if we were to subtract 27 from the header byte, which has the 2 least-significant bits high/on/one, then we'd be mucking with whatever recId was set by the ec library.

qlrd: Thanks for your concerns @jdlcdl (EDIT: i would be glad if i can be corrected on any misunderstanding of mine).

I did that from an interpretation between BitcoinJ sample on BIP137 and [`ecdsa_sign_recoverable`](https://github.com/diybitcoinhardware/embit/blob/master/src/embit/util/py_secp256k1.py#L322C26-L322C40).

From what i understood, on the signing side, embit gave a 65-byte recoverable signature with `recId`, internally -- (`sig + bytes([0..3])`).

After, we read that and packed the `byte0 = 27 + recId + 4`.

So, when `raw_sig` came with `byte0` (`27 + recId + 4`) — which is why i did `(header - 27) & 3` and asked for a review on that.

odudex:
PR is adding 1,659 lines. Can you make it more concise?

qlrd:
> PR is adding 1,659 lines. Can you make it more concise?

I will try here, thought in some ways but just started to evaluate it

qlrd:
since majority of diffs are tests, IMO would be simpler to split PRs, with the first one as bip137 (this one, with all checks stated here -- (EDIT: in this thread) and the second as bip322, wdyt @odudex, @jdlcdl?

EDIT: Needed to update ksigner to verify with uncompressed keys: https://github.com/selfcustody/krux-file-signer/pull/51

qlrd:
removed all bip322 as suggested before the diff gets big and added some battery of manual tests to reproduce

qlrd:
Sorry for copilot messages, forgot to deactivate it

qlrd:
I think is ready to review with suggested `diff`less approach, tests and checks. Let me know @odudex @jdlcdl (also prepared a follow-up for bip322).

odudex (approved):
Looks good to me!

qlrd:
@odudex, baked a `feat/bip322` based on this one. Do you think it's worth to add it as draft or wait?

odudex:
Let's merge this, then you focus on BIP322

qlrd:
> Let's merge this, then you focus on BIP322

Just to document a required review from @joaozinhom (i think his fresh knowledge could be worth of).

src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -143,15 +144,8 @@ def _sign_at_address(self, message, derivation_str, address=""):
...
-        message_hash = hashlib.sha256(
-            hashlib.sha256(
-                b"\x18Bitcoin Signed Message:\n"
-                + compact.to_bytes(len(message))
-                + message
-            ).digest()
-        ).digest()
-
-        sig = self.ctx.wallet.key.sign_at(derivation, message_hash)
+        script_type = self.get_script_type_from_path(derivation_str) or "p2pkh"
```
joaozinhom, line 147: The recid fix itself is correct — I decoded the new vectors and they line up (`KN/4…` → byte 40 = `P2WPKH_HEADER + 1`, `IyH8…` → 35 = `P2SH_P2WPKH_HEADER + 0`).

What this line exposes is that the SD flow now decides the script type twice, from two independent sources. `_sign_at_address_from_sd` reads it from the file on line 217 and uses it to build the address on line 223, but doesn't pass it down on line 225 — so this line re-derives it from the path instead.

Before this PR that didn't matter: the header was always 31–34 and asserted nothing, so a disagreement was invisible. Now the header is authoritative, and the two sources can contradict each other:

```
path                 file says   path implies   address displayed                            header
m/84h/0h/0h/0/0      p2wpkh      p2wpkh         bc1qcr8te4kr609gcawutmrza0j4xv80jy8z306fyu     39   ok
m/0h/0h/0h/0/0       p2wpkh      None→p2pkh     bc1qaaauaz73eyjyla73u7ahrx0vrlsgf0elwqrhyv     31   ✗
```

The second row doesn't need a malformed file. `_is_valid_derivation_path` only checks BIP32 syntax, so it accepts any purpose; `get_script_type_from_path` then finds no match in `SINGLESIG_SCRIPT_PURPOSE` (which only has 44/49/84/86), returns `None`, and the `or "p2pkh"` fallback here guesses p2pkh while the address on screen is a `bc1q…`. The `.txt` we export on lines 341–347 pairs that bech32 address with a header telling the verifier to recover a p2pkh one, so a strict verifier rejects it.

Repro — drop in `tests/`, passes on this branch:

```python
from .pages.home_pages.test_home import tdata
from .pages import create_ctx


def test_sd_script_type_mismatch(mocker, m5stickv, tdata):
    from krux.pages.home_pages.sign_message_ui import SignMessage
    from krux.wallet import Wallet
    from krux.input import BUTTON_ENTER

    # non-standard purpose; file declares p2wpkh
    file_content = b"msg\nm/0h/0h/0h/0/0\np2wpkh"

    ctx = create_ctx(mocker, [BUTTON_ENTER, BUTTON_ENTER],
                     Wallet(tdata.SINGLESIG_SIGNING_KEY))
    sig, _, address = SignMessage(ctx)._sign_at_address_from_sd(file_content)

    assert address.startswith("bc1q")   # p2wpkh address
    assert 31 <= sig[0] <= 34           # p2pkh header
```

Nobody loses funds over it, and the common case (Sparrow-generated files) is correct. But the PR does introduce a state where the screen shows one address and the signature claims another with no indication. Giving `_sign_at_address` an optional `script_type` and having the SD caller pass the one it already validated on line 218 — falling back to the path otherwise — removes the whole class in about three lines.

src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -248,6 +244,8 @@ def sign_standard_message(self, data):
             )
             if not self.prompt(t("Proceed?"), BOTTOM_PROMPT_LINE):
                 return ""
+        else:
+            message_hash = bip137.message_commitment(data)
```
joaozinhom, line 248: This overwrite makes the label on line 252 inaccurate — the screen still says `SHA256:` but now renders the BIP-137 commitment, which is `sha256(sha256(magic ‖ varint ‖ message))`. For `b"A test message."`:

```
sha256sum of the content : 60e76ce856c4bd873cc66e424d214ced480d77c34091aad5eb190b2cd9668bd8
what the screen shows    : 6d0485d73f55fe44f8040b0c5e818bebc8319ce3e61970e8f1fb2376858c47ac
```

That comparison is the user's only check that the device is signing what they think it is before pressing Sign, and the docs promise it explicitly — `docs/getting-started/usage/navigating-the-main-menu.en.md` line 330: *"you will be shown a preview of the message's SHA-256 hash for confirmation before signing"*. Someone verifying with `sha256sum` now sees a mismatch with no way to tell whether it's a bug, the wrong file, or intended behaviour.

It's also inconsistent within the same screen: on the `is_raw_hash` branch the label is still correct, since it shows the 32 bytes the user supplied. Same label, two different kinds of value depending on what was loaded.

Minor, related: `_compute_message_hash` still computes `hashlib.sha256(data).digest()` on line 277 and this line discards it two lines later. Folding the commitment into that helper would fix the dead computation and give one place to name the value.

src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -257,7 +255,13 @@ def sign_standard_message(self, data):
         if not self.prompt(t("Sign?"), BOTTOM_PROMPT_LINE):
             return ""
 
-        sig = self.ctx.wallet.key.sign(message_hash).serialize()
+        key = self.ctx.wallet.key
+        if is_raw_hash:
+            sig = key.sign(message_hash).serialize()
+        else:
+            _, sig = bip137.sign(
```
joaozinhom, line 262: Question rather than a change request: was the format change on this path intentional? The PR description is about the recid at derived addresses, and this is a separate behavioural change that isn't mentioned.

For non-raw-hash input we used to emit DER over `sha256(data)`; now it's a 65-byte BIP-137 compact signature. Decoding the old and new vectors for case 2 (`"hello world"`) in `tests/pages/home_pages/test_sign_message_ui.py` line 71:

```
before:  70 bytes, byte0 = 0x30   -> DER SEQUENCE
after:   65 bytes, byte0 = 0x28   -> BIP-137 header (39 + recid = p2wpkh)
```

Release signing is unaffected, which I checked: the firmware hash is 64 hex chars, so `_compute_message_hash` sets `is_raw_hash=True` and line 260 still produces DER.

The case that changes is signing a plaintext or binary file loaded from SD. That used to yield a signature accepted by `openssl sha256 <file -binary | openssl pkeyutl -verify -pubin -inkey pubkey.pem -sigfile sig.bin` — the exact command in `./krux verify` line 262 — using the hex public key we export right afterwards. A 65-byte compact signature isn't, and `navigating-the-main-menu.en.md` line 334 still tells users that exported key "can be used by others to verify your signature".

Worth noting too that the header carries no usable meaning here. `key.derivation` is the account node (`src/krux/key.py` lines 180–186), so a default singlesig wallet emits header 40, and a strict verifier would recover the pubkey and derive `bc1ql5f64jdzjsvgehlpxvdgm9ygp0xta7xpnueh03` — the p2wpkh address of the account node, which no wallet ever displays. For multisig/miniscript `key.script_type` is `p2wsh`, which falls through `build_header` and silently gets a p2pkh header.

If this is wanted, the docs need to follow. If it isn't, keeping BIP-137 to the at-address flow would scope the PR tightly to the bug it names.

joaozinhom (changes requested):
Left three inline comments two things I'd change plus a question about the standard-message format. A few small things below, none blocking.

`src/krux/bip137.py` line 47: the `compressed` parameter is unreachable in production nothing in krux passes `compressed=False`, only the tests do. It's also what makes the expression on lines 63–68 hard to read, for a branch that can't be taken. Since this is new API, easier not to ship it than to remove it later.
`src/krux/bip137.py` line 28: `RECID_OFFSET` isn't a recid offset, it's the compressed-pubkey offset (+4). Worth renaming now while the module is new.
`tests/test_bip137.py` lines 153 and 186: `print("Case: %d", i)` uses a comma instead of `%`, so it prints the literal `%d`. Line 230 has it right.

Claude helped me summarize everything and make my text clear.

src/krux/pages/home_pages/sign_message_ui.py
qlrd: Great finding @joaozinhom, i will try to fix this one.

src/krux/pages/home_pages/sign_message_ui.py
qlrd: Yup, this one have, IMO, a conceptual usage that is more a decision of UI/UX for pleb user flow than tech correctness: the majority of users will not understand the diff between `sha256` and `sha256(sha256(magic | varint | message))` defined in BIP137; i think those will add more information. 

The first path is conceptually lenient; should we accept that we could just call `sha256` for normal user even at cost of wrong concept and avoid more explanations (anyway, the user just wanna sign)?

The second path is conceptually strict; it would require add a little translation and diff in docs (what could be fixed in a follow-up if this path is chosen).

qlrd: > Minor, related: `_compute_message_hash` still computes `hashlib.sha256(data).digest()` on line 277 and this line discards it two lines later. Folding the commitment into that helper would fix the dead computation and give one place to name the value.

Yup, didn't touched this one i think because could be more than a `chore` or `refactor`. Maybe a suggestion or a followup?

joaozinhom: yeah i think this can be a follow up...

src/krux/pages/home_pages/sign_message_ui.py
qlrd: > Question rather than a change request: was the format change on this path intentional? The PR description is about the recid at derived addresses, and this is a separate behavioural change that isn't mentioned.

Yes, and actually this is what @odudex and @jdlcdl warned me about break firmware sign IIRC. Before i did like you described, since it's a separated behavioral change. But from what i understand, it touches a fundamental part of firmware and a firmware signature isn't the same thing as bip137 signature. Where you think should I point on docs your suggestion?

joaozinhom: docs/getting-started/usage/navigating-the-main-menu.en.md around line 330 — that's where I cited it. And also around line 334 where the openssl pkeyutl verification command is shown, since that command only works for DER over plain SHA-256, not a 65-byte BIP-137 compact signature.

qlrd: Yup, 

> DER over plain SHA-256, not a 65-byte BIP-137 compact signature.

This was the exact moral barrier to me 🤕

PR #874, opened 2026-06-08
Make settings reads non-mutating and harden malformed settings handling

kkdao:
### What is this PR for?
Makes the settings store safer and lighter. Store.get() reads without JSON-copying the whole settings tree, and a corrupted settings file is handled instead of crashing (at boot, on read, and on write, where it also repairs itself). Valid settings behave exactly the same.

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, build and tested on Maix Amigo

- [x] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other (small refactor of Store.get)

Testing: full test suite green (1089), lint 10/10. New tests fail on the old code. On the Amigo: settings save and load on flash and SD persist across restart, a corrupted settings file boots to defaults, and editing a setting with that corrupt file saves fine and repairs it.

### DIFF
--- a/src/krux/settings.py
+++ b/src/krux/settings.py
@@ -215,16 +215,17 @@ class Store:
             pass
 
     def get(self, namespace, setting_name, default_value):
-        """Returns a setting value under the given namespace, or default value if not set"""
-        s = json.loads(
-            json.dumps(self.settings)
-        )  # deepcopy to avoid building out namespaces
+        """Returns a setting value under the given namespace, or default value if not set.
+
+        Read-only walk: never mutates self.settings, so no defensive copy is
+        needed. A missing or non-dict intermediate namespace returns the default.
+        """
+        s = self.settings
         for level in namespace.split("."):
-            s[level] = s.get(level, {})
-            s = s[level]
-        if setting_name not in s:
-            return default_value
-        return s[setting_name]
+            s = s.get(level)
+            if not isinstance(s, dict):
+                return default_value
+        return s.get(setting_name, default_value)
 
     def set(self, namespace, setting_name, setting_value):
         """Stores a setting value under the given namespace if new/changed.
--- a/src/krux/settings.py
+++ b/src/krux/settings.py
@@ -170,11 +170,7 @@ class Store:
         self._load_settings()
 
         # Define location based on what was loaded or default undefined
-        self.file_location = (
-            self.settings.get("settings", {})
-            .get("persist", {})
-            .get("location", "undefined")
-        )
+        self.file_location = self._persisted_location("undefined")
 
         # Settings not found on SD, or 'persist.location' key not defined
         if SD_PATH not in self.file_location:
@@ -184,10 +180,25 @@ class Store:
 
         # Settings persist location will point to SD (if defined) else defaults to flash
         self.file_location = Store.get_vfs_location(
-            self.settings.get("settings", {})
-            .get("persist", {})
-            .get("location", FLASH_PATH)
+            self._persisted_location(FLASH_PATH)
         )
+
+    def _persisted_location(self, default):
+        """Reads settings.persist.location from a possibly-malformed settings dict.
+
+        The persisted file is only validated to be a top-level dict on load, so a
+        hand-edited/corrupted file may have a non-dict where a namespace is
+        expected. Walk defensively and fall back to the default instead of
+        raising during construction.
+        """
+        node = self.settings
+        for level in ("settings", "persist"):
+            if not isinstance(node, dict):
+                return default
+            node = node.get(level)
+        if not isinstance(node, dict):
+            return default
+        return node.get("location", default)
 
     @classmethod
     def get_vfs_location(cls, location):
--- a/src/krux/settings.py
+++ b/src/krux/settings.py
@@ -184,13 +184,9 @@ class Store:
         )
 
     def _persisted_location(self, default):
-        """Reads settings.persist.location from a possibly-malformed settings dict.
-
-        The persisted file is only validated to be a top-level dict on load, so a
-        hand-edited/corrupted file may have a non-dict where a namespace is
-        expected, or a bogus location value. Walk defensively and return only a
-        known location (SD_PATH/FLASH_PATH); otherwise fall back to the default.
-        """
+        """Reads persist.location defensively; returns a known path or default."""
+        # A corrupted/hand-edited file may have a non-dict at any namespace level
+        # or a bogus location value — walk defensively and validate before returning.
         node = self.settings
         for level in ("settings", "persist"):
             if not isinstance(node, dict):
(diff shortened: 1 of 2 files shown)


src/krux/settings.py
```diff
@@ -184,11 +180,29 @@ def __init__(self):
...
+            if not isinstance(node, dict):
+                return default
+            node = node.get(level)
+        if not isinstance(node, dict):
+            return default
+        location = node.get("location", default)
+        if location not in (SD_PATH, FLASH_PATH):
+            return default
+        return location
+
```
qlrd, line 201: The first phrase is good. Longer docstrings can end up in firmware via `__doc__` and add a little flash (AFAIK — not sure exactly how MicroPython strips them).

```suggestion
    def _persisted_location(self, default):
        """Reads settings.persist.location from a possibly-malformed settings dict."""
        # The persisted file is only validated to be a top-level dict on load, so a
        # hand-edited/corrupted file may have a non-dict where a namespace is
        # expected, or a bogus location value. Walk defensively and return only a
        # known location (SD_PATH/FLASH_PATH); otherwise fall back to the default.
        node = self.settings
        for level in ("settings", "persist"):
            if not isinstance(node, dict):
                return default
            node = node.get(level)
        if not isinstance(node, dict):
            return default
        location = node.get("location", default)
        if location not in (SD_PATH, FLASH_PATH):
            return default
        return location

```

src/krux/settings.py
```diff
@@ -215,24 +229,30 @@ def _load_settings(self):
...
     def get(self, namespace, setting_name, default_value):
-        """Returns a setting value under the given namespace, or default value if not set"""
-        s = json.loads(
-            json.dumps(self.settings)
-        )  # deepcopy to avoid building out namespaces
+        """Returns a setting value under the given namespace, or default value if not set.
+
+        Read-only walk: never mutates self.settings, so no defensive copy is
+        needed. A missing or non-dict intermediate namespace returns the default.
+        """
```
qlrd:
```suggestion
        """Returns a setting value under the given namespace, or default value if not set."""
        # Read-only walk: never mutates self.settings, so no defensive copy is
        # needed. A missing or non-dict intermediate namespace returns the default.
```

src/krux/settings.py
```diff
@@ -215,24 +229,30 @@ def _load_settings(self):
...
+                return default_value
+        return s.get(setting_name, default_value)
 
     def set(self, namespace, setting_name, setting_value):
         """Stores a setting value under the given namespace if new/changed.
         Does NOT automatically save settings to flash or sd!
+
+        A non-dict intermediate level (from a corrupted/hand-edited file) is
+        replaced with a fresh dict, so writing a setting cannot raise and also
+        repairs the malformed structure.
```
qlrd:
```suggestion
        """Stores a setting value under the given namespace if new/changed.
        Does NOT automatically save settings to flash or sd!
        """
        # A non-dict intermediate level (from a corrupted/hand-edited file) is
        # replaced with a fresh dict, so writing a setting cannot raise and also
        # repairs the malformed structure.
```

src/krux/settings.py
```diff
@@ -184,11 +180,29 @@ def __init__(self):
...
+    def _persisted_location(self, default):
+        """Reads settings.persist.location from a possibly-malformed settings dict.
+
+        The persisted file is only validated to be a top-level dict on load, so a
+        hand-edited/corrupted file may have a non-dict where a namespace is
+        expected, or a bogus location value. Walk defensively and return only a
+        known location (SD_PATH/FLASH_PATH); otherwise fall back to the default.
+        """
+        node = self.settings
+        for level in ("settings", "persist"):
```
qlrd, line 191: Optional nit: those strings could be a module-level constant, e.g.

```suggestion
SETTINGS_PERSIST_LEVELS = ("settings", "persist")
```

and then `for level in SETTINGS_PERSIST_LEVELS:` — note **`persist`**, not `persisted`. Non-blocking; current inline tuple is fine too.

qlrd (changes requested):
cACK with a warning that i'm not suitable to evaluate if it make settings lighter and safer (i would avoid assertive until a better review). My role on this review is just some style nits. 

First is the conventional commtis adopted on `develop` branch. IMO the best path for a reviewer, is to separate tests into one `fix!` (with `!` IMO because we don't know yet if this could break with the things, e.g., testing a loaded settings in a older firmware), and one `test` (and optionally a `docs` with a change on `CHANGELOG.md`).

IMO the `fix`/`fix!` commit message (and PR description) could be: 

> `Store.get()` reads without JSON-copying the whole settings tree, and a corrupted settings file is handled instead of crashing (at boot, on read, and on write, where it also repairs itself), saving and loading on flash and SD persist across restart, a corrupted settings file boots to defaults, and editing a setting with that corrupt file saves fine and repairs it.

If you choose `fix!`, recommend to address it on a follow-up. I encourage too to open a issue (describing what lead you to write the PR before write code), so we can link the issue with a `Fix #<Number of Issue>` at the end of commit message, have a good tracing history about firmware (and close automatically them when the PR merges -- also `git absorb` get its strenght on these conventional commits, i recommend).

The nits:

jdlcdl:
> if it make settings lighter and safer

After reviewing changes, I believe that these claims are valid.

While I have not verified what happens w/o this PR, I believe that a corrupted settings.json would either "boom" or at least raise an error from json.loads() (and hopefully settings structure would be defaulted).  This PR is making much more effort to validate stored settings not only for a valid json structure (the style of validation w/o this pr) but for meaningful settings values, with testing that proves old code fails what is likely to occur if user edits json by hand or plays with flashing different versions versions where settings structure changes (easy thanks to krux-installer).

I believe this PR is:
* safer: (for non-json, for valid json but values which are bogus, fixes the problem instead of leaving room to raise an error),
* lighter: at the very least, a json.loads(json.dump(settings-data)) round trip is avoided (it was a deepcopy trick), and potential error messages (if not crashes) would be avoided for the user.

odudex (approved):
Thanks @kkdao ! Consistent improvement is made from small optimizations like this.

Will just wait @jdlcdl 's device tests to merge.

jdlcdl:
ACK tested at f2bc6e3 on "amigo" and "yahboom_devkit".

I'll note that behavior is nice, in that if user were to edit settings to bogus values... not the entire settings would get defaulted, only the settings which are bogus.  However, if the settings.json file is no longer valid json, then a full set of defaults are loaded and would overwrite the corrupted settings.json -- losing some custom values (arguably safe to correct a broken json file).

I'll approve but it's just my opnion.  @qlrd's inputs are appropriate so do see them @kkdao

jdlcdl (approved):

src/krux/settings.py
kkdao: Done, thank you for reviewing thoroughly!

PR #873, opened 2026-06-07
Harden docs against supply-chain attacks: remove all third-party assets

odudex:
### What is this PR for?
Following the recent polyfill.io incident (a CDN extension injected malware into the docs), this removes **every** remote asset the docs load in visitors' browsers and adds a CSP so it can't recur.

### Changes
- **Math** — pre-render the entropy page's display equations to committed local SVGs (inline math → HTML); drop the cdnjs MathJax + require.js loads and `pymdownx.arithmatex`. `render_docs_math.py` regenerates the SVGs from the markdown (dev-only matplotlib, no JS).
- **Fonts** — use the system font stack (`theme.font: false`) instead of Google Fonts.
- **Video tutorials** — title + link list instead of YouTube thumbnails.
- **Emoji** — inline Material icons in place of unicode shortcode emoji (which load from the twemoji CDN).
- **Iframes** — remove the gitmind.com mind-map embeds (and the navigation page that only held them).
- **CSP** — add a `Content-Security-Policy` meta tag (`overrides/main.html`) locking scripts/styles/fonts/frames to `'self'`, blocking any external `<script>` at the browser level.
- gitignore `public/` and `htmlcov/` build output.

### Changes made to:
- [ ] Code
- [ ] Tests
- [X] Docs
- [ ] CHANGELOG

### Did you build the code and tested on device?
- Firmware not affected

### What is the purpose of this pull request?
- [ ] Bug fix
- [ ] New feature
- [X] Docs update
- [ ] Other

PR #875, opened 2026-06-08
docs: serve templates page at folder URL via index.md

odudex:
### What is this PR for?
Fix odd image paths issue on templates docs page pointed by @jdlcdl 
Rename templates.en.md to index.en.md so the page is served at the folder URL, fixing the confusing leading '..' in image paths. Wrap it in its own nav section so navigation.indexes labels it 'Templates' instead of absorbing it into 'Interface'.

### Changes made to:
- [ ] Code
- [ ] Tests
- [x] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [x] Docs update
- [ ] Other


jdlcdl:
ACK tested at 625079a

This PR matches similar changes I'd tested when originally trying to resolve this WARNING (noted in #873)... and also covers the krux binary grids and tinyseed templates that I had not noticed (nice catch!).

As well, "index" instead of "templates" is a much better name.  (I did compare templates.en.md and index.en.md files to make sure that changes were ONLY the removal of "../" where necessary).

jdlcdl (approved):

PR #877, opened 2026-06-09
fix: narrow bare except blocks to stop hiding real errors

kkdao:
### What is this PR for?

Replaces five bare `except:` blocks with more specific exceptions, so real errors and signals like `KeyboardInterrupt` are not silently swallowed.

Changed:
- `display_qr_codes` catches only `StopIteration`
- `Key.extract_fingerprint` catches `Exception`
- `qr.py` catches `IndexError` / `Exception`
- `has_sd_card` catches `Exception`

The behavior-changing cases have regression tests that fail on the old code. `max_qr_bytes` keeps the same behavior and now has fallback coverage.

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, with the tzt device. Tested animated QR, QR scan (valid + invalid), SD detection, and fingerprint all verified, no crashes.

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other — code quality / robustness

### DIFF
--- a/src/krux/key.py
+++ b/src/krux/key.py
@@ -202,7 +202,7 @@ class Key:
                 Key.extract_root(mnemonic, passphrase, network).child(0).fingerprint,
                 pretty,
             )
-        except:
+        except Exception:
             pass
         return ""
 
--- a/src/krux/pages/__init__.py
+++ b/src/krux/pages/__init__.py
@@ -294,7 +294,7 @@ class Page:
         while not done:
             try:
                 code, num_parts = next(code_generator)
-            except:
+            except StopIteration:
                 code_generator = to_qr_codes(data, qr_data_width, qr_format)
                 code, num_parts = next(code_generator)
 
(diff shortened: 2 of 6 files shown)


jdlcdl:
Wow, I never realized there was a difference between "except" and "except Exception".

As a rule-of-thumb in the far past for other projects, I prefer to have very precise exception handling (more than base Exception) and for everything else to "Boom!" as an uncaught error so that it's obvious.  But this certainly deserves more discussion with others as it's a common krux pattern to just "except:".  If bare except is a bad idea, maybe the entire codebase needs to be checked/fixed.

qlrd:
cACK with a context that i din't read the code (yet) but it's sounds good to me on `linter` aspect (that `pyright` showed them many yellow points on my editor but `pylint` not).

odudex:
This must be tested in device, especially animated QR code generation, was it tested? The PR description is not following the standard checks.

kkdao:
Updated PR description. Won't happen again.

odudex:
Thank you!

PR #878, opened 2026-06-13
refactor: remove duplicated work in display orientation/init paths

kkdao:
### What is this PR for?

Cleanup-only refactor of the display orientation/init paths. Behavior is unchanged.

- Read Settings() once in the orientation/init paths instead of rebuilding it repeatedly.
- Reuse the existing is_flipped_orientation helper instead of a duplicated check.

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, built and tested on the Yahboom. Boots normally. Display flips 180 degrees and reverts back correctly.

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/src/krux/krux_settings.py
+++ b/src/krux/krux_settings.py
@@ -502,8 +502,8 @@ class Settings(SettingsNamespace):
 
     def is_flipped_orientation(self):
         """Returns flipped orientation setting"""
-        return hasattr(Settings().hardware, "display") and getattr(
-            Settings().hardware.display, "flipped_orientation", False
+        return hasattr(self.hardware, "display") and getattr(
+            self.hardware.display, "flipped_orientation", False
         )
 
     def label(self, attr):
--- a/src/krux/display.py
+++ b/src/krux/display.py
@@ -244,20 +244,14 @@ class Display:
     def to_landscape(self):
         """Changes the rotation of the display to landscape"""
         if self.portrait:
-            hardware = Settings().hardware
-            flipped = hasattr(hardware, "display") and getattr(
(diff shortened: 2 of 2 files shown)


src/krux/krux_settings.py
```diff
@@ -502,8 +502,8 @@ def __init__(self):
 
     def is_flipped_orientation(self):
         """Returns flipped orientation setting"""
-        return hasattr(Settings().hardware, "display") and getattr(
-            Settings().hardware.display, "flipped_orientation", False
+        return hasattr(self.hardware, "display") and getattr(
+            self.hardware.display, "flipped_orientation", False
```
qlrd, line 506: This is the part of i have more doubts. AFAIK, `Settings()` was a style used from the beggining.

IMO, this part will require both @odudex, @jdlcdl.

qlrd (commented):
LGTM to `lint` and `format` aspect (coherent in the sense that the refactor remove more lines than adding). But i have a concern that put me in a block and need other opinions

src/krux/krux_settings.py
jdlcdl: IMO, this seems fine.  I understand both forms doing the same thing, however the newer form doesn't need to re-instantiate and initialize another Settings() object, since it can use itself.

odudex: Yes, it's fine and good!

odudex:
Thank you! The same optimization can be done in `touch.py`:96 and `touch.py`:192.
It's worth a follow-up PR!

PR #881, opened 2026-06-17
refactor: dedupe flipped-orientation check in touch.py via is_flipped_orientation

kkdao:
### What is this PR for?

Follow-up to #878. Applies the same optimization in touch.py: replaces the duplicated inline flipped-orientation check in valid_position() and _store_points() with the existing Settings().is_flipped_orientation() helper. 

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, built and tested on Yahboom

Taps land correctly, and swipes work in both normal and flipped orientation.

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other (refactor / dedup)

### DIFF
--- a/src/krux/touch.py
+++ b/src/krux/touch.py
@@ -93,9 +93,7 @@ class Touch:
     def valid_position(self, data):
         """Checks if touch position is within buttons area"""
 
-        if hasattr(Settings().hardware, "display") and getattr(
-            Settings().hardware.display, "flipped_orientation", False
-        ):
+        if Settings().is_flipped_orientation():
             data = (self.height - data[0], self.width - data[1])
 
         if self.x_regions and data[0] < self.x_regions[0]:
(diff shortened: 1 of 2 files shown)


jdlcdl:
ACK tested at c8c9169 and yahboom_devkit (and simulator which, expectedly?!?!, doesn't flip screen but does flip controls).

On this pr and others I've recently had problems getting past "poe vulture" checks.  But recently I tested pr_881 which uses "astral.sh's uv" instead of "poetry" as a virtualenv manager.  Since uv updated my environment, I was able to switch back to this pr branch, without re-installing poetry and mucking with the virtualenv, and now I get past vulture checks on this pr, branch "develop", and I suspect other branches as well.

jdlcdl (approved):

odudex:
Thank you!

PR #882, opened 2026-06-17
perf: cache Settings() namespace tree as a singleton

odudex:
### What is this PR for?
Following @kkdao 's settings optimizations, this one cache Settings() namespace tree as a singleton and avoids rebuilding multiple namespace objects on every Settings() call

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, build and tested on Amigo and WonderMV

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/src/krux/krux_settings.py
+++ b/src/krux/krux_settings.py
@@ -490,8 +490,19 @@ class Settings(SettingsNamespace):
     """The top-level settings namespace under which other namespaces reside"""
 
     namespace = "settings"
+    _instance = None
+
+    def __new__(cls):
+        # Cache the namespace tree: values are read live from the `store`
+        # singleton, so reusing the instance avoids rebuilding ~16 objects
+        # on every Settings() call without staling any setting.
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
 
     def __init__(self):
+        if getattr(self, "_built", False):
+            return
         self.wallet = DefaultWallet()
         self.security = SecuritySettings()
         self.hardware = HardwareSettings()
@@ -499,6 +510,7 @@ class Settings(SettingsNamespace):
         self.encryption = EncryptionSettings()
         self.persist = PersistSettings()
         self.appearance = ThemeSettings()
+        self._built = True
 
     def is_flipped_orientation(self):
         """Returns flipped orientation setting"""
--- a/tests/pages/test_login.py
+++ b/tests/pages/test_login.py
@@ -10,6 +10,7 @@ def mocker_printer(mocker):
 @pytest.fixture
 def mock_retro_compatibility(mocker, amigo):
     from krux.settings import CategorySetting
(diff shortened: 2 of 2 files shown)


qlrd (approved):
ACK 5cfdaba

src/krux/krux_settings.py
```diff
@@ -490,15 +490,27 @@ class Settings(SettingsNamespace):
     """The top-level settings namespace under which other namespaces reside"""
 
     namespace = "settings"
+    _instance = None
+
+    def __new__(cls):
```
jdlcdl, line 495: Unsure: @classmethod decorator, or not, above this overridden method? and how it applies to micropython.

odudex: `__new__` is a special case in Python: the interpreter implicitly treats it as a static method, even though its first parameter is conventionally named `cls`. From what I saw, should be written bare, both in CPython and MicroPython.

jdlcdl:
Thank you for explaining the one comment I had, I have resolved that comment and tested on amigo's terminal with compiler flag turned on... and am convinced that you're right.

ACK tested at 5cfdaba on amigo

jdlcdl (approved):

src/krux/krux_settings.py
qlrd: Made a simple test [here](https://gist.github.com/qlrd/cac09c083e80b9b0174a1be31a7cc9f3) with @odudex explanation and got this result. If @classmethod is used, the CI could raise exception from what i understood (while in micropython it works without issues). Also saw that the last tested case is considered a issue shadowing (since the interpreter implicity treat as staticmethod, so it "binds twice instead one" the class, and if you ommit *args, it raise exception because expect 2 args):

**CPython**:

```bash
$> uv run python verify-decorators.py
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.12`.
verify __new__ decorators
Runtime: 3.12.13 (main, May 30 2026, 15:12:00) [Clang 17.0.0 (clang-1700.0.13.5)]

[1] bare __new__(cls)  PASS

[2] @classmethod __new__(cls)
      FAIL  TypeError: T.__new__() takes 1 positional argument but 2 were given

[3] @classmethod __new__(cls, *args, **kwargs)  PASS

------------------------------------------------------------
Summary: 2 PASS, 1 FAIL
```

**Micropython**:

```bash
$> uv run micropython verify-decorators.py
warning: No `requires-python` value found in the workspace. Defaulting to `>=3.12`.

verify __new__ decorators
Runtime: 3.4.0

[1] bare __new__(cls)  PASS

[2] @classmethod __new__(cls)  PASS

[3] @classmethod __new__(cls, *args, **kwargs)  PASS

------------------------------------------------------------
Summary: 3 PASS, 0 FAIL
```

PR #885, opened 2026-06-18
fix: handle encrypted mnemonic storage edge cases

kkdao:
### What is this PR for?

Hardens encrypted mnemonic storage so missing IDs, malformed `seeds.json`, or unexpected non-object storage data do not crash encrypted mnemonic load/store/decrypt flows.

Changed:
- Missing encrypted mnemonic IDs now return `None` instead of raising.
- Malformed storage files are treated as empty storage and can be repaired on write.
- Valid but unexpected non-object `seeds.json` data is treated as empty storage.
- Storage file reads now catch only expected file/JSON errors.

The behavior-changing cases have regression tests that fail on the old code.

### Changes made to:
- [x] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, with the tzt. Tested encrypted mnemonic store/load on flash and SD, wrong key failure, correct key recovery, duplicate ID refusal, and no crashes. `uv run poe pre-commit` passed, 1113 tests.

- [x] Bug fix
- [ ] New feature
- [ ] Docs update
- [ ] Other

### DIFF
--- a/src/krux/encryption.py
+++ b/src/krux/encryption.py
@@ -36,18 +36,23 @@ QR_CODE_ITER_MULTIPLE = 10000
 class MnemonicStorage:
     """Handler of stored encrypted seeds"""
 
+    @staticmethod
+    def _load_mnemonics(contents):
+        mnemonics = json.loads(contents)
+        return mnemonics if isinstance(mnemonics, dict) else {}
+
     def __init__(self) -> None:
         self.stored = {}
         self.stored_sd = {}
         try:
             with SDHandler() as sd:
-                self.stored_sd = json.loads(sd.read(MNEMONICS_FILE))
+                self.stored_sd = self._load_mnemonics(sd.read(MNEMONICS_FILE))
         except (OSError, ValueError):
             # missing/unreadable SD card or malformed JSON -> start empty
             pass
         try:
             with open(FLASH_PATH_STR % MNEMONICS_FILE, "r") as f:
-                self.stored = json.loads(f.read())
+                self.stored = self._load_mnemonics(f.read())
         except (OSError, ValueError):
             # missing/unreadable flash file or malformed JSON -> start empty
             pass
@@ -89,12 +94,10 @@ class MnemonicStorage:
 
     def decrypt(self, key, mnemonic_id, sd_card=False):
         """Decrypt a selected encrypted mnemonic from a file"""
-        try:
-            if sd_card:
-                stored_value = self.stored_sd.get(mnemonic_id)
-            else:
-                stored_value = self.stored.get(mnemonic_id)
-        except:
+        source = self.stored_sd if sd_card else self.stored
+        stored_value = source.get(mnemonic_id) if isinstance(source, dict) else None
+        if not isinstance(stored_value, dict):
+            # unknown id, or a corrupt/non-dict storage entry -> nothing to decrypt
             return None
 
         if stored_value.get("b64_kef"):
@@ -122,7 +125,7 @@ class MnemonicStorage:
                 with SDHandler() as sd:
                     contents = sd.read(MNEMONICS_FILE)
                     orig_len = len(contents)
-                    mnemonics = json.loads(contents)
+                    mnemonics = self._load_mnemonics(contents)
             except (OSError, ValueError):
                 # no existing/readable file or malformed JSON -> write fresh
                 orig_len = 0
@@ -143,7 +146,7 @@ class MnemonicStorage:
             try:
                 # load current MNEMONICS_FILE
                 with open(FLASH_PATH_STR % MNEMONICS_FILE, "r") as f:
-                    mnemonics = json.loads(f.read())
+                    mnemonics = self._load_mnemonics(f.read())
             except (OSError, ValueError):
                 # no existing/readable file or malformed JSON -> write fresh
                 pass
--- a/src/krux/encryption.py
+++ b/src/krux/encryption.py
@@ -38,8 +38,7 @@ class MnemonicStorage:
 
     @staticmethod
     def _load_mnemonics(contents):
-        mnemonics = json.loads(contents)
-        return mnemonics if isinstance(mnemonics, dict) else {}
+        return json.loads(contents)
 
     def __init__(self) -> None:
         self.stored = {}
--- a/src/krux/encryption.py
+++ b/src/krux/encryption.py
@@ -33,6 +33,11 @@ FLASH_PATH_STR = "/" + FLASH_PATH + "/%s"
 QR_CODE_ITER_MULTIPLE = 10000
 
 
+class StorageCorruptedError(Exception):
+    """Stored mnemonics file exists but is not a valid object; it is left
+    untouched instead of being overwritten, so its data can be recovered."""
+
+
 class MnemonicStorage:
     """Handler of stored encrypted seeds"""
 
@@ -85,11 +90,11 @@ class MnemonicStorage:
 
     def list_mnemonics(self, sd_card=False):
         """List all seeds stored on a file"""
-        mnemonic_ids = []
         source = self.stored_sd if sd_card else self.stored
-        for mnemonic_id in source:
-            mnemonic_ids.append(mnemonic_id)
-        return mnemonic_ids
+        if not isinstance(source, dict):
+            # corrupt/non-dict storage -> nothing to list
+            return []
+        return list(source)
 
     def decrypt(self, key, mnemonic_id, sd_card=False):
         """Decrypt a selected encrypted mnemonic from a file"""
@@ -120,14 +125,21 @@ class MnemonicStorage:
         mnemonics = {}
         if sd_card:
             # load current MNEMONICS_FILE
+            orig_len = 0
             try:
                 with SDHandler() as sd:
                     contents = sd.read(MNEMONICS_FILE)
                     orig_len = len(contents)
                     mnemonics = self._load_mnemonics(contents)
-            except (OSError, ValueError):
-                # no existing/readable file or malformed JSON -> write fresh
-                orig_len = 0
+            except OSError:
+                # missing file -> write a fresh store
+                pass
+            except ValueError as exc:
+                # corrupt JSON -> preserve for recovery
+                raise StorageCorruptedError(MNEMONICS_FILE) from exc
+            if not isinstance(mnemonics, dict):
+                # wrong shape -> preserve for recovery
+                raise StorageCorruptedError(MNEMONICS_FILE)
 
             # save the new MNEMONICS_FILE
             try:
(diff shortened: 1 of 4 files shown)


jdlcdl:
Would a corrupted seeds.json file on sdcard or in flash, with some valid entries but unparseable as json because of other corruption, result in overwriting otherwise valid encrypted mnemonic entries?  My worry would be someone mucking with their seeds.json on sdcard from their computer's editor (in write mode instead of read-only) and then later losing access to all entries in seeds.json because they exited-and-saved without noticing they'd broken json formatting.

Perhaps at least a warning when reading a corrupted seeds.json file (though no entries should be rather alarming to the user expecting their mnemonics) is justified so that user does not blindly save another mnemonic overwriting the entire file (which could likely be "mostly" recovered even if corrupted)

kkdao:
Good catch, that is a real data-loss path. Scope-wise though, this PR only narrows the exception types. For unparseable JSON the behavior is unchanged from `develop` since `ValueError` was already caught and already fell back to writing fresh. I agree a warning/refusal-to-overwrite for corrupt `seeds.json` would be a good safety improvement, but that’s a storage/UI behavior change and deserves its own issue/PR with device testing - I think. Happy to open that follow-up.

odudex:
I think @jdlcdl is right, we shouldn't change the behavior and silently remove otherwise recoverable data without a prompt warning.

Other nit, on line 94 `if isinstance(source, dict) else None` seems to be a redundant check, dead code.

kkdao:
Dropped the non-dict JSON coercion so a wrong-shape `seeds.json` is no longer silently treated as empty on the next store, matching `develop`. Kept the `isinstance(source, dict)` guard in `decrypt()`, since once the coercion is gone it’s load-bearing and prevents wrong-shape storage from crashing on `.get`. Removed the 4 coercion tests as rest of the suite is green. Tested again SD store/load round-trip, no SD store, and corrupted `seeds.json` handling. Corrupt-file warning stays as separate follow-up.

src/krux/encryption.py
```diff
@@ -73,6 +79,8 @@ def stretch_key(key, salt, iterations):
             plaintext = kef._unpad(decryptor.decrypt(payload), pkcs_pad=False)
             return plaintext.decode()
         except:
+            # broad on purpose: any failure here means a wrong key or
+            # incompatible legacy ciphertext -> return None
```
qlrd, line 88: here, but maybe deserves a better explanations, since touch the PR purpose

src/krux/encryption.py
```diff
@@ -36,18 +36,24 @@
...
+        except (OSError, ValueError):
+            # missing/unreadable SD card or malformed JSON -> start empty
             pass
         try:
             with open(FLASH_PATH_STR % MNEMONICS_FILE, "r") as f:
-                self.stored = json.loads(f.read())
-        except:
+                self.stored = self._load_mnemonics(f.read())
+        except (OSError, ValueError):
+            # missing/unreadable flash file or malformed JSON -> start empty
```
qlrd, line 61: here

src/krux/encryption.py
```diff
@@ -36,18 +36,24 @@
...
     def __init__(self) -> None:
         self.stored = {}
         self.stored_sd = {}
         try:
             with SDHandler() as sd:
-                self.stored_sd = json.loads(sd.read(MNEMONICS_FILE))
-        except:
+                self.stored_sd = self._load_mnemonics(sd.read(MNEMONICS_FILE))
+        except (OSError, ValueError):
+            # missing/unreadable SD card or malformed JSON -> start empty
```
qlrd, line 55: here

src/krux/encryption.py
```diff
@@ -85,12 +93,10 @@ def list_mnemonics(self, sd_card=False):
...
-        try:
-            if sd_card:
-                stored_value = self.stored_sd.get(mnemonic_id)
-            else:
-                stored_value = self.stored.get(mnemonic_id)
-        except:
+        source = self.stored_sd if sd_card else self.stored
+        stored_value = source.get(mnemonic_id) if isinstance(source, dict) else None
+        if not isinstance(stored_value, dict):
+            # unknown id, or a corrupt/non-dict storage entry -> nothing to decrypt
```
qlrd, line 104: here, but maybe deserves a better explanations, since touch the PR purpose

src/krux/encryption.py
```diff
@@ -118,8 +124,9 @@ def store_encrypted_kef(self, mnemonic_id, kef_envelope, sd_card=False):
                 with SDHandler() as sd:
                     contents = sd.read(MNEMONICS_FILE)
                     orig_len = len(contents)
-                    mnemonics = json.loads(contents)
-            except:
+                    mnemonics = self._load_mnemonics(contents)
+            except (OSError, ValueError):
+                # no existing/readable file or malformed JSON -> write fresh
```
qlrd: here

src/krux/encryption.py
```diff
@@ -132,20 +139,23 @@ def store_encrypted_kef(self, mnemonic_id, kef_envelope, sd_card=False):
                         contents += " " * (orig_len - len(contents))
                     sd.write(MNEMONICS_FILE, contents)
             except:
+                # broad on purpose: any failure to save means the store failed
```
qlrd, line 154: here

src/krux/encryption.py
```diff
@@ -132,20 +139,23 @@ def store_encrypted_kef(self, mnemonic_id, kef_envelope, sd_card=False):
...
                 return False
         else:
             try:
                 # load current MNEMONICS_FILE
                 with open(FLASH_PATH_STR % MNEMONICS_FILE, "r") as f:
-                    mnemonics = json.loads(f.read())
-            except:
+                    mnemonics = self._load_mnemonics(f.read())
+            except (OSError, ValueError):
+                # no existing/readable file or malformed JSON -> write fresh
```
qlrd: here

src/krux/encryption.py
```diff
@@ -132,20 +139,23 @@ def store_encrypted_kef(self, mnemonic_id, kef_envelope, sd_card=False):
...
                 pass
             try:
                 # save the new MNEMONICS_FILE
                 with open(FLASH_PATH_STR % MNEMONICS_FILE, "w") as f:
                     mnemonics[mnemonic_id] = {"b64_kef": b64_kef}
                     f.write(json.dumps(mnemonics))
             except:
+                # broad on purpose: any failure to save means the store failed
```
qlrd, line 176: here

tests/test_encryption.py
```diff
@@ -612,3 +612,216 @@ def test_customize_pbkdf2_iterations_create_and_decode(m5stickv):
...
+# expect (OSError, ValueError), matching the OSError convention already used in
+# sd_card.py. The behaviour for a missing/unreadable file or malformed JSON is
+# unchanged ("storage starts empty" / "first store still writes"); the change
+# is that an *unexpected* error is no longer silently swallowed -- it now
+# propagates, so real bugs stop hiding. Each "propagates_unexpected_error" test
+# is the one that fails on the old bare-except code.
+# ---------------------------------------------------------------------------
+
+
+# --- __init__ SD load (self.stored_sd) ---
```
qlrd, line 630: IMO this is already explained on commit messages, no need to repeat it

tests/test_encryption.py
```diff
@@ -612,3 +612,216 @@ def test_customize_pbkdf2_iterations_create_and_decode(m5stickv):
...
+    sd.read.return_value = "not valid json {{{"
+    sdhandler = mocker.MagicMock()
+    sdhandler.return_value.__enter__.return_value = sd
+    mocker.patch("krux.encryption.SDHandler", new=sdhandler)
+    with patch("krux.encryption.open", new=mocker.mock_open(read_data="{}")):
+        storage = MnemonicStorage()
+    assert storage.stored_sd == {}
+
+
+# --- __init__ flash load (self.stored) ---
```
qlrd, line 664: No need

tests/test_encryption.py
```diff
@@ -612,3 +612,216 @@ def test_customize_pbkdf2_iterations_create_and_decode(m5stickv):
...
+
+
+def test_init_flash_load_propagates_unexpected_error(m5stickv, mocker):
+    from krux.encryption import MnemonicStorage
+
+    # SD load fails with an expected error so only the flash load can raise.
```
qlrd, line 670: IMO deserves a nice explanation

tests/test_encryption.py
```diff
@@ -612,3 +612,216 @@ def test_customize_pbkdf2_iterations_create_and_decode(m5stickv):
...
+
+    mocker.patch("krux.encryption.SDHandler", side_effect=OSError)
+    with patch(
+        "krux.encryption.open", new=mocker.mock_open(read_data="not valid json {{{")
+    ):
+        storage = MnemonicStorage()
+    assert storage.stored == {}
+
+
+# --- store_encrypted_kef SD read-before-write ---
```
qlrd, line 697: no need

tests/test_encryption.py
```diff
@@ -612,3 +612,216 @@ def test_customize_pbkdf2_iterations_create_and_decode(m5stickv):
...
+    mocker.patch("krux.sd_card.SDHandler.read", return_value="not valid json {{{")
+    with patch("krux.sd_card.open", new=mocker.mock_open(read_data="{}")) as m:
+        success = storage.store_encrypted_kef(
+            "KEFecbID", KEF_ENVELOPE_ECB, sd_card=True
+        )
+    assert success is True
+    m().write.assert_called_once_with(KEF_ECBENTROPY_ONLY_JSON)
+
+
+# --- store_encrypted_kef flash read-before-write ---
```
qlrd, line 760: no need

tests/test_encryption.py
```diff
@@ -612,3 +612,216 @@ def test_customize_pbkdf2_iterations_create_and_decode(m5stickv):
...
+    assert success is True
+    write_handle().write.assert_called_once_with(KEF_ECBENTROPY_ONLY_JSON)
+
+
+# ---------------------------------------------------------------------------
+# decrypt() must not crash on a missing id or non-dict storage.
+#
+# storage.get(id) returns None for an unknown id. decrypt() should return None
+# instead of raising AttributeError when there is no stored entry.
+# ---------------------------------------------------------------------------
```
qlrd, line 821: IMO already explained on c6422ff888c3f4f9ef6b085f391a8be8bb2c1696

qlrd (commented):
LGTM to me on general, so i will just made a manual `lints` :)

Some manual lecture and non-blockable nits (mostly unnecessary, IMHO, comments that are already explained in the commit messages or in the code itself). 

Others i think it deserves a better explanation.

odudex:
Opened a PR to you to preserve eventual corrupt seeds.json instead of overwriting on store:
https://github.com/kkdao/krux/pull/3

src/krux/encryption.py
qlrd:
```suggestion
            # any failure here means a wrong key or
            # incompatible legacy ciphertext -> return None
```

src/krux/encryption.py
qlrd:
```suggestion
                # any failure to save means the store failed
```

src/krux/encryption.py
qlrd:
```suggestion
                # any failure to save means the store failed
```

tests/test_encryption.py
```diff
@@ -726,21 +726,35 @@ def test_store_sd_read_oserror_still_writes(m5stickv, mocker, mock_file_operatio
...
+
+
+def test_store_sd_read_non_dict_json_raises_and_preserves(
+    m5stickv, mocker, mock_file_operations
+):
+    from krux.krux_settings import Settings
+    from krux.encryption import MnemonicStorage, StorageCorruptedError
+
+    storage = MnemonicStorage()
+    Settings().encryption.version = "AES-ECB"
```
qlrd, line 752: just thinking (need to a deep stydy on kef): it return same results with other versions when possible?

qlrd:
(non-blocking suggestions)

jdlcdl (approved):
ACK tested on simulator and Amigo.  I think users will appreciate the warning and chance to recover a corrupted seeds.json instead of thinking it was blank and over-writing possibly-recoverible file with new entries.

odudex:
Thank you!

PR #887, opened 2026-06-21
fix: Bug. PSBTv2 support in trim()

Garrysoon:
## Problem

When signing a PSBTv2 file, Krux silently downgraded it to v0. The `trim()` method in `psbt.py` created `PSBT(self.psbt.tx)` without passing the `version` parameter — embit defaults to v0, dropping all v2-specific fields:

- Global: `tx_version`, `locktime`, `input_count`, `output_count`
- Per-input: `txid`, `vout`, `sequence`
- Per-output: `value`, `script_pubkey`

This meant a signed PSBTv2 sent back to Sparrow would be parsed as v0, potentially causing issues with transaction assembly.

## Solution

Pass `version=self.psbt.version` to the PSBT constructor in `trim()`, and copy v2-specific fields from the original PSBT when present.

### Changes

| File | Lines | Description |
|------|-------|-------------|
| `src/krux/psbt.py` | +18 / -1 | `trim()` preserves PSBTv2 version and v2 fields |
| `tests/test_psbtv2.py` | +155 | 6 tests covering parse, serialize, trim, backward compatibility |

### How it works

```python
# Before (bug):
psbt = PSBT(self.psbt.tx)  # version defaults to 0

# After (fix):
psbt = PSBT(self.psbt.tx, version=self.psbt.version)  # preserves v2

For v2 PSBTs, trim() now also copies:

tx_version, locktime from global scope
txid, vout, sequence per input
value, script_pubkey per output
For v0 PSBTs — all v2 fields are None, copy is skipped, behavior identical to before.

Testing
✅ 6/6 PSBTv2 unit tests passing
✅ 30/30 existing PSBT tests passing (15 pre-existing fails are unrelated mock issues)
⚠️ Tested only on simulator — no hardware testing with real PSBTv2 files from Sparrow/SeedSigner yet

### DIFF
--- a/src/krux/psbt.py
+++ b/src/krux/psbt.py
@@ -734,6 +734,8 @@ class PSBTSigner:
             return
 
         trimmed_psbt = PSBT(self.psbt.tx, version=self.psbt.version)
+        trimmed_psbt.tx_version = self.psbt.tx_version
+        trimmed_psbt.locktime = self.psbt.locktime
         for i, inp in enumerate(self.psbt.inputs):
             if inp.final_scriptwitness:
                 trimmed_psbt.inputs[i].final_scriptwitness = inp.final_scriptwitness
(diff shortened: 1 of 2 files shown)


qlrd:
Thxs for finding @Garrysoon.

Some nits: 

- found that you hardcoded some already used constants, so it could be simplified (e.g., "abandon * 11 + about" have a constant for that)
- please read the [contributing guide](https://github.com/selfcustody/krux/blob/develop/CONTRIBUTING.md); we didn't shipped yet to `main` (so maybe wasn't possible to check it), but you need to propose that this PR is being aimed to the correct `branch`
- IMO this is not a `feat` commit, but instead a `fix` one

AFAIK this is (or could -- i don't have opinions on that) already being addressed on [embit#87](https://github.com/diybitcoinhardware/embit/issues/87). Let's wait @odudex and @notTanveer can confirm the trim fix you proposed.

Garrysoon:
> Thxs for finding @Garrysoon.
> 
> Some nits:
> 
> * found that you hardcoded some already used constants, so it could be simplified (e.g., "abandon * 11 + about" have a constant for that)
> * please read the [contributing guide](https://github.com/selfcustody/krux/blob/develop/CONTRIBUTING.md); we didn't shipped yet to `main` (so maybe wasn't possible to check it), but you need to propose that this PR is being aimed to the correct `branch`
> * IMO this is not a `feat` commit, but instead a `fix` one
> 
> AFAIK this is (or could -- i don't have opinions on that) already being addressed on [embit#87](https://github.com/diybitcoinhardware/embit/issues/87). Let's wait @odudex and @notTanveer can confirm the trim fix you proposed.

done

odudex:
Please adapt the PR to `feat/silent-payments` branch so it can be evaluated. In this branch full PSBTv2 support is being introduced together with an Embit update.

PR #889, opened 2026-06-23
fix: Feature/keyboard contour, pre-computed cache

Garrysoon:
### Visual
- **Contour style**: clean outlined keys with 2px margin
- **Selected key**: filled rounded rectangle (touch) / double outline (buttons)
- **Function keys**: colored outlines (Go=green, Esc=red, Del=yellow)
- **Keyset indicator**: 3 grouped bars (Letters / Numbers / Symbols)

### Performance
- **Pre-computed key cache**: labels, x-offsets, colors calculated once per keyset change
- **0 `string_width_px` calls per frame** (was 30)
- **Single loop** draw instead of nested for loops

### UX
- **ABC button**: toggles case only (abc ↔ ABC)
- **Swipe left/right**: skips letter keysets, cycles numbers/symbols only
- **More button**: shows "abc" on number/symbol keysets, "ABC"/"abc" toggle on letter keysets
- **3-bar indicator**: groups abc/ABC as one unit

### Files
- `src/krux/pages/keypads.py` — contour draw_keys, key cache, toggle_case, 3-bar indicator
- `src/krux/pages/__init__.py` — ABC → toggle_case, remove swipe hint


Garrysoon:
64 commits mistake

PR #386, opened 2024-05-09
Fix dockerfile build process

tadeubas:
I had updated my Docker to the latest version (`Docker version 26.1.2, build 211e74b`) and my cache gone. I was having lots of issues with github on the clone step of the build process of the Dockerfile, now is permanently fixed!

This latest version of Docker resolved some issues with cache too, now the build process is lightning fast if you do a release for multiple devices for example.

To update your docker on Linux or WSL type `sudo apt-get update && sudo apt-get upgrade`, then to be sure, close WSL wait 2 to 3 minutes and open it again (this will force a complete restart of the WSL)

### What is the purpose of this pull request? 

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Other

PR #893, opened 2026-06-26
fix(uv): add src to poe task PYTHONPATH

odudex:
### What is this PR for?
Ensure poe-launched commands can import the local krux package under uv, where the project itself is not installed as a package.

### Changes made to:
- [x] Infrastructure
- [ ] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

- No, simulator only

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [X] Other

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,6 +82,9 @@ urtypes = { path = "vendor/urtypes", editable = true }
 [tool.pytest.ini_options]
 pythonpath = ["src"]
 
+[tool.poe.env]
+PYTHONPATH = "${POE_ROOT}/src${PYTHONPATH:+:${PYTHONPATH}}"
+
 [tool.poe.tasks]
 # format tasks
 format-src = "black src"


qlrd (commented):
cACK — works correctly on the locked version. Just one nit. I had trouble following exactly what this line does, so I used some AI to do a deep dig through on poe (EDIT: docs). Concerns below are basically mine, but rewritten on aspect `"why this is the correct fix for the locked version of poe?"`:

--

`pyproject.toml` declares `poethepoet>=0.42.1,<1`, but `:+/:- `parameter expansion inside [tool.poe.env] values on versions:

- v0.43.0 — :-/:+ introduced, but scoped to cmd tasks (changelog links to cmd.html#parameter-expansion-operators).
- v0.45.0 — "Preserve quotes in :+/:- operator arguments" — still cmd-task context.
- v0.46.0 — "Use AST parser for all templating to support :- and :+ param expansion operators everywhere" (PR #386).

`${PYTHONPATH:+:${PYTHONPATH}}` lives in an env value, so before 0.46.0 it goes through the old non-AST interpolator.

Risk is low but real: `uv.lock` pins `0.46.0` and on CI `uv sync --frozen` always installs that, so no live breakage. 

The exposure is anyone resolving to the declared floor (downstream packager, manual pin, stale lock): on `0.42.1–0.45.x` the `:+` maybe wouldn't expand, producing a malformed ` PYTHONPATH` and silently reintroducing the import failure this PR fixes.

Suggestion: bump the "poethepoet>=0.46.0,<1" too.

-- 

The tests that i made to confirm your fix are some verbose, so not shown here, but i can share in some gist.

EDIT: poe just bumped to [0.47.0](https://github.com/selfcustody/krux-installer/pull/273)

PR #904, opened 2026-07-16
Krux Apps, small utility programs to enhance Krux (follow up)

PMK:
### What is this PR for?

Follow-up of https://github.com/selfcustody/krux/pull/485

Harden security and move example kapps outside the repo.

- Re-verify the kapp signature inside execute_flash_kapp on every execution,
  menu path and startup path alike. Flash content is never trusted from an
  earlier check.
- Run the startup kapp only after TC code / flash-hash verification, and
  never skip the SD firmware-update check (boot.py reordered; startup kapp
  no longer gates check_for_updates).
- Always restart the device after a kapp executes, startup kapps included,
  so a misbehaving kapp cannot leave a poisoned module state driving login,
  TC checks or wallet handling.
- Remove the kapp from sys.modules after execution on the flash path.
- Use a separate KAPP_SIGNER_PUBKEYS tupple to verify kapps instead of the
  firmware signer key. This keeps firmware and kapp signing separate, lets
  different kapps be signed by different maintainers, and still fails closed
  if no trusted key is set.
- Treat only a missing/unreadable .sig as "unsigned"; crypto failures now
  surface as errors instead of silently prompting the user to delete the app.
- Verify the bytes actually written to flash against the user-approved hash
  after the SD->flash copy (TOCTOU/corruption guard).
- Cap .mpy size (256 KiB) before copying to flash.
- Drop deleted unsigned apps from startup.json.
- Validate the kapp contract: a .mpy without a callable `run` is refused with
  a clear "Not a Krux app" error instead of an opaque failure.
- Log execution failures via sys.print_exception for debuggability.
- Replace remaining bare excepts with specific ones (OSError on file reads).
- Document the Kapp contract and security model in the module docstring and
  the Tools feature docs.

One sandboxed loader (_with_flash_kapp) replaces the two divergent ad-hoc
import blocks in execute_flash_kapp and the SD-install startup-flag check. Its
teardown - close the flash import sandbox, chdir back out, drop the module
from sys.modules, gc - runs in a finally block, so it happens even when import
or the kapp itself raises.

Also updates the mpy-all poe task from poetry to uv (develop moved to uv).

### Changes made to:
- [x] Code
- [x] Tests
- [x] Docs
- [x] CHANGELOG

- [ ] Bug fix
- [x] New feature
- [x] Docs update
- [ ] Other

cc @tadeubas and @kravens


qlrd:
it duplicate #485 (basically full `src/krux/pages/kapps.py`)

PMK:
> it duplicate #485 (basically full `src/krux/pages/kapps.py`)

Correct. It's a follow up with the review comments and the mentioned improvements/fixes in the first post.

@tadeubas: let me know if you have feedback, comments or anything else.

tadeubas:
Hey @PMK, sorry for the late reply. I can't dedicate much time to OSS at the moment, but I did have a chance to look through some of the changes you've made.

One concern I have is that some users have proposed small OPSEC-focused Kapps that should run at startup **before the Krux logo is displayed**, so the device can't be identified as a Krux device. Because of that, I don't think the logo should appear before the startup Kapp runs.

That said, I think it's perfectly fine for the firmware update process to happen beforehand.

tadeubas:
> One concern I have is that some users have proposed small OPSEC-focused Kapps that should run at startup **before the Krux logo is displayed**, so the device can't be identified as a Krux device. Because of that, I don't think the logo should appear before the startup Kapp runs.

The same concern applies to the TC code and flash hash verification happening beforehand. Those checks identify the device as a Krux device, which defeats the OPSEC goal. The Kapp must run before any of those checks.

tadeubas:
If we "always restart the device after a kapp executes, including startup kapps," how would users be able to use the device if a startup kapp is configured? I think the goal is to let users keep a startup kapp enabled while still being able to use the original firmware. Otherwise, this defeats the purpose.

My idea is that only small, signed (trusted) kapps should be allowed to run at startup. Since they don't modify the firmware and have been audited by the developers, they're suitable for OPSEC-related tasks that need to execute before the firmware starts. In that case, they wouldn't need to trigger a restart because they don't pose the same risks as arbitrary kapps.

tadeubas:
"...lets different kapps be signed by different maintainers, and still fails closed if no trusted key is set."

This is an interesting idea and definitely worth discussing. I like the approach, but I think we'd need to talk through the implications before deciding on it.

tadeubas:
"Log execution failures via sys.print_exception for debuggability."

The final version should not include debugging code such as `sys.print_exception`. It is useful for local testing and development, but it should be removed before the code is committed.

tadeubas:
I haven't tested the code yet, but I noticed that all the tests are passing and that new ones were added, which is definitely a good sign 👏

PR #910, opened 2026-07-27
Speed up the test suite and exercise the real secp256k1

odudex:
### What is this PR for?
Test infrastructure only, no src/krux changes. Full suite: 151s to 85s (1121 tests).

Stub gc.collect() in tests. Krux calls it to manage the device's small heap; under CPython it just walks the big mock-object test heap. Profiling put 35.3s across 2821 calls there. Autouse fixture in conftest.py; test_camera.py patches it itself and still passes. 151s to 118s, coverage byte-identical.

Build the C libsecp256k1 that embit pins. vendor/embit already ships secp256k1-zkp plus a Makefile, but nothing built it, so tests ran embit's pure Python EC. That backend also behaves differently: firmware.bin.bad.sig raises while parsing, where libsecp256k1 parses it and fails verification like the device does. So test_upgrade_fails_when_sig_is_bad was passing via the blanket except: and the real "Bad signature" branch never ran in CI. Now it does: firmware.py 99% to 100%. Adds poe secp256k1-build / poe secp256k1-check, both wired into CI so a failed build fails loudly instead of falling back silently. 118.2s to 84.6s.

Also fixes pre-existing uv.lock drift (embit 0.8.0 vs the vendored 0.8.1), folded into the first commit by a rebase.

Verify

uv run poe secp256k1-build
uv run poe secp256k1-check   # prints "C library"
uv run poe tests             # 1121 passed

Contributors need gcc and make for the one-time build; without it the suite still passes, just slower. 

### Changes made to:
- [ ] Code
- [x] Tests
- [x] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,25 @@ vulture.ref = "vulture-src vulture_whitelist.py"
 vulture-make = { shell = "vulture src --make-whitelist > vulture_whitelist.py" }
 vulture-whitelist.ref = "vulture-make"
 
+# secp256k1 tasks
+# Builds the libsecp256k1-zkp pinned by the embit submodule and drops it where
+# embit's ctypes backend looks for it. Without it embit falls back to its pure
+# Python EC implementation, which is slower and does not always behave like the
+# C library the firmware runs.
+secp256k1-build = { shell = """
+make -C vendor/embit/secp256k1
+mkdir -p vendor/embit/src/embit/util/prebuilt
+cp vendor/embit/secp256k1/build/libsecp256k1_* vendor/embit/src/embit/util/prebuilt/
+""" }
+secp256k1-check = { shell = '''python -c "
+import sys
+from embit.util import secp256k1
+
+using_c = hasattr(secp256k1, '_secp')
+print('embit secp256k1 backend:', 'C library' if using_c else 'pure Python')
+sys.exit(0 if using_c else 1)
+"''' }
+
 # test tasks
 test-clean = """python -c 'import shutil, os; os.path.exists("htmlcov") and shutil.rmtree("htmlcov")'"""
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -108,6 +108,25 @@ vulture.ref = "vulture-src vulture_whitelist.py"
...
+cp vendor/embit/secp256k1/build/libsecp256k1_* vendor/embit/src/embit/util/prebuilt/
+""" }
+secp256k1-check = { shell = '''python -c "
+import sys
+from embit.util import secp256k1
+
+using_c = hasattr(secp256k1, '_secp')
+print('embit secp256k1 backend:', 'C library' if using_c else 'pure Python')
+sys.exit(0 if using_c else 1)
+"''' }
```
qlrd, line 128: Just an additional comment from a `mac tester`.

I tried to verify if I was really using the pinned lib (had issues on with `secp256k1` in the past and did a double check). Just changed a little to confirm (not a suggestion, just a verbose review):

```toml
secp256k1-check = { shell = '''python -c "
import sys
from embit.util import secp256k1

using_c = hasattr(secp256k1, '_secp')
print('embit secp256k1 backend:', 'C library' if using_c else 'pure Python')
if using_c:
    print('loaded from:', secp256k1._secp._name)
sys.exit(0 if using_c else 1)
"''' }
```

and correctly loaded:

```
loaded from: /my/path/vendor/embit/src/embit/util/prebuilt/libsecp256k1_darwin_arm64.dylib
```

qlrd (approved):
tACK beb74888da57fe974427d4d6085b0844aadc59ee

Just one nit about a veery-specific macos behaviour (maybe @joaozinhom could confirm). IDK if it's worth to check more on that, but lmk if you wanna:

Issue #915, opened 2026-08-02
[Enhancement] Double entropy method / input for mnemonic creation


comassky:
Hello,

With everything going on with Coldcard, the issue of entropy is becoming increasingly important.

Even though Krux seems relatively unconcerned, this remains a major issue

Currently, seed generation requires one entropy method (photo, 6/20 dice).

I think it would be a good idea, during the seed generation process, to add a second entropy method (even if it's the same method, 2 photos entropy ?)  in order to:

- Increase the strength of the entropy
- Avoid exposure to issues related to a specific entropy source (if the entropy camera has a future vulnerability, it will be mitigated by the other method)

In principle, this dual process would allow us to have confidence in the generated entropy.

It’s possible to do this “manually” using an XOR, but that’s not very intuitive for the “average person.”

qlrd:
Thx for feedback @comassky, i will try to give my point (is more a doubt about your idea than answering something).

> With everything going on with Coldcard, the issue of entropy is becoming increasingly important.
> ...
> It’s possible to do this “manually” using an XOR, but that’s not very intuitive for the “average person.”

I think is important to specify what is "everything". A overview given by [this report](https://wizardsardine.com/blog/coldcard-rng-vulnerability/)<sup>1</sup> would lead me to interpret what you meant as "everything (...) the issue of entropy" sumarized as `TRNG xor PRNG -> result entropy`, in something analog to the idea of `camera entropy XOR dice entropy -> result entropy`. This is what you meant?

-- 

**1**: Is good to read carefully all. But the part that i meant related to your idea starts on **The advertised design, and the real code**.

comassky:
A slight slip of the tongue: by "everything," I meant "the whole ColdCard affair."

I know the issue is software-related, involving ridiculously low software entropy (even though they had a TRNG chip in the hardware...).

The idea behind my proposal is simply the option (or requirement?) to provide two entropy inputs.

This could be done using two different types (e.g., a camera and dice).

Or, at the very least, using two inputs (e.g., two photos) from the same source.

Combining two entropy inputs—especially from different sources—provides even greater peace of mind.

It Can be (Non-exhaustive list) : 
```
camera entropy XOR camera entropy
camera entropy XOR dice entropy
dice entropy XOR dice entropy
```

exorcist-transfix-welsh:
[...]

PR #917, opened 2026-08-03
Add double entropy workflow

comassky:
This pull request introduces an optional feature for mixing entropy from two sources when generating a new mnemonic, enhancing security by allowing users to combine entropy from camera snapshots and/or dice rolls. The implementation adds user prompts for adding, confirming, or aborting the use of a second entropy source, and updates both the core logic and test coverage to support these flows.

**Enhanced entropy generation and mixing:**

* Added support for optionally mixing two entropy sources (camera, D6, or D20 dice) when generating a new mnemonic, with user prompts to guide the process and handle edge cases such as selecting the same source twice or aborting the second entropy capture (`src/krux/pages/login.py`).
* Introduced utility methods for entropy mixing, normalization, and user confirmation, and refactored the entropy collection logic to support the new flows (`src/krux/pages/login.py`).
* Updated the dice rolls entropy capture to accept a pre-selected mnemonic length, supporting the new two-source flows (`src/krux/pages/new_mnemonic/dice_rolls.py`).

**Testing and validation:**

* Added comprehensive tests for all new entropy mixing flows, including: mixing two sources, cancelling or aborting the second source, handling repeated source selection, and double mnemonic generation with two sources (`tests/pages/test_login.py`).

**Constants and code structure:**

* Defined new constants for entropy source selection to simplify menu handling and internal logic (`src/krux/pages/login.py`).

### DIFF
--- a/src/krux/pages/login.py
+++ b/src/krux/pages/login.py
@@ -113,53 +113,65 @@ class Login(MnemonicLoader):
             hashlib.sha256(entropy_1).digest() + hashlib.sha256(entropy_2).digest()
         ).digest()
 
-    def _maybe_add_second_entropy(self, entropy_bytes, len_mnemonic, first_source):
-        """Optionally collect and mix a second entropy source"""
+    def _show_entropy_hash(self, title, entropy_bytes):
+        """Display a hex-encoded hash and wait for user acknowledgement"""
         import binascii
 
+        entropy_hash = binascii.hexlify(entropy_bytes).decode()
+        self.ctx.display.clear()
+        self.ctx.display.draw_centered_text(
+            title + "\n\n%s" % entropy_hash,
+            highlight_prefix=":",
+        )
+        self.ctx.input.wait_for_button()
+
+    def _confirm_single_source_fallback(self, reason_text):
+        """Ask user if flow should continue with only the first entropy source"""
+        warning_msg = reason_text
+        warning_msg += "\n\n"
+        warning_msg += t("Proceed with a single source?")
+        return self.prompt(warning_msg, BOTTOM_PROMPT_LINE)
+
+    def _confirm_same_source_proceed(self):
+        """Warn user when selecting the same entropy source twice"""
+        warning_msg = t("Same source selected.")
+        warning_msg += "\n"
+        warning_msg += t("Entropy gain may be limited.")
+        warning_msg += "\n\n"
+        warning_msg += t("Proceed anyway?")
+        proceed_same_source = self.prompt(warning_msg, BOTTOM_PROMPT_LINE)
+        self.ctx.display.clear()
+        return proceed_same_source
+
+    def _maybe_add_second_entropy(self, entropy_bytes, len_mnemonic, first_source):
+        """Optionally collect and mix a second entropy source"""
         if not self.prompt(t("Add a 2nd entropy source?"), BOTTOM_PROMPT_LINE):
             return entropy_bytes
 
         while True:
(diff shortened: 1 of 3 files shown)


comassky:
Work in progress

jdlcdl:
In light of recent low-entropy prng news, it's only responsible to review and try to improve how captured natural entropy makes its way from the user to krux.  For that, I love the focus, thoughts and action to have submitted this pr.

I'm undecided on (but leaning away-from) whether krux should express to the user any paranoia about its own entropy sources -- and extending that to force users to merge more than one source. Visible failure when entropy can be measured as insufficient, otherwise to use entropy that passes internal pattern and shannon-entropy checks w/o warnings is the current krux behavior.  This is simplest for users to understand and verify, and krux has tools to allow adding more entropy (mnemonic XOR, editing the mnemonic before loading, editing grid layouts like tinyseed).

As mentioned previously, mnemonic XOR is fully verifiable by hand and in-device, which I prefer to the idea of combining hashes (sure, these are reported on screen and could be verified externally, but user cannot verify that the image captured actually hashes to the hex digest reported on screen).

I'll enjoy watching more progress on this PR (if still in progress) and further comments from others.

Thank you for submitting!

comassky:
> In light of recent low-entropy prng news, it's only responsible to review and try to improve how captured natural entropy makes its way from the user to krux. For that, I love the focus, thoughts and action to have submitted this pr.
> 
> I'm undecided on (but leaning away-from) whether krux should express to the user any paranoia about its own entropy sources -- and extending that to force users to merge more than one source. Visible failure when entropy can be measured as insufficient, otherwise to use entropy that passes internal pattern and shannon-entropy checks w/o warnings is the current krux behavior. This is simplest for users to understand and verify, and krux has tools to allow adding more entropy (mnemonic XOR, editing the mnemonic before loading, editing grid layouts like tinyseed).
> 
> As mentioned previously, mnemonic XOR is fully verifiable by hand and in-device, which I prefer to the idea of combining hashes (sure, these are reported on screen and could be verified externally, but user cannot verify that the image captured actually hashes to the hex digest reported on screen).
> 
> I'll enjoy watching more progress on this PR (if still in progress) and further comments from others.
> 
> Thank you for submitting!

Thanks for the feedback!

I actually wanted to stick somewhat to the approach Trezor takes with its triple source of entropy:

> - The hardware random number generator (TRNG) built into the device's microcontroller. 
> - The Secure Element found in recent models (like the Safe series), which provides an additional layer of security and randomness.
> - External entropy provided by the host (computer or smartphone via the Trezor Suite app), generated by the operating system.


I knew this would get people talking :D

I like the idea of ​​XORing and so on, but I find the process too complex for the average user—and a potential source of errors with disastrous consequences.

My mindset is more about providing strong entropy (sourced from multiple places to protect against a single point of failure) to users without them even realizing it.

qlrd:
This is related to #915 right? If yes, i would like to ask to put some `fix/addresses #915`-like message on commit message just to control flow of opened/closed issues. Also suggest to put it in `draft` while working and reopening when done (so the notification goes when opened again).

> As mentioned previously, mnemonic XOR is fully verifiable by hand and in-device, which I prefer to the idea of combining hashes (sure, these are reported on screen and could be verified externally, but user cannot verify that the image captured actually hashes to the hex digest reported on screen).

(EDIT: ACKed on that) @jdlcdl, this was a common behaviour 1 or 2 days IIRC after incident on KruxBR groups. Like an social experiment inverted concept of "Mnemonic XOR backup" to "Mnemonic XOR DIY entropy". Was very creative way to do lemonade.

PR #921, opened 2026-08-06
Bump GitHub Actions to Node 24 releases

odudex:
### What is this PR for?
Bump GitHub Actions to Node 24 releases

To be merged on `develop` too

### Changes made to:
- [ ] Code
- [x] Tests
- [ ] Docs
- [ ] CHANGELOG

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,9 +17,9 @@ jobs:
 
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
       - name: Setup Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v7
         with:
           python-version: ${{ matrix.py-version }}
       - uses: psf/black@stable
@@ -51,11 +51,11 @@ jobs:
         py-version: [ "3.11.5", "3.12.13"]
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
         with:
           submodules: recursive
       - name: Install uv
-        uses: astral-sh/setup-uv@v4
+        uses: astral-sh/setup-uv@v9.0.0
         with:
           enable-cache: true
       - name: Set up Python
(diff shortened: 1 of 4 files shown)


qlrd:
ACK c0a47ea5f2a57f399906456c1aa3b59ede236f3f

jdlcdl:
ACK at c0a47ea (untested),

Did verify that in all changes, previous versions would not have been supported after Sep 16, 2026 -- and new versions meet/exceed Node24 support  (but I don't know if elsewhere needs change... nor how to test).

* tests.yml:      `- uses: psf/black@stable` might this become `psf/black@25.1.0`?

note: our uv.lock has us using black-26.5.1 and I've not had probs.

.github/workflows/tests.yml
```diff
@@ -17,9 +17,9 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v7
       - name: Setup Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v7
         with:
           python-version: ${{ matrix.py-version }}
       - uses: psf/black@stable
```
jdlcdl, line 25: @stable or @25.1.0 for psf/black?

qlrd:
> Did verify that in all changes, previous versions would not have been supported after Sep 16, 2026 (but I don't know if elsewhere needs change... nor how to test).

there's a way to test locally, but will require a lot of docker space: https://github.com/nektos/act

> tests.yml: - uses: psf/black@stable might this become psf/black@25.1.0?

Good catch, AFAIK is better to pin on tagged versions

.github/workflows/tests.yml
qlrd: IMO `black@v<some desired version>` and have my re-(ut)ACK :)

jdlcdl (approved):

.github/workflows/tests.yml
qlrd: https://github.com/psf/black/blob/main/docs/integrations/github_actions.md From what i understood and interpret here, `@stable` would always use the most updated while `@<version>` pins. Since it's `black`, maybe no problem, but if some bump break thinks, better keep on `@<version>`.

jdlcdl:
Thank you

PR #925, opened 2026-08-10
feat: Silent Payments (BIP-352) — address, descriptor and signing

notTanveer:
Adds Silent Payments (BIP-352) to Krux as a fourth wallet policy: receiving to
an `sp1`/`tsp1` address, exporting the scan key and `sp()` descriptor, signing
PSBTs that pay Silent Payment outputs (BIP-375/376), and recognising our own SP
change so it is not shown as a spend to a stranger.

This is a re-worked history of the Silent Payments work; it is based on
`develop` and split into 8 reviewable commits.

## Commits

| Commit | What it does |
| --- | --- |
| `chore(deps): bump embit` | Pulls in BIP-352/370/374/375/376/392 support, a bech32 fix for larger payloads, and secp256k1 fixes. Adapts the Datum Tool to `bech32_decode` now raising `Bech32DecodeError` instead of returning a sentinel. |
| `feat(sp): add Single-sig SP as a fourth policy type` | `TYPE_SILENT_PAYMENT` wired through key derivation (`m/352h/<coin>h/<n>h`), the wallet model, settings and wallet creation. Policy always uses P2TR. |
| `feat(sp): show and export the SP address, scan key and descriptor` | Address screen renders the SP address as Text/QR with block highlighting; pub-key screen exports the scan key with its `[fingerprint/derivation]` origin and the wallet as an `sp(...)` descriptor; descriptor screen accepts `sp()`. |
| `feat(sp): sign PSBTs paying Silent Payment outputs` | Krux acts as the BIP-375 Signer: derives the Taproot output scripts, attaches global ECDH shares and DLEQ proofs, returns an **unfinalized** PSBT so the coordinator can verify outputs before finalizing. |
| `feat(sp): detect own SP change and self-transfer outputs` | Derives BIP-352 scan/spend keys from the seed for any policy and matches them against SP outputs; label 0 is change, any other label is a self-transfer. |
| `feat(sp): add Silent Payments translation entries` | Registers the new strings across the locale files and baked modules. |

## Notes for reviewers

- **Unfinalized PSBT is deliberate.** embit finalizes taproot key-path inputs on
  signing, so the inputs are explicitly cleared afterwards. Leaving them
  finalized makes coordinators skip the merge that picks up the derived output
  scripts.
- **Eligibility and ownership are checked while loading**, not after review, so a
  PSBT this seed cannot sign is rejected up front. Only the BIP32 half runs
  there; the ECDH/DLEQ work happens once inside `sign_with()`.
- **Validation is a single point for every load path.** A validation failure must
  never fall back to another encoding, which would re-read the PSBT with
  `CompressMode.CLEAR_ALL` and strip the SP fields.
- **DLEQ `aux_rand` is hedged from the seed and the PSBT** rather than the K210
  RNG. BIP-374 binds the proof nonce to the statement, so the near-deterministic
  input is safe.
- **Trimmed QR artifact** keeps the SP outputs, ECDH shares and DLEQ proofs, but
  drops the coordinator-authored BIP-376 `sp_tweak` fields (~70 bytes per input
  on the animated QR).
- **Signing warns** that the `sp1`/`tsp1` address must be verified with the
  recipient, never the derived P2TR shown on-chain.
- **SP change detection works across policies**: the same seed loaded as plain
  single-sig still recognises its SP change. A wallet on a non-default SP
  account fails the ownership check and the output stays a spend, which is the
  safe direction.


qlrd:
Thxs @notTanveer/@odudex i will do a proper check on my side and LYN.

qlrd:
all LGTM with sparrow against `2.5.3` (compiled). Tested a first time with:

```bash
uv sync --frozen --reinstall-package=uUR
uv run poe secp256k1-build
uv run poe secp256k1-check
uv run poe pre-commit
./krux build-release
```

And tested with Amigo only for now.

jdlcdl:
Wow, this is so neat!

The commonly used krux-devs' "gramps" wallet on signet now has a silent-payment tx:
6bf8cdddc2ca513ec03dd17ae520804fe520fae1c6aea140d4afdde036ebc540
...of 1m sats from gramps' native-segwit to his SP (acct#0) wallet.

Continuing with testing, another tx exists in the same wallet sending 50k sats back to gramps native-segwit and collecting the rest into a different SP address as change:
ea0cf1dbbf45528150875c9c6ad33623e3591b80c27a5b2a9a2138c8f93423f7

All working for me up to this point as-of commit 2ddb821, even though this PR is still in draft.

---

Problem when paying to self (is that even something one would try?)

I did provoke an error along the way, which is an odd-case (not a real use-case, or would it be de-consolidating?),
where I setup to send the original tx back to gramps w/ change and ALSO an output to gramps' very own SP address. Sparrow set it up just fine, krux pr_925 branch signed w/o errors (showing same SP addresses that Sparrow shows), but upon loading the signed PSBT back into Sparrow and expecting to be able to broadcast/update-tx, an error about invalid SP address appears in a dialog window.
pr925-SP-invalid-sp-address
I cannot share the PSBT here, but I have a copy of it saved if needed for inspection.

qlrd:
> Problem when paying to self (is that even something one would try?)

had exact same error in a first try with other @notTanveer SP PR, with same context reported by you.

notTanveer:
> an error about invalid SP address appears in a dialog window.

this goes away once [drongo#50](https://github.com/sparrowwallet/drongo/pull/50) lands.

it's a spec issue more than a sparrow one. bip-375's text says to sort the silent
payment codes lexicographically and then by output index in ascending order if there are multiple silent payment codes with the same scan/spend keys before assigning `k`, but the test vectors and the
reference validator shipped in bitcoin/bips still assign `k` by output index, so
implementations ended up split depending on which of the two they built against.
[bitcoin/bips#2207](https://github.com/bitcoin/bips/pull/2207) brings the vectors
and the reference back in line with the text, and it's still open.

krux follows the text, so nothing changes on our side. drongo#50 moves sparrow to
the same order and then the two agree.

it only shows up when a tx has two or more sp outputs sharing a scan key, so paying
the same address twice, or paying yourself with sp change like in this case. with a
single sp output `k` is always 0 and everyone agrees.

thanks for reporting, @jdlcdl & @qlrd

qlrd:
> > an error about invalid SP address appears in a dialog window.
> 
> this goes away once [drongo#50](https://github.com/sparrowwallet/drongo/pull/50) lands.
> 
> it's a spec issue more than a sparrow one. bip-375's text says to sort the silent payment codes lexicographically before assigning `k`, but the test vectors and the reference validator shipped in bitcoin/bips still assign `k` by output index, so implementations ended up split depending on which of the two they built against. [bitcoin/bips#2207](https://github.com/bitcoin/bips/pull/2207) brings the vectors and the reference back in line with the text, and it's still open.
> 
> krux follows the text, so nothing changes on our side. drongo#50 moves sparrow to the same order and then the two agree.
> 
> it only shows up when a tx has two or more sp outputs sharing a scan key, so paying the same address twice, or paying yourself with sp change like in this case. with a single sp output `k` is always 0 and everyone agrees.
> 
> thanks for reporting, @jdlcdl & @qlrd

thxs for feedback @notTanveer. I will give more check on those PRs.

notTanveer:
rebased

jdlcdl:
> drongo#50 moves sparrow to the same order and then the two agree.

I confirm that with Sparrow 2.5.4 and no changes (except for your rebase) to krux solves the issue where Sparrow errored upon scanning the signed SP input paying to an external address, its own SP address and one SP change address, is resolved.

I had errors broadcasting but I suspect it is some new problem between Sparrow 2.5.4 and frigate 2.5.3.  I was able to view the updated (extracted) tx and broadcast it via bitcoin-cli.  Sparrow 2.5.4 had no problem identifying the new utxos while in mempool and once confirmed.

Thank you!

src/krux/pages/home_pages/addresses.py
```diff
@@ -171,28 +171,27 @@ def _qr_highlight_addr(self, formatted_text, y_offset):
...
-                        else lambda: SaveFile(self.ctx).save_file(
-                            addr,
-                            "address",
-                            "address",
-                            t("Address") + ": ",
-                            ADDRESSES_FILE_EXTENSION,
-                            save_as_binary=False,
-                        )
-                    ),
+                    None if not self.has_sd_card() else _save_addr_to_sd,
```
qlrd, line 194: just thoughts, not blockable, why not:

```suggestion
                    (
                        None
                        if not self.has_sd_card()
                        else lambda: SaveFile(self.ctx).save_file(
                            addr,
                            "sp-address",
                            "sp-address",
                            file_extension=".txt",
                            save_as_binary=False,
                            prompt=False,
                        )
                    )
```

src/krux/pages/home_pages/addresses.py
```diff
@@ -171,28 +171,27 @@ def _qr_highlight_addr(self, formatted_text, y_offset):
...
+                # save_file()'s bool return would be read as a menu status.
+                SaveFile(self.ctx).save_file(
+                    addr,
+                    "sp-address",
+                    "sp-address",
+                    file_extension=".txt",
+                    save_as_binary=False,
+                    prompt=False,
+                )
+
```
qlrd, line 190: just thouhgts, not blockable, why not:

```suggestion
```

qlrd (commented):
just nit/doubt. IMP htis could be less (not too much) diff

src/krux/pages/home_pages/addresses.py
notTanveer: better readability

src/krux/pages/wallet_settings.py
```diff
@@ -386,6 +395,13 @@ def _derivation_path_str(self, policy_type, script_type, network, account):
         if policy_type == TYPE_MULTISIG and script_type == P2SH:
             return P2SH_DEFAULT_DERIVATION
 
+        if policy_type == TYPE_SILENT_PAYMENT:
+            from ..key import Key
```
qlrd: Just to comply with #930:

```suggestion
            from krux.key import Key
```

tests/pages/test_wallet_settings.py
```diff
@@ -1052,3 +1052,62 @@ def test_change_derivation_path_not_hardened_node(amigo, mocker, tdata):
...
+        == "m/352h/0h/0h"
+    )
+    assert (
+        wallet_settings._derivation_path_str(
+            TYPE_SILENT_PAYMENT, P2TR, NETWORKS["test"], 0
+        )
+        == "m/352h/1h/0h"
+    )
+    # A P2SH multisig wallet carries account=None; switching policy to SP must
+    # fall back to account 0 instead of raising
```
qlrd, line 1079: I think it's worth of a follow-up, wdyt @jdlcdl ?

tests/pages/test_login.py
```diff
@@ -1945,3 +1945,31 @@ def test_load_key_from_tiny_seed_image_back_is_silent(m5stickv, mocker):
...
+        login = Login(ctx)
+        login._load_key_from_words(MNEMONIC.split())
+
+        assert ctx.input.wait_for_button.call_count == len(BTN_SEQUENCE)
+        # Silent Payment policy forces the P2TR script type
+        assert ctx.wallet.is_silent_payment()
+        assert ctx.wallet.key.policy_type == TYPE_SILENT_PAYMENT
+        assert ctx.wallet.key.script_type == P2TR
+    finally:
+        Settings().wallet.policy_type = original_policy
```
qlrd, line 1976:
```suggestion
def test_load_wallet_silent_payment_policy(amigo, mocker, tdata):
    from krux.pages.login import Login
    from krux.krux_settings import Settings
    from krux.key import NAME_SILENT_PAYMENT, TYPE_SILENT_PAYMENT, P2TR, P2WPKH:w
    
    from krux.input import BUTTON_ENTER

    original_policy = Settings().wallet.policy_type
    Settings().wallet.policy_type = NAME_SILENT_PAYMENT
    try:
        BTN_SEQUENCE = [
            BUTTON_ENTER,  # Confirm words
            BUTTON_ENTER,  # Load Wallet
        ]
        ctx = create_ctx(mocker, BTN_SEQUENCE)
        login = Login(ctx)
        login._load_key_from_words(tdata.TEST_12_WORD_MNEMONIC)

        assert ctx.input.wait_for_button.call_count == len(BTN_SEQUENCE)
        # Silent Payment policy forces the P2TR script type
        assert ctx.wallet.is_silent_payment()
        assert ctx.wallet.key.policy_type == TYPE_SILENT_PAYMENT
        assert ctx.wallet.key.script_type == P2TR
    finally:
        Settings().wallet.policy_type = original_policy
        
    assert Settings().wallet.policy.policy_type == P2WPKH
```

tests/test_wallet.py
```diff
@@ -521,6 +522,68 @@ def test_init_miniscript(mocker, m5stickv, tdata):
     assert wallet.policy is None
 
 
+def test_init_silent_payment(mocker, m5stickv, tdata):
+    from krux.wallet import Wallet
+    from krux.key import Key, TYPE_SILENT_PAYMENT, P2TR
+    from embit.networks import NETWORKS
+
+    key = Key(
+        "olympic term tissue route sense program under choose bean emerge velvet absurd",
```
qlrd: Some pre-loaded key of your choice:

```suggestion
        tdata.TEST_12_WORD_MNEMONIC
```

tests/test_wallet.py
```diff
@@ -521,6 +522,68 @@ def test_init_miniscript(mocker, m5stickv, tdata):
...
+    assert len(addresses_from_1) == 0
+
+
+def test_silent_payment_load_rejected(mocker, m5stickv, tdata):
+    from krux.wallet import Wallet
+    from krux.key import Key, TYPE_SILENT_PAYMENT
+    from embit.networks import NETWORKS
+
+    key = Key(
+        "olympic term tissue route sense program under choose bean emerge velvet absurd",
```
qlrd: same

tests/test_wallet.py
```diff
@@ -521,6 +522,68 @@ def test_init_miniscript(mocker, m5stickv, tdata):
...
+        wallet.load("some wallet data", 0)
+
+
+def test_determine_descriptor_policy_silent_payment(mocker, m5stickv, tdata):
+    from krux.wallet import Wallet
+    from krux.key import Key, TYPE_SILENT_PAYMENT, P2TR
+    from embit.networks import NETWORKS
+
+    key = Key(
+        "olympic term tissue route sense program under choose bean emerge velvet absurd",
```
qlrd: same

tests/test_wallet.py
```diff
@@ -2054,3 +2117,37 @@ def test_wallet_load_rejects_descriptor_m_greater_than_n(mocker, m5stickv):
...
         wallet.load("fake_data", 0)
+
+
+def test_silent_payment_obtain_sp_address(mocker, m5stickv, tdata):
+    from krux.wallet import Wallet
+    from krux.key import Key, TYPE_SILENT_PAYMENT
+    from embit.networks import NETWORKS
+
+    key = Key(
+        "olympic term tissue route sense program under choose bean emerge velvet absurd",
```
qlrd: same

tests/test_wallet.py
```diff
@@ -2054,3 +2117,37 @@ def test_wallet_load_rejects_descriptor_m_greater_than_n(mocker, m5stickv):
...
+    assert list(wallet.obtain_addresses()) == [address]
+
+
+def test_silent_payment_validate_xpub_match_skipped(mocker, m5stickv, tdata):
+    from krux.wallet import Wallet
+    from krux.key import Key, TYPE_SILENT_PAYMENT
+    from embit.networks import NETWORKS
+
+    key = Key(
+        "olympic term tissue route sense program under choose bean emerge velvet absurd",
```
qlrd: same

src/krux/pages/home_pages/addresses.py
qlrd: sure, no problems to me, keep as it.

tests/pages/test_login.py
qlrd: i have doubts about last line, please check it

tests/pages/test_wallet_settings.py
jdlcdl: If I understand correctly, I agree that None in the normal "account" node of a derivation path for bip352 (and many others), should default to "0h".

tests/test_wallet.py
notTanveer: used tdata.TEST_MNEMONIC1 instead

src/krux/pages/home_pages/addresses.py
notTanveer: i could have used the save address file extension function here, but that's save as .csv format. and fo SP wallet, which only has ONE static address it doesn't make sense so used .txt here.

now that i think of it we can possibly create a utils function to save with name and file extension as params

qlrd: this could be amazing!

notTanveer: done in, [cf49505](https://github.com/selfcustody/krux/pull/925/commits/cf495051aa1b0de0b63a3f02bc5a5e518b4ea92d)

PR #926, opened 2026-08-10
Add Disclaimer - Krux Research and Development Focus

odudex:
### What is this PR for?
This PR aims to harden and better inform about what is Krux. I does it by embedding a disclaimer in firmware

### Changes made to:
- [x] Code
- [x] Tests
- [x] Docs
- [ ] CHANGELOG

- [x] Yes, build and tested on TZT

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


qlrd:
cACK, i will give proper review and LYN fast (and calm) as possible.

tadeubas:
> "Innovative features may have undiscovered flaws that endanger funds."

This basically true of any Bitcoin wallet/software, but phrasing it this way makes Krux sound exceptionally unsafe. It also doesn't really give the user actionable information.

> "Research and development project made by nerds building tools for their own interests "

It is implicit that this isn't a commercial, professionally supported product. IMHO this doesn't add much.

The disclaimer also  doesn't meaningfully improve security. Users should already understand that self-custody means they are responsible for evaluating and accepting the risks of the software/hardware they use. Most importantly, if the goal is to communicate risk, I'd rather be specific about the experimental nature of particular features than put a blanket warning on the whole project.

odudex:
> This basically true of any Bitcoin wallet/software, but phrasing it this way makes Krux sound exceptionally unsafe. It also doesn't really give the user actionable information.

> It is implicit that this isn't a commercial, professionally supported product. IMHO this doesn't add much.
 
> The disclaimer also doesn't meaningfully improve security. Users should already understand that self-custody means they are responsible for evaluating and accepting the risks of the software/hardware they use. Most importantly, if the goal is to communicate risk, I'd rather be specific about the experimental nature of particular features than put a blanket warning on the whole project.

I agree with most of your comments. The issue I see is that many users have come, and continue to come, to Krux through people who present it as a product. Although we have expressed our disapproval of this behavior, it hasn't worked, so in my opinion we need to push back against it more emphatically.

qlrd:
> The disclaimer also doesn't meaningfully improve security. Users should already understand that self-custody means they are responsible for evaluating and accepting the risks of the software/hardware they use. Most importantly, if the goal is to communicate risk, I'd rather be specific about the experimental nature of particular features than put a blanket warning on the whole project.


Agree on all and your position and is very rationale (ideally i really desire that ethos); but sometimes is necessary to point specific things clearly and with verbosity.

qlrd (approved):
tACK e971560c54e3cb065960cd746bfc16554bcffe11

in the context of code quality: 

```bash
git checkout krux-RnD-focus
uv sync --frozen --reinstall-package uUR
uv run poe pre-commit
sudo systemctl start docker
./krux build-release
uv run python reproductibility.py
```

All passed on those steps and checked the `bin` too.

Liked very much the idea of "reproductible disclaimer", great work!

Also flashed and tested on Amigo. One thought, but non-blockable is about the used color, but i think would be neat hear @Naman015 about that.

odudex:
replaced red warning color with `highlight_color`

jdlcdl (approved):

PR #930, opened 2026-08-13
General refact, and removing dead calls 

joaozinhom:
### What is this PR for?
This Pr intention is to solve a lot of minor nits that i notice reading the krux codebase.

### Changes made to:
- [x] Code
- [ ] Tests
- [ ] Docs
- [ ] CHANGELOG

- [x] Yes, build and tested on macbook air m4

- [ ] Bug fix
- [ ] New feature
- [ ] Docs update
- [x] Other


tests/pages/home_pages/test_addresses.py
```diff
@@ -1,4 +1,3 @@
-import pytest
```
qlrd, line 1: i thought that was a weird requirement by linter

src/krux/pages/encryption_ui.py
```diff
@@ -116,6 +116,31 @@ def decrypt_kef(ctx, data):
...
 
 
+def try_decrypt_kef(ctx, data, decode=True):
+    """Wraps decrypt_kef, mapping its exceptions to an error tag.
+
+    Returns a (data, error) tuple where error is one of:
+      None        decrypted, or not KEF / user declined (data unchanged)
+      "decrypt"   KEF envelope found but decryption failed
+      "load"      decrypted but the result could not be decoded
+    """
```
qlrd, line 126: I liked it because the function name is a common rust nit and explains better with summarized code what will do :)

For micropython, I would replace this with based on the concept that every byte of docstrings will be added to firmware IIRC (need confirmation). WDYT?

```suggestion
    """Wraps decrypt_kef, mapping its exceptions to an error tag."""
    # Returns a (data, error) tuple where error is one of:
    # None        decrypted, or not KEF / user declined (data unchanged)
    # "decrypt"   KEF envelope found but decryption failed
    # "load"      decrypted but the result could not be decoded
```

src/krux/hashes.py
```diff
@@ -0,0 +1,79 @@
...
+
+def sha256d(data):
+    """
+    Double SHA-256 digest, `SHA256(SHA256(data))`, as used by Bitcoin.
+
+    Returns the digest bytes directly, since there is no meaningful streaming
+    use for it. Never display this as "SHA256" -- it will not match what the
+    user computes by hand.
+    """
+    return _backend.sha256(_backend.sha256(data).digest()).digest()
```
qlrd: Non-blockable, but just being the verbose guy here: wdyt about `doubleSha256` or something a like just to be more standardized (idk if `sha256d` is, please share your thoughts). Also a doc nit:

```suggestion
def sha256d(data):
    """Double SHA-256 digest, `SHA256(SHA256(data))`, as used by Bitcoin."""
    # Returns the digest bytes directly, since there is no meaningful streaming
    # use for it. Never display this as "SHA256" -- it will not match what the
    # user computes by hand.
    return _backend.sha256(_backend.sha256(data).digest()).digest()
```

src/krux/hashes.py
```diff
@@ -0,0 +1,79 @@
...
+
+def sha256(data=b""):
+    """
+    Single SHA-256 hasher, as `hashlib.sha256`.
+
+    Returns a hasher object, so it accepts `.update()` for streaming and
+    `.digest()` for the result. Use this for any hash the user is expected to
+    reproduce or verify externally.
+    """
+    return _backend.sha256(data)
```
qlrd:
```suggestion
def sha256(data=b""):
    """Single SHA-256 hasher, as `hashlib.sha256`."""
    # Returns a hasher object, so it accepts `.update()` for streaming and
    # `.digest()` for the result. Use this for any hash the user is expected to
    # reproduce or verify externally.
    return _backend.sha256(data)

**Review by qlrd — COMMENTED** [review]
LGTM.

I think it's a reasonable PR and i share some thouhgts on that like the imports and naming the `try` function on kef (so @jdlcdl eyes). It need some check on codecov report too since some untouched missing lines -- IDK if it's related. 

Since this touches a important module i will just share some thoughts on docsstrings nits for now an a doubt about the `sha256d` name (althought the code explains by itself IMO).:

**Comment by odudex** [comment]
Something I don't think is worth changing in a refactor: 985f0597 silently moves several hashes from software to K210 hardware SHA

This is labeled chore: refact, but it is a real on-device behavior change, not a no-op. bip39.entropy_checksum, key.py:349, capture_entropy.capture, dice_rolls, sign_message_ui, and encryption.MnemonicStorage._deprecated_decrypt previously used MaixPy's software hashlib (a full implementation, port/src/uhashlib/hashlib.c); they now route through uhashlib_hw, the hardware accelerator.

File: src/krux/pages/home_pages/addresses.py [hunk 5]
```diff
@@ -287,35 +287,27 @@ def _scan_highlight_addr(self, result_message):
...
+        from krux.pages.qr_capture import QRCodeCapture
+        from krux.pages.encryption_ui import try_decrypt_kef
 
         qr_capture = QRCodeCapture(self.ctx)
         data, qr_format = qr_capture.qr_capture_loop()
         if data is None or qr_format != FORMAT_NONE:
             self.flash_error(t("Failed to load"))
             return MENU_CONTINUE
 
-        try:
```
jdlcdl, line 301: note to self: review later.

joaozinhom:
> Something I don't think is worth changing in a refactor: [985f059](https://github.com/selfcustody/krux/commit/985f0597d2a338697d7b4b1dc64084e62137c2d9) silently moves several hashes from software to K210 hardware SHA
> 
> This is labeled chore: refact, but it is a real on-device behavior change, not a no-op. bip39.entropy_checksum, key.py:349, capture_entropy.capture, dice_rolls, sign_message_ui, and encryption.MnemonicStorage._deprecated_decrypt previously used MaixPy's software hashlib (a full implementation, port/src/uhashlib/hashlib.c); they now route through uhashlib_hw, the hardware accelerator.

Good catch! I'll fix `hashes.py` to explicitly import and wrap `hashlib` (software backend), just with no silent backend swap. I'll push the fix squashed into the original commit shortly to keep the review clean.

qlrd (commented):
LGTM on a 2nd round of manual lecture. I will do a strict review with that and LYN

qlrd (changes requested):
3rd round :)

Wen running without rebase, all went well:

```bash
gh pr checkout 930
uv sync --frozen --reinstall-package=uUR
uv run poe pre-commit
```

But due to recent updates, also did:

```bash
git reset --hard cleanup/dependencies_and_dead_code
git rebase origin/develop
```

And got some conflicts locally but easy to resolve (i think).

joaozinhom:
> 3rd round :)
> 
> Wen running without rebase, all went well:
> 
> ```shell
> gh pr checkout 930
> git submodule update --recursive
> uv sync --frozen --reinstall-package=uUR
> uv run poe pre-commit
> ```
> 
> But due to recent updates, also did:
> 
> ```shell
> git reset --hard cleanup/dependencies_and_dead_code
> git rebase origin/develop
> ```
> 
> And got some conflicts locally but easy to resolve (i think).

i will take a look

joaozinhom:
@qlrd you thinki should rebase to develop and change the base here on PR? iv already made it in other branch and everithing looks fine https://github.com/joaozinhom/krux/pull/6

qlrd:
> @qlrd you thinki should rebase to develop and change the base here on PR? iv already made it in other branch and everithing looks fine [joaozinhom#6](https://github.com/joaozinhom/krux/pull/6)

i think would be good. I noted now that is pointed to `main`

src/krux/pages/home_pages/sign_message_ui.py
```diff
@@ -22,28 +22,28 @@
...
 from embit import bip32, compact, script
 from embit.networks import NETWORKS
-import hashlib
 import binascii
-from .. import MENU_CONTINUE, LOAD_FROM_CAMERA, LOAD_FROM_SD, Menu
-from ..utils import Utils
-from ...key import SINGLESIG_SCRIPT_PURPOSE, P2PKH, P2WPKH, P2TR, P2SH_P2WPKH
-from ...themes import theme
-from ...display import (
+from krux.hashes import sha256, sha256d
```
qlrd, line 26: IIRC `embit` have some `hashes` lib that could be useful to use some instead:

```suggestion
from embit.hashes import sha256, double_sha256
```

qlrd (commented):
some nit, i think this will reduce `diff` while keep functionality, wdyt?

src/krux/pages/home_pages/sign_message_ui.py
joaozinhom: embit don't use in any place the uhashlib_hw(hardware accelerated one), so if we apply this suggestion we lost the modularization of the developer can choose from the import file which one he wants to use and in addition to that if in the future we need more specific hash functions we have a file to have this operations...

qlrd: good catch, maybe worth to check on embit side @jdlcdl, @odudex ?

PR #935, opened 2026-08-18
FIX: handle back

notTanveer:
### What is this PR for?
GOING BACK SHOULD NOT BE A FAILURE. this pr fixes that across the sd file browser and the qr scanner.

backing out of the load-method menu already returned silently, but backing out of the deeper sd-card file picker fell through to the "failed to load" error. utils.load_file() returns ("", None) for both user cancellation and a real load failure, so the load flows now treat an empty filename as a user cancellation.

the qr scanner had the same bug. qr_capture_loop() returns (None, None) only when the user leaves the scanner, never on a decode failure, so leaving the camera flashed the same false error. that half is fixed the same way.

this is covered for:

load_psbt()
_load_message()
_load_wallet_data()

### why load_file() now reports its own errors
reading an empty filename as "cancelled" at the call site only works if a real failure never looks the same. it did. load_file() also returned ("", None) when there was no sd card, so pulling the card after the load menu was drawn turned a real failure into a silent no-op instead of an error. has_sd_card() is a live hot-plug check and the load menu only runs it once, when it is built.

so rather than adding a cancel check at each call site, load_file() now reports its own failures. "sd card not detected." for a missing card, "failed to load" for an oserror during the read. an empty filename then means one thing everywhere: nothing was loaded, and the user has already been told why.

### when "failed to load" should actually show up
it should mean we got something and it was unusable. an i/o error reading the card, or bytes that would not decode. it should never mean the user pressed back.

once both halves report themselves, three of the old guards are unreachable and are removed:

sign_psbt() and sign_message() — the camera path returns early, the sd path returns early, and read_binary() never returns None
_load_wallet_data() — same, plus the `except OSError: pass` that load_file() now owns

the guard that stays is the one on wallet_data.decode(), which is the real case: a file that loaded fine but is not text.

datum_tool.read_file() had that same `except OSError: pass` and then read the unbound `contents`, raising NameError instead of showing an sd error. load_file() owns that path now, so the swallow is gone.

### to reproduce
go to sign → message → back, or back out from the sd-card file picker or the qr scanner while loading a psbt, message, or wallet.

for the failure side, pull the sd card after the load menu is drawn but before picking a file.

### Changes made to:
- [X] Code
- [X] Tests

- [X] Yes, build and tested on Maixpy Amigo

- [X] Bug fix

### DIFF
--- a/src/krux/pages/datum_tool.py
+++ b/src/krux/pages/datum_tool.py
@@ -366,15 +366,8 @@ class DatumToolMenu(Page):
         """Handler for the 'Read File' menu item"""
         from .utils import Utils
 
-        if not self.has_sd_card():
-            self.flash_error(t("SD card not detected."))
-            return MENU_CONTINUE
-
         utils = Utils(self.ctx)
-        try:
-            filename, contents = utils.load_file(prompt=False)
-        except OSError:
-            pass
+        filename, contents = utils.load_file(prompt=False)
 
         if not contents:
             return MENU_CONTINUE
--- a/src/krux/pages/encryption_ui.py
+++ b/src/krux/pages/encryption_ui.py
@@ -503,7 +503,7 @@ class EncryptionKey(Page):
 
         while True:
             if key in (None, "", b"", ESC_KEY, MENU_CONTINUE):
-                self.flash_error(t("Failed to load"))
+                # user cancelled the keypad or left the QR scanner
                 return None
 
             self.ctx.display.clear()
--- a/src/krux/pages/home_pages/addresses.py
+++ b/src/krux/pages/home_pages/addresses.py
@@ -292,7 +292,9 @@ class Addresses(Page):
 
         qr_capture = QRCodeCapture(self.ctx)
         data, qr_format = qr_capture.qr_capture_loop()
-        if data is None or qr_format != FORMAT_NONE:
+        if data is None:  # user left the QR scanner
+            return MENU_CONTINUE
+        if qr_format != FORMAT_NONE:
             self.flash_error(t("Failed to load"))
             return MENU_CONTINUE
 
--- a/src/krux/pages/home_pages/home.py
+++ b/src/krux/pages/home_pages/home.py
@@ -222,13 +222,15 @@ class Home(Page):
         load_method = self.load_method()
 
         if load_method > LOAD_FROM_SD:
-            return (None, None, "")
+            return None  # user chose Back
 
         if load_method == LOAD_FROM_CAMERA:
             from ..qr_capture import QRCodeCapture
 
             qr_capture = QRCodeCapture(self.ctx)
             data, qr_format = qr_capture.qr_capture_loop()
+            if data is None:
+                return None  # user left the QR scanner
             return (data, qr_format, "")
 
         # If load_method == LOAD_FROM_SD
@@ -241,6 +243,8 @@ class Home(Page):
             prompt=False,
             only_get_filename=True,
         )
+        if not psbt_filename:
+            return None  # user backed out of the SD file browser
         return (None, FORMAT_NONE, psbt_filename)
 
     def _sign_menu(self, signer, psbt_filename, outputs):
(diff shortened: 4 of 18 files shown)


qlrd:
i needed to rebase to run correctly. Can you check your side?

qlrd:
Strongly cACK. I ran with simulator and liked very much the flow. I will do a proper review on code (also, reproduced on macos the rebase issue).

notTanveer:
> i needed to rebase to run correctly. Can you check your side?

i was able to test it without the rebase. but anyways it's done now

odudex:
I like and agree with the fix.

Three items to act on. No correctness bugs were found in the changed code;
these are cleanup, scope accuracy, and consistency.

## 1. `datum_tool.read_file()` now checks for the SD card twice

`src/krux/pages/datum_tool.py:369-371` still does its own `has_sd_card()` plus
`flash_error(t("SD card not detected."))`, and then `load_file()` does exactly
the same check.

Two consequences:

* Two `has_sd_card()` probes per read. Each one remounts the card and repaints
  "Checking for SD card…".
* In the exact hot-plug race this PR is about, two consecutive error flashes.

The outer guard is now dead code. Behavior is identical without it:
`load_file()` flashes, returns `("", None)`, and `if not contents: return
MENU_CONTINUE` catches it.

```python
    def read_file(self):
        """Handler for the 'Read File' menu item"""
        from .utils import Utils

-       if not self.has_sd_card():
-           self.flash_error(t("SD card not detected."))
-           return MENU_CONTINUE
-
        utils = Utils(self.ctx)
        filename, contents = utils.load_file(prompt=False)
```

That file is already touched by commit `f7dbc7b7`, so the change fits the PR.

## 2. The QR scanner half is fixed in 3 places, not "across the qr scanner"

The PR description opens with "this pr fixes that across the sd file browser and
the qr scanner", but the identical `data is None` then
`flash_error(t("Failed to load"))` pattern survives at five other sites:

| Location | Flow |
| --- | --- |
| `src/krux/pages/mnemonic_loader.py:315` | `load_key_from_qr_code` |
| `src/krux/pages/mnemonic_loader.py:303` | `TinyScanner.scanner()`, which also returns `None` only after the user breaks out (`tiny_seed.py:826`) |
| `src/krux/pages/wallet_settings.py:147` | `_load_qr_passphrase` |
| `src/krux/pages/home_pages/addresses.py:296` | `scan_address`, the `data is None` half only. The `qr_format != FORMAT_NONE` half is a genuine failure and should stay. |
| `src/krux/pages/encryption_ui.py:506` | `encryption_key()` flashes it when the user cancels the keypad (`ESC_KEY`) or leaves the scanner |

The body's "this is covered for: `load_psbt` / `_load_message` /
`_load_wallet_data`" is accurate. The opening sentence is not.

Pick one:

* Extend the fix to the five sites above, or
* Narrow that first line to the three sign/load flows actually covered and note
  the rest as follow-up.

## 3. The three sibling loaders disagree on their cancel contract

`load_psbt()` and `_load_message()` return either `None` or a 3-tuple.
`_load_wallet_data()` kept `(None, None, False)` with a `wallet_data is None`
sentinel check at the call site (`wallet_descriptor.py:323`).

Both are correct. `_load_wallet_data` has three separate `return None, None,
False` paths and they all hold. But three functions solving one problem two
different ways is friction for the next reader. Optional: settle on one shape.

## Trivial

`tests/pages/home_pages/test_wallet_descriptor.py:3` pulls `assert_not_flashed`
through `.test_home` rather than from `..` directly. It matches the file's
existing `from .test_home import tdata, create_ctx` line, so it is consistent,
just indirect.

odudex (approved):
ACK - Tested on TZT

jdlcdl:
tACK at 609e179 on working amigo (and yahboom_devkit w/ failing cam sensor).

Thank you!

jdlcdl (approved):

qlrd (approved):
tACK https://github.com/selfcustody/krux/commit/609e179336d652505b49cdc8f79c654350f522d2 on amigo and wonder_mv

PR #945, opened 2026-09-03
dice: use min-entropy instead of Shannon entropy

### LINKED ISSUE #923: 50/99 dice rolls: Add safety margin because assuming perfectly dice with 2.58496 bits per roll may be a misconception
I don't want to disturb you, but I learned about your project just today. Since it seems there is some heavy development ongoing I wanted to bring my issue to your attention which I created today in the SeedSigner project:

https://github.com/SeedSigner/seedsigner/issues/989

SatsCzar:
## Summary

Closes #923.

Krux currently gates dice-roll entropy (progress bar, "Poor entropy!" warning, "Stats for Nerds") on **Shannon's entropy** of the observed roll distribution. Shannon's entropy is an *average-case* measure — it can stay high even when one particular outcome (e.g. a face on a biased/loaded die) is meaningfully more likely than the others. That is exactly the scenario an attacker exploits, since a brute-force/dictionary attack targets the single most probable sequence, not the average one.

This PR switches the dice-roll entropy calculation, progress bar, and warning gate to **min-entropy**, `H_min(X) = -log2(p_max)`, the conservative, worst-case measure recommended by [NIST SP 800-90B](https://csrc.nist.gov/pubs/sp/800/90/b/final) for entropy source assessment. Min-entropy bounds an attacker's best single-guess success probability directly (`Pr[best guess] ≤ 2^-H_min`), so it does not overstate security the way Shannon's entropy can for a biased ("viciado") source. This is also the conclusion reached by Liu et al. (2022), *Min-entropy estimation for semiconductor superlattice true random number generators*, Scientific Reports 12, 2948:

> "Shannon entropy often leads to overestimating total security when applied to a weak source... it fulfills to know the min-entropy to construct a good random number generator, which is a very conservative method and provides the lower bound extracted from the entropy source."

### Why the minimum roll counts also changed

Because min-entropy is driven only by the single most frequent outcome, it is far less forgiving of ordinary finite-sample noise than Shannon's formula. A Monte Carlo simulation (8,000 trials per scenario, fair die) showed the previous `MIN_ROLLS` values were calibrated against Shannon and would almost never satisfy the min-entropy target even with a perfectly fair die — e.g. for 50 D6 rolls (12-word target), the *best possible* integer distribution ([9,9,8,8,8,8]) only reaches ~124 bits of min-entropy, already short of the 126-bit (128 − 2 tolerance) threshold, and the average real sample was ~25 bits short.

`MIN_ROLLS` were recalibrated from the same simulation so that ~90% of honest/fair-die sessions clear the min-entropy target at the minimum roll count (same 2-bit `ENTROPY_TOLERANCE` as before):

| Scenario | Before | After |
|---|---|---|
| D6, 12 words | 50 | 70 |
| D6, 24 words | 99 | 125 |
| D20, 12 words | 30 | 60 |
| D20, 24 words | 60 | 90 |

Pattern detection is unchanged — it still uses Shannon's entropy over the roll derivatives to flag arithmetic-progression patterns, which is a separate, structural check unrelated to the raw entropy estimate above.

Camera-based entropy capture (`capture_entropy.py`) is out of scope: its Shannon calculation runs in a native MicroPython C module in the `firmware/MaixPy` submodule, so changing it would require a firmware rebuild.

## Changes

- `dice_rolls.py`: min-entropy calculation (`min_entropy_sum`), used by `calculate_entropy()`, the progress bar and the "Poor entropy!" gate; recalibrated `MIN_ROLLS`; "Stats for Nerds" now shows "Min-entropy:" instead of "Shannon's entropy:".
- i18n: added the `"Min-entropy:"` translation string across all locales.
- Tests: updated dice-roll fixtures and expected mnemonics for the new roll counts.
- Docs: `entropy.en.md` now explains min-entropy alongside Shannon's entropy, with a worked example.

## References

- NIST SP 800-90B, *Recommendation for the Entropy Sources Used for Random Bit Generation*: https://csrc.nist.gov/pubs/sp/800/90/b/final
- Liu, J., Xie, J., Chao, L., Wu, H., Ding, P., Chen, X., Feng, H. (2022). *Min-entropy estimation for semiconductor superlattice true random number generators*. Scientific Reports, 12, 2948. https://doi.org/10.1038/s41598-022-06815-2

## Test plan

- [x] `pytest tests/pages/new_mnemonic/test_dice_rolls.py` (13 passed)
- [x] `pytest tests/pages/test_capture_entropy.py` (7 passed, unaffected)
- [x] `python i18n/i18n.py validate`
- [x] `black --check` on all changed Python files

---

## Update 07/09

Following @jdlcdl's review, this PR **no longer raises the minimum roll counts**. Roll sequences recorded alongside a backup have to stay replayable — re-verifying `rolls → mnemonic` against another implementation, or explaining the SHA256 pre-image to demonstrate ownership, are real practices that a higher minimum would break. The original proposal above is kept for history.

Min-entropy is still the measure that is displayed and warned about, but the recalibrated counts became a **recommended** amount instead of a hard requirement:

| Scenario | Minimum (unchanged, enforced) | Recommended (new) |
|---|---|---|
| D6, 12 words | 50 | 70 |
| D6, 24 words | 99 | 125 |
| D20, 12 words | 30 | 60 |
| D20, 24 words | 60 | 90 |

- The **minimum** is still the hard gate: below it, "Not enough rolls!" blocks generation, exactly as before this PR.
- The **recommended** amount is shown on the intro screen ("Recommended: %d rolls") and is what the roll segment of the progress bar and its green outline are measured against.
- Rolling between the minimum and the recommended amount is allowed, and shows the existing "Estimated entropy: Poor!" warning with "Proceed anyway?". That warning does show on every session that stops at the old minimum, which is expected: 128/256 bits of min-entropy are not reachable there — the best possible distribution for 50 D6 rolls is ~123.7 bits and the median fair-die sample is ~102.
- "Stats for Nerds" now shows **both** measures, as suggested: "Min-entropy: X bits" and "Shannon's entropy: Y bits".

Changes since the first version of this PR:

- `dice_rolls.py`: `*_MIN_ROLLS` restored to 50/99/30/60; new `*_RECOMMENDED_ROLLS` constants (70/125/60/90); recommended amount displayed on the intro screen and used by the progress bar; Shannon's entropy displayed next to min-entropy in "Stats for Nerds".
- i18n: added `"Recommended: %d rolls"` across all locales; `"Shannon's entropy:"` is in use here again.
- Tests: expected mnemonics and the poor-rolls fixture are back to their pre-PR values; `test_good_rolls`/`test_stats_for_nerds` roll the recommended amount; new `test_minimum_rolls_warns_but_proceeds` covers generating at the bare minimum with the warning shown.
- Docs: new "Minimum and Recommended Rolls" section in `entropy.en.md` explaining both amounts and why the minimum is preserved.

Test plan for this update:

- [x] `pytest tests/pages/new_mnemonic/test_dice_rolls.py` (14 passed)
- [x] `python i18n/i18n.py validate`
- [x] `black --check src tests`
- [x] Built and flashed on a Maix Amigo to check the screens

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SatsCzar:
Before:
before
After:
after

docs/getting-started/features/entropy.en.md
```diff
@@ -142,8 +162,8 @@ To mitigate this issue, Krux has implemented a pattern detection algorithm that
 
 ## What Krux Does?
 
-- Krux requires a minimum number of rolls based on theoretical entropy.
-- Krux warns the user if low Shannon's entropy, calculated with the actual rolls, is detected.
+- Krux requires a minimum number of rolls, calibrated so that a fair die's min-entropy reliably reaches the target bit count at that minimum.
```
qlrd, line 165: I went similarly on this concept of calibration and we discussed about how much abstract is this concpetion of "calibration". Maybe it's different here, but idk for sure.

docs/getting-started/features/entropy.en.md
```diff
@@ -142,8 +162,8 @@ To mitigate this issue, Krux has implemented a pattern detection algorithm that
...
+- Krux warns the user if low min-entropy, calculated with the actual rolls, is detected.
```
qlrd, line 166: this too

src/krux/pages/new_mnemonic/dice_rolls.py
```diff
@@ -39,15 +39,17 @@
...
 D20_STATES = [str(i + 1) for i in range(20)]
 
-D6_12W_MIN_ROLLS = 50
-D6_24W_MIN_ROLLS = 99
-D20_12W_MIN_ROLLS = 30
-D20_24W_MIN_ROLLS = 60
+D6_12W_MIN_ROLLS = 70
+D6_24W_MIN_ROLLS = 125
+D20_12W_MIN_ROLLS = 60
+D20_24W_MIN_ROLLS = 90
```
qlrd, line 45: Changing those numbers give some chills, but is about a non-stronger opinion cc @jdlcdl @joaozinhom @odudex @tadeubas

qlrd (changes requested):
1st round: needs to be based on `develop` (see CONTRIBUTING.md).

Besides that, i think it's kinda related with #745 (in the conceptual approach), but not sure, worth to a paralel compare @SatsCzar.

Below some nits:

tests/pages/new_mnemonic/test_dice_rolls.py
```diff
@@ -145,7 +160,9 @@ def test_new_12w_from_d6(m5stickv, mocker):
             BUTTON_ENTER,  # 1 press to confirm SHA
         ]
     )
-    MNEMONIC = "diet glad hat rural panther lawsuit act drop gallery urge where fit"
+    MNEMONIC = (
+        "reflect net crush light tobacco cat boil begin laptop fish kangaroo speed"
+    )
```
qlrd:
```suggestion
    MNEMONIC = "reflect net crush light tobacco cat boil begin laptop fish kangaroo speed"
```

tests/pages/new_mnemonic/test_dice_rolls.py
```diff
@@ -213,7 +230,9 @@ def test_new_12w_from_d6_on_amigo_device(amigo, mocker):
...
```
qlrd:
```suggestion
    MNEMONIC = "reflect net crush light tobacco cat boil begin laptop fish kangaroo speed"
```

qlrd (commented):
some test nits, non-blockable

jdlcdl:
Just my first thoughts: to start with another way to measure (as proposed with great docs to explain what it means) and to display "Entropy bits: min: x; shannon: y" but not yet to force users to roll until a new threshold is met.

src/krux/pages/new_mnemonic/dice_rolls.py
SatsCzar: I understand your concern. Just to clarify, changing the minimum number of rolls is not directly related to issue #923, but rather to the difference between Shannon entropy and min-entropy.

In the images from this comment (https://github.com/selfcustody/krux/pull/945#issuecomment-5519422012) showing the before and after, a similar number of biased-die rolls were used in both cases. With Shannon entropy, the result was 297 bits, which passed the entropy check, while with min-entropy it resulted in only 178 bits.

The issue is that keeping the same minimum number of rolls (50 for 12 words and 99 for 24 words) would result in the user seeing the insufficient entropy warning 100% of the time.

SatsCzar:
> some test nits, non-blockable

Hey @qlrd, thank you for pointing out the good formatting practices.

However, this was an automatic formatting by the black formatting library installed in the project, and it follows the rule of not allowing line lengths greater than 88, according to: https://github.com/psf/black/blob/main/src/black/const.py#L1

Adjusting this will cause lint-black in CI to break.

SatsCzar:
> Just my first thoughts: to start with another way to measure (as proposed with great docs to explain what it means) and to display "Entropy bits: min: x; shannon: y" but not yet to force users to roll until a new threshold is met.

Interesting @jdlcdl. We could continue allowing seed generation with the pre-PR number of dice rolls, while displaying a "low entropy" warning, and have the UI recommend a higher number of rolls.

In that sense, we could have two separate variables: one for the minimum and another for the recommended number of rolls.

The advantage is that we would avoid introducing a breaking change for existing users who previously generated their seeds using fewer rolls.

I'm not sure how common it is for users who have already generated their seeds to later recover them using the original dice-roll sequence. What do you think?

jdlcdl:
^^^ above sounds reasonable to me (more measurements instead of less are better).

I honestly don't know if it's common to record everything that goes into creating a mnemonic... it certainly is no harm as long as bundled together with archives that are well-secured.  Imagine suspecting a bug in the past and wanting to re-verify rolls->mnemonic... or finding out that someone else can learn your mnemonic or bip32 xprv... the one who can explain how they arrived at bip39 entropy (the pre-image string of dice rolls fed into sha256) could convincingly be the real owner.   For tests, it's very common to record rolls and try them against different implementations while gaining confidence in a signer.

qlrd:
> r, this was an automat

Was editor automatic one? because it happens a lot with me and i just delete diff, runs black locally and generally is accepted by both local and CI. I ask because found weird to not accept in CI, because the 80 lines have some degree of acceptance on pass the chars limit.

qlrd:
One thing that is bothering me is that the opened https://github.com/SeedSigner/seedsigner/issues/989 has no upstream discussion with similar content. I'm reading the documentation, but it would take a while.

src/krux/pages/new_mnemonic/dice_rolls.py
```diff
@@ -80,8 +82,15 @@ def shannon_sum(self, distribution, sample_size):
             unit_entropy -= probability * (probability and math.log2(probability))
         return unit_entropy
 
+    def min_entropy_sum(self, distribution, sample_size):
+        """Calculates min-entropy (H_min = -log2(p_max)) of a given distribution"""
+        import math
+
+        p_max = max(distribution) / sample_size
+        return -math.log2(p_max)
```
qlrd, line 90: llm nit:

A 70-sample plug-in estimate cannot distinguish a fair die from a mildly loaded one; the security posture is unchanged (warning only, "Proceed anyway" remains) but the docs overstate what the check assures.

src/krux/pages/new_mnemonic/dice_rolls.py
```diff
@@ -39,15 +39,17 @@
...
+D20_24W_MIN_ROLLS = 90
 MIN_ENTROPY_12W = 128
 MIN_ENTROPY_24W = 256
 
-# Min. rolls hardly will reach min. entropy according to Shannon's index
-# With a small tolerance, excessive low entropy warnings won't pop up when min. rolls are used.
+# Min-entropy (H_min = -log2(p_max)) is a worst-case measure: even a fair die needs
+# noticeably more rolls than Shannon's average-case entropy would suggest to reliably
+# clear the bit target above. These minimums are calibrated so that ~90% of fair-die
+# rolls reach it at the minimum count; the small tolerance below covers the rest.
```
qlrd, line 52: i think it's kinda innacurate claim, i mean, i think deserves better explanation. llm nit: ` State that -log2(max/N) is biased low by O(sqrt N) bits at small N and the counts compensate for that`

docs/getting-started/features/entropy.en.md
```diff
@@ -128,11 +128,31 @@ Calculating Shannon's entropy on a real sample of dice rolls provides insights i
...
+![H_{min,total} = N \times \left(-\log_2(p_{max})\right)](img/math/eq-25.svg){ .math-display }
+
+### Example
+
+Using the same 50-roll sample from the Shannon example above (`4, 9, 7, 10, 12, 8` occurrences for faces 1 through 6), the most frequent face is `5`, rolled 12 times:
+
+![p_{max} = \frac{12}{50} = 0.24](img/math/eq-26.svg){ .math-display }
+![H_{min} = -\log_2(0.24) \approx 2.059 \text{ bits per roll}](img/math/eq-27.svg){ .math-display }
+![H_{min,total} = 50 \times 2.059 \approx 103 \text{ bits}](img/math/eq-28.svg){ .math-display }
+
```
qlrd, line 148: llm nit: `Arithmetic is right (102.9 bits) but "precisely why more rolls are required" names the wrong cause. Suggested fix: Add that a fair die's true min-entropy is 2.585 bits per roll, 129 bits at 50 rolls, and the requirement comes from the estimator.`

tests/pages/new_mnemonic/test_dice_rolls.py
```diff
@@ -11,67 +11,82 @@
...
-    + [1] * 10  # 10 number 2 presses
-    + [2] * 10  # 10 number 3 presses
-    + [3] * 10  # 16 number 4 presses
-    + [4] * 6  # 2 number 5 presses
-    + [5] * 4  # 2 number 6 presses
-)
+# rolls which will result in poor min-entropy AND a detectable pattern: cycling
+# through only faces 1,2,3 skews the distribution (p_max ~= 1/3) and produces a
+# constant derivative between consecutive rolls
+POOR_ROLLS_TOUCH_SEQUENCE = [0, 1, 2] * 23 + [0]
```
qlrd, line 17: llm nit: `New poor fixture has 108 bits and a 73% pattern score, so it always fires both warnings; no test isolates the poor-entropy warning (coupling pre-existing).`

src/krux/translations/pt.py
```diff
@@ -195,6 +195,7 @@
     "Média",
     "Mensagem",
     "Mensagem:",
+    "Min-entropia:",
```
qlrd, line 198: I found this a little weird, why not "Entropia mínima"?

tests/pages/new_mnemonic/test_dice_rolls.py
SatsCzar: Fixed in https://github.com/selfcustody/krux/pull/945/commits/c3ba5f1bedc63b04140cee03d4bb8546268261a1

tests/pages/new_mnemonic/test_dice_rolls.py
SatsCzar: Fixed in https://github.com/selfcustody/krux/pull/945/commits/c3ba5f1bedc63b04140cee03d4bb8546268261a1

SatsCzar:
> > r, this was an automat
> 
> Was editor automatic one? because it happens a lot with me and i just delete diff, runs black locally and generally is accepted by both local and CI. I ask because found weird to not accept in CI, because the 80 lines have some degree of acceptance on pass the chars limit.

I'm not sure how to answer you because I don't have experience with Python code formatters (my background is in ESLint and Prettier).

I only know that CI lint-black broke: https://github.com/selfcustody/krux/actions/runs/34005814199/job/101412658985?pr=945

SatsCzar:
> ^^^ above sounds reasonable to me (more measurements instead of less are better).
> 
> I honestly don't know if it's common to record everything that goes into creating a mnemonic... it certainly is no harm as long as bundled together with archives that are well-secured. Imagine suspecting a bug in the past and wanting to re-verify rolls->mnemonic... or finding out that someone else can learn your mnemonic or bip32 xprv... the one who can explain how they arrived at bip39 entropy (the pre-image string of dice rolls fed into sha256) could convincingly be the real owner. For tests, it's very common to record rolls and try them against different implementations while gaining confidence in a signer.

@jdlcdl Adjusted in https://github.com/selfcustody/krux/pull/945/commits/63ef927a6138b5f701ed5b0a7475c55e7fff8224, now maintaining a minimum number of dice rolls and also a recommended value (meeting the minimum entropy).

It has also been adjusted to display Shannon's entropy along with the minimum entropy on the statistics screen.

I think I'll add an update section to the pull request with these changes that have occurred, which are different from the original proposal I had conceived.

tests/pages/new_mnemonic/test_dice_rolls.py
SatsCzar: It was corrected as a side effect of 63ef927a

docs/getting-started/features/entropy.en.md
SatsCzar: Good catch, thanks. Fixed in ba7d0e78. the paragraph now states that a fair die's true min-entropy is `-log2(1/6) ≈ 2.585 bits/roll` (same as Shannon in that case), which would total `50 × 2.585 ≈ 129 bits` at 50 rolls — already above the 128-bit target — and attributes the practical need for more rolls to the plug-in estimator's sampling noise instead, pointing to the "A Note on Sample Size" section for the detail.

src/krux/translations/pt.py
SatsCzar: Fixed in cad51f4

